PR #17, opened 2023-06-16
using sh instead of xterm

ZxPat:
I removed the need of xterm and it's now using sh to run the script without opening a new  xterm.


ivan-hc:
Hi, and thank you. The reason I've included xterm among the dependences is to give a clear output on the main terminal's window, I'm looking for something better. More details in this disscussion https://github.com/ivan-hc/AM-Application-Manager/discussions/16 this is why I can't merge this PR. Thank you again.

PR #18, opened 2023-06-16
Use builtin env. variables and replace sed with tail. 

DLopezJr:
First batch of general optimizations.

1. Replaced ```uname -n``` with ```printf '%s\n' "$HOSTTYPE"```. 
- ```"$HOSTTYPE"``` is a builtin environment variable in bash.

2. Replaced ```who | awk '{print $1}'``` with ```$(printf '%s\n "$USER"```.
- ```"$USER"``` is a builtin environment variable in POSIX sh. 

3. Replaced all instances of ```sed '1d'``` with ```tail +2```.
- ```tail``` is faster than ```sed``` in this context.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -2,8 +2,8 @@
 
 AMPATH="/opt/am"
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM-application-manager/main"
-arch=$(uname -m)
-currentuser=$(who | awk '{print $1}')
+arch=$(printf '%s\n' "$HOSTTYPE")
+currentuser=$(printf '%s\n' "$USER")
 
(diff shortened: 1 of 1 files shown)


ivan-hc:
Hi, thank you for this PR, just tested, everything works well. You're the first contributor on this repo in two years this repository exists! :)

ivan-hc:
Hi @DLopezJr 

https://github.com/ivan-hc/AppMan/issues/8

today I had this issue with a Fedora user, seems that the command is `tail -n +2`, not just `tail +2`, so I corrected this for both AM and AppMan

DLopezJr:
Good catch, the POSIX spec does follow that syntax you added.

PR #32, opened 2023-06-25
Changed dependency commands to POSIX compliance.

DLopezJr:
I would like to be able to run AM on distros like Alpine or KISS which do not have bash. I will slowly start porting it over to POSIX sh. 

Here are the details of the changes:

= echo and test are builtin commands for all shells. It's not necessary to check for them. 
== You can test it with this command: $ type echo test

= egrep was deprecated for "grep -E". I don't see egrep actually being used in this file. Is it used anywhere in the repo?
== https://www.gnu.org/software/grep/manual/grep.html

= Changed the bash "[[" command to the POSIX "[" command.

= Replaced "which" with the POSIX alternative, "command -v".

= Replaced echo with printf: https://unix.stackexchange.com/a/65819

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -5,11 +5,10 @@ AMREPO="https://raw.githubusercontent.com/ivan-hc/AM-application-manager/main"
 arch=$(printf '%s\n' "$HOSTTYPE")
 currentuser=$(printf '%s\n' "$USER")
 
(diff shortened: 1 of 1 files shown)


ivan-hc:
I've tested the command "am -l", this is the output:

    /usr/local/bin/am: riga 11: [: : attesa espressione intera

    Install the above and try again

DLopezJr:
Good catch! Was failing due to "$deps" being empty. Got it working with 'am -l' now.

PR #34, opened 2023-06-26
Changed printf to HEREDOC

DLopezJr:
Sorry for the quick change. Just remembered that long printf statements are easier to read as a heredocument.

= https://mywiki.wooledge.org/HereDocument


ivan-hc:
Don't worry, I think is already OK this way, just merged the previous commit also in appman. Thank you.

Do not misunderstand me, but I don't think is strongly necessary. `printf` is already OK to me.

PR #55, opened 2023-08-19
Update brave-appimage

BlueCobbleStone:
Fix brave appimage URL (actually not working)


ivan-hc:
thanks, I had problems in the past with that repo because the big amount of releases per hour, this is why I added a query for the last 100 pages (see `?per_page=100`). I hope we don't need to change that again.

PR #59, opened 2023-08-26
Add GoldenDict-ng

xz-dev:
I had use `sort x86_64-apps -o x86_64-apps`, so maybe there are some change except goldendict-ng


ivan-hc:
> I had use `sort x86_64-apps -o x86_64-apps`, so maybe there are some change except goldendict-ng

thank you for this!

PR #72, opened 2023-10-02
Add Falcon

Errorts:
Audio File Encoder. Extracts audio tracks from an audio CD image to separate tracks


ivan-hc:
Hi, there is an error, the app name is "flacon", not "falcon" as you said https://github.com/flacon/flacon

PR #101, opened 2024-01-02
Rename ungoogled-chromium, fix bash warning

stasadev:
Add a correct name for `ungoogled-chromium`.

And fix a bash warning:

```bash
$ appman install test
-----------------------------------------------------------------------
 >> START OF ALL INSTALLATION PROCESSES <<
-----------------------------------------------------------------------

 "test" is NOT a valid argument, please try again!

appman: line 883: break: only meaningful in a `for', `while', or `until' loop
```

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -880,7 +880,7 @@ case "$1" in
 										break; done
 								esac
 							else
-								echo ' "'$arg'" is NOT a valid argument, please try again!'; echo ""; break
+								echo ' "'$arg'" is NOT a valid argument, please try again!'; echo ""; exit 1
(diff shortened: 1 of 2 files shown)


ivan-hc:
I've changed the name in the script ungoogled-chromium, sorry, today I'm very busy with some works on GNOME-related AppImages.

About the fix in APP-MANAGER, no, it is not a good idea (for me at least), because if you install agument 1, 2 and 3 and 2 does not exists it exits the installation and 3 is not installed.

ivan-hc:
It is enough to remove "break" instead, thank you for the idea

PR #142, opened 2024-02-14
fixed cat << EOF

zen0bit:
I just wrongly placed one line


ivan-hc:
Now tell me why after this pull request I get this message every time I open the terminal:
```
bash: /etc/bash_completion.d/am-completion.sh: riga 3: EOF non atteso durante la ricerca di "'"
```
Now I'm getting really nervous!

PR #211, opened 2024-03-05
Fix sed causes wrong substitution when local path contains `/opt`

directentis1:
My use case when local path contains `/opt` (`/home/username/opt/Applications'), it causes this sed command doing wrong.

I think `sed -i -E "s#([^_[:space:]])/opt/#\1$APPSPATH/#g" ./$arg` is more suitable, but I'm not sure 😅️

### DIFF
--- a/modules/install.am
+++ b/modules/install.am
@@ -103,7 +103,7 @@ while [ -n "$1" ]; do
 
 					function _install_appman_local_patch(){
 						sed -i "s# /opt# $APPSPATH#g" ./$arg
-						sed -i "s#/opt/#$APPSPATH/#g" ./$arg
+						sed -i "s#Icon=/opt/#Icon=$APPSPATH/#g" ./$arg
 						sed -i "s# /usr/local/bin# $(xdg-user-dir USER)/.local/bin#g" ./$arg
 						sed -i "s# /usr/bin# $(xdg-user-dir USER)/.local/bin#g" ./$arg
 						sed -i "s# /usr/local/games# $(xdg-user-dir USER)/.local/bin#g" ./$arg


directentis1:
> This may cause issues with other configurations.
> 
> Try to invert lines 105 and 106 instead.

@ivan-hc 
hmm, `sed -i "s#/opt/#$APPSPATH/#g` still changes every instance of `/opt` regardless of whether it is preceded by a space or not.
It's hard to figure out the neatest way to write code in such an ambiguous case..
So I think, `sed -i -E "s#([^_[:space:]])/opt/#\1$APPSPATH/#g" ./$arg` is stills more suitable

image

ivan-hc:
have you updated the module?

ivan-hc:
I've done some tests with a configuration like yours and worked

ivan-hc:
I've installed brave without problems

what is the content of your ~/.config/appman/appman-config file?

ivan-hc:
I'm sure you wrote the whole path when initialized AppMan for the first time (while it should be only "opt/AppImages")

ivan-hc:
I've installed both brave and vivaldi-stable, without problems (I've tested with "opt/AppImages" in the appman-config file).

Maybe the "brave" script is still in the appman/.cache directory. Thry to clean the directory using `appman -c`

directentis1:
> I've installed brave without problems
> 
> what is the content of your ~/.config/appman/appman-config file?

It's `opt/AppImages`... I had set it from yesterday.
I think maybe some issues with linking command (from APPSPATH/app dir to .local/share/application), cause Brave run smoothly directly from APPSPATH/brave/brave.

PR #325, opened 2024-04-02
More XDG shenanigans download.am

Samueru-sama:
Tested it by downloading and installing libreoffice and it worked.

Also replaced some instances of $(xdg-user-dir USER) for just $HOME, it is more efficient this way since we don't need to call xdg-user-dir binary that often.

### DIFF
--- a/modules/download.am
+++ b/modules/download.am
@@ -1,5 +1,12 @@
 #!/usr/bin/env bash
 
+# Check XDG variable
+if [ -n "$XDG_DATA_HOME" ]; then
+        DATADIR="$XDG_DATA_HOME"
+else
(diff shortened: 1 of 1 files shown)


ivan-hc:
@Samueru-sama in first you should resolve the conflicts

also, try not to push these commits to main but to [dev](https://github.com/ivan-hc/AM/tree/dev)

Samueru-sama:
> @Samueru-sama in first you should resolve the conflicts
> 
> also, try not to push these commits to main but to [dev](https://github.com/ivan-hc/AM/tree/dev)

Okay sorry about that, I think it is ready now.

PR #333, opened 2024-04-02
More XDG changes

Samueru-sama:
With this change I think we can remove the variable check in the download module, since the variable will always be passed down from APP-MANAGER right?

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -191,6 +191,7 @@ cat <<-HEREDOC >> "$AMPATH/options"
 	--force-latest
 	--home
 	--launcher
+        --less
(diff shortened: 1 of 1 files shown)


Samueru-sama:
> > With this change I think we can remove the variable check in the download module, since the variable will always be passed down from APP-MANAGER right?
> 
> no problems

Well in that case what needs to be deleted is lines 3 to 8 of the `downloads.am` module. I don't do it because I don't fully understand how the variable is being passed from one module to the other lol.

ivan-hc:
just tested, I'll did it for you with the next commit

PR #337, opened 2024-04-02
Even more XDG changes.

Samueru-sama:
All of these previous PRs probably should have been a single PR, sorry. 

I'm assuming that `$DATADIR` gets passed down from APP-MANAGER to every module always btw. Also I have no idea why I get a diff in the end of `esac` in sandboxes.am


ivan-hc:
> I'm assuming that `$DATADIR` gets passed down from APP-MANAGER to every module always btw. Also I have no idea why I get a diff in the end of `esac` in sandboxes.am

I see, strange.

PR #341, opened 2024-04-03
Optimization in the files.am module.

Samueru-sama:
This cuts down the time it takes for `am -f --less` to print the info from 32ms to ~23ms. 

benchmark: 
image

(The first 3 tests were using this PR while the later 3 tests are with the old method). 

I was playing with adding AM to fastfetch and I noticed a delay when printing the info, usually in fastfetch each module takes less than 10ms to load and here it was taking 33ms so I took a look at the script, it is far from perfect still but still an improvement.

You may want to download the file and compare it on your PC because the github diff gets confused by the lines being moved around btw. It thinks that I renamed certain variables when that is because similar strings changed places in the script and now the diff thinks that I renamed them and that results in it thinking that I renamed appimage-type2 to unknown for example.

### DIFF
--- a/modules/files.am
+++ b/modules/files.am
@@ -1,98 +1,87 @@
 #!/usr/bin/env bash
 
-function _files_simple(){
-	echo ""
-	echo $(echo "YOU HAVE INSTALLED "
-	cd "$APPSPATH" &&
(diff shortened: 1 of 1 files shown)


ivan-hc:
Tested on both AM and AppMan. Thank you

PR #343, opened 2024-04-03
Make xdg-user-dirs an optional dependency

Samueru-sama:
This changes make xdg-user-dirs an optional dependency. Tested it with the template module with and without `xdg-user-dir` and it worked like it should.

Introduces a new variable named `SCRIPTDIR` which if `xdg-user-dir` is present it gets set as `xdg-user-dir DESKTOP` and everything works like it used to.

If `xdg-user-dir` is not present then the variable gets defined as `$HOME` and the `template.am` module puts the am-scripts dir in `$HOME` instead for example.

(Sorry about the readme.md changing in this pr lol).

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -3,7 +3,7 @@
 DIR="$( cd "$( dirname "$0" )" && pwd )"
 CLI=$(basename "$0")
 
-if ! command -v xdg-user-dir 1>/dev/null; then echo ' ⚠️ WARNING: "xdg-user-dir" command not found (required by modules)'; exit; fi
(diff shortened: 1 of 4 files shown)


ivan-hc:
Fun fact:

hours 10:15 AM here in Italy, I did a `am -d anydesk` to tet a new feature in "install.am" and... the script was downloaded in "$HOME" and I said "WTF???"

2 minutes later... I've discovered that the owner of "AM" (me) was not updated with "main"

after this, the script was finally downloaded in /home/ivan/Scrivania

I feel so stupid...

PR #358, opened 2024-04-06
Add safety checks hyperfine

Samueru-sama:
`if [ -z "$APP" ]; then exit 1; fi` prevents the script from continuing if for whatever reason `$APP` were to be empty (this scenario would be somebody accidentally deleting it while changing the name of the application for example). This is important because later on, you will pipe an `rm -rf /usr/local/bin/$APP` to the `remove` script, if the script allows for an empty variable it will result in a remove script `rm -rf /usr/local/bin/` and that would be very bad. 

The other change: `echo "rm -rf \"/opt/$APP\" \"/usr/local/bin/$APP\"" >> /opt/$APP/remove` adds the quotations to the path for rm -rf, that way an accidental space in the `remove` script rm -rf doesn't result in catastrophic results either.

Btw Ivan get sed ready because I think this is something that needs to be done on every script, specially the check for the empty variable.

### DIFF
--- a/programs/x86_64/hyperfine
+++ b/programs/x86_64/hyperfine
@@ -1,6 +1,7 @@
 #!/bin/sh
 
 APP=hyperfine
+if [ -z "$APP" ]; then exit 1; fi
 SITE="sharkdp/hyperfine"
 
 # CREATE THE FOLDER
@@ -9,7 +10,7 @@ cd /opt/$APP
 
 # ADD THE REMOVER
 echo '#!/bin/sh' >> /opt/$APP/remove
-echo "rm -R -f /opt/$APP /usr/local/bin/$APP" >> /opt/$APP/remove
+echo "rm -rf \"/opt/$APP\" \"/usr/local/bin/$APP\"" >> /opt/$APP/remove
 chmod a+x /opt/$APP/remove
 
 # DOWNLOAD THE ARCHIVE


ivan-hc:
Forgive me, I don't want to question your observations, but the scripts are almost all the same. If I had written something catastrophic, I would have known about it a couple of years ago (with a lot of complaints in tow).

Also, how can the "$APP" variable be empty if it was already assigned to the line above?

Samueru-sama:
> Forgive me, I don't want to question your observations, but the scripts are almost all the same. If I had written something catastrophic, I would have known about it a couple of years ago (with a lot of complaints in tow).
> 
> Also, how can the "$APP" variable be empty if it was already assigned to the line above?

It can happen if the user accidentally deletes the line if they try to change the name of the application for example. 

Do you know the steambug? that bug was there was several years until somebody triggered it by running the steam script with `bash steam.script` and that caused a variable to be empty and result in a `rm -rf /*`. 

And the second issue can be caused by adding an accidental space, which happened recently in some kde theme bug. And also the bumblebee bug had it. The quotations prevents that bug from ever happening. 

(I would really add those fixes if I were you, I don't wanna end in a Kevin Fang videos detailing an rm -rf bug from AM lol).

ivan-hc:
> It can happen if the user accidentally deletes the line if they try to change the name of the application for example.

The scripts are provided as they are, if the user intends to change them for fun that's his problem, not our.

> Do you know the steambug? that bug was there was several years until somebody triggered it by running the steam script with `bash steam.script` and that caused a variable to be empty and result in a `rm -rf /*`.

It happened to me a few weeks ago with a poorly written option, but it's solved. Never happened in the three years I've been on github, and it won't happen again.

As far as the scripts are concerned, I repeat, they are almost all the same and in all this time they have never caused problems, at most not an icon or a launcher appeared in the applications menu.

> And the second issue can be caused by adding an accidental space, which happened recently in some kde theme bug. And also the bumblebee bug had it. The quotations prevents that bug from ever happening.

The problem was with third party themes managed by KDE. This was not a problem in KDE itself.

> (I would really add those fixes if I were you, I don't wanna end in a Kevin Fang videos detailing an rm -rf bug from AM lol).

I don't know who this guy is, but if he did it, say, change the name of the script and delete everything by changing some spaces... so without even knowing what he's doing... it means he's an idiot, more than LTT when has installed the Steam.deb in PopOS.

---------------------------

Having said that, I have worked in recent years to make the scripts as similar as possible to each other, in order to make it easier for me to publish apps without wasting too much time compiling things (individual update scripts, adding icons and launchers... .).

Your speech is flawless, you're right!

But in reality I was already inclined to rewrite all the scripts, or maybe "have them rewritten by others", that's why so much interest in promoting third-party repositories, and that's why the discussion.

Honestly, maintaining all these projects led to stress, I got too carried away, neglecting the rest.

If you think there's something I can do other than rewrite everything myself, suggest something to me. Today I solved the `-a` problem with third-party Markdown by first going through the most difficult solutions, down to adding very few lines.

What should I try to fix in the 1960+ installation scripts we have in the database?

ivan-hc:
I could edit the templates themselves and republish all the scripts... but there are too many. And it's not the first time I've rewritten hundreds of scripts at once. Trust me, it's a pain. I need help.

Samueru-sama:
> The scripts are provided as they are, if the user intends to change them for fun that's his problem, not our.

You don't really see a problem with having a potential rm -rf on empty variables? 

>The problem was with third party themes managed by KDE. This was not a problem in KDE itself.

The problem was that the third party theme was doing an rm -rf on a variable, which due to a change between kde5 and kde6 resulted in that variable changing to `/some/path/ /` (notice the space) and that caused the issue. 

>it means he's an idiot, more than LTT when has installed the Steam.deb in PopOS.

Didn't debian change apt after this anyway? It is true that people dont read what the terminal says when they are installing something. they just want their software installed.

On pacman if a package is going to remove essential dependencies, you have to pass the `--nodeps --nodeps` flag twice, you will never be prompted to confirm to remove essential packages otherwise, to exactly prevent stuff like what linus did lol. 

>What should I try to fix in the 1960+ installation scripts we have in the database?

Well at the very least inject the `if [ -z "$APP" ]; then exit 1; fi` line in every script after the APP variable, this really prevents the big potential issue which is the variable ever becoming empty with rm -rf. 

I think this should be possible with sed to make it faster, but I haven't looked into it tbh so I don't know. 

The other thing which is `echo "rm -rf \"/opt/$APP\" \"/usr/local/bin/$APP\"" >> /opt/$APP/remove` (that introduces the quotations in the paths) will be something that will likely need more manual intervention to fix. And this starts with this script I wrote in this very pr. 

I love your work Ivan and yeah that is a lot of stress (tbh I'm stressed with this very hyperfine script I made that is why I made this pr fixing it lol, I can't image in your case which you made thousands of them). 

>I don't know who this guy is

He has a good vid detailing the steambug on youtube (just search for his channel) and it also talks about all the very complicated solutions that everyone posted before showing that the solution was just to check that the variable wasn't empty,  that was it. It may not be a perfect solution but checking if the variable is empty has prevented the bug from every happening again.

ivan-hc:
But the "$APP" variable is not empty at all!

I don't know what editor you use to write scripts, I use Mousepad (from XFCE), ant with the color scheme this is the result:

Istantanea_2024-04-07_04-40-44

the violet "$APP" and "$version" environment variables will be replaced with the value of green parts.

More preciselly, the part where the "remove" script is created, the pink text will remain as it is, while the violet one will be changed with the green value at line 3 (`APP=hyperfine`, so it wil be `hyperfine`).

Why should it fail?

ivan-hc:
Also Shelcheck says that your script is quite perfect!

I say "quite" because I've not put  `| return` after `cd` if it fails... but it is perfect!

ivan-hc:
Furthermore, I, being the author of the template, to verify that everything works I just have to compare your script with my template and see the different parts, to verify that the script is ok.

Samueru-sama:
>But the "$APP" variable is not empty at all!

I know, this change is that in the case it ever becomes empty (due to user error) it doesn't result in a rm -rf on an empty variable. 

I also use mousepad with this color scheme: https://imgur.com/hTRO4gB.png but once again this change is to prevent potential user error, which will happen one day, I've helped enough linux users in reddit and similar places to know this lol. 

>I say "quite" because I've not put | return after cd if it fails... but it is perfect!

Also yeah, one of the reasons why `cd` is when `$APP` is empty, which wouldn't even happen if the script stopped when it checks that `$APP` is empty.

(I am not telling you to not add the return btw, just that this also shows why it is a good idea to make sure that the variable isn't empty before continuing). 

>Furthermore, I, being the author of the template, to verify that everything works I just have to compare your script with my template and see the different parts, to verify that the script is ok.

I don't feel ok with the way the script written right now xd.

ivan-hc:
> > But the "$APP" variable is not empty at all!
> 
> I know, this change is that in the case it ever becomes empty (due to user error) it doesn't result in a rm -rf on an empty variable.

again, we're only responsible of what we write, not about what the newbies want to play with without knowledges. We can't overwatch everything.

> I also use mousepad with this color scheme: https://imgur.com/hTRO4gB.png but once again this change is to prevent potential user error, which will happen one day, I've helped enough linux users in reddit and similar places to know this lol.

yeah! lol

> > I say "quite" because I've not put | return after cd if it fails... but it is perfect!
> 
> Also yeah, one of the reasons why `cd` can fail is when `$APP` is empty, which wouldn't even happen if the script stopped when it checks that `$APP` is empty.
> 
> (I am not telling you to not add the return btw, just that this also shows why it is a good idea to make sure that the variable isn't empty before continuing).

the directories will be all created with the designed name, from the variable, that is NOT empty

> > Furthermore, I, being the author of the template, to verify that everything works I just have to compare your script with my template and see the different parts, to verify that the script is ok.
> 
> I don't feel ok with the way the script written right now xd.

The first time, we were all afraid of becoming fathers too early, it is with time that the experience... "comes" :rofl:

Samueru-sama:
> > > But the "$APP" variable is not empty at all!
> > 
> > 
> > I know, this change is that in the case it ever becomes empty (due to user error) it doesn't result in a rm -rf on an empty variable.
> 
> again, we're only responsible of what we write, not about what the newbies want to play with without knowledges. We can't overwatch everything.
> 
> > I also use mousepad with this color scheme: https://imgur.com/hTRO4gB.png but once again this change is to prevent potential user error, which will happen one day, I've helped enough linux users in reddit and similar places to know this lol.
> 
> yeah! lol
> 
> > > I say "quite" because I've not put | return after cd if it fails... but it is perfect!
> > 
> > 
> > Also yeah, one of the reasons why `cd` can fail is when `$APP` is empty, which wouldn't even happen if the script stopped when it checks that `$APP` is empty.
> > (I am not telling you to not add the return btw, just that this also shows why it is a good idea to make sure that the variable isn't empty before continuing).
> 
> the directories will be all created with the designed name, from the variable, that is NOT empty
> 
> > > Furthermore, I, being the author of the template, to verify that everything works I just have to compare your script with my template and see the different parts, to verify that the script is ok.
> > 
> > 
> > I don't feel ok with the way the script written right now xd.
> 
> The first time, we were all afraid of becoming fathers too early, it is with time that the experience... "comes" 🤣
> 
> EDIT: from "commit" to "condom" is a moment 😆

Ok I think fucked it up, I added Ryujinx and it became part of this PR lol. 

>again, we're only responsible of what we write, not about what the newbies want to play with without knowledges. We can't overwatch everything.

Well at least this change does overwatch that, and it is very easy to add. (at least from now on if you don't want to do it retroactively).

ivan-hc:
@Samueru-sama  If you care so much, I'll add these changes you've made to the code... but you have to do me a favor: don't change the writing of the scripts too much. They have already been tested. It ends up that if I have to compare them with the basic models I get confused, and that's where the problem lies. Try to stick to the basic models, please.

These templates have never given anyone any problems. Furthermore, I have adopted a standard format to better control whether someone writes a regular script or adds malicious code.

Let's say that YOU make a mistake in compiling the "remove" script, perhaps because you put an extra accent or slash... I'm the one who is accepting your PR by mistake, and it ends up that due to your mistake, we delete everyone a user's data.

So, leave it to me to supervise the scripts that go into my repository and database. I don't want problems.

ivan-hc:
I've checked the scripts, everything is OK.

But please, I want this is the last time that you change these scripts like this.

Again, I've adopted and published a standard to make it easier for me to review new scripts and PR. Even just changing "-R -f" to "-rf" is an extra flag that I need to check when comparing the base to the new. I don't want any additional headaches, so please, try to keep the standard from now on, at least on my repo.

Thank you.

PR #364, opened 2024-04-07
Add deadbeef-devel

Samueru-sama:
Hey I don't know the rules about devel/nightly packages being here, so I just made a PR with one instead of asking xd. 

This adds the nightly build of deadbeef, it is released on the same package as the stable builds, made it because the stable releases of deadbeef move very slow. 

The way the #LINK is originally done for the stable deadbeef package made no sense to me, so I changed it for symlinks like other packages do, also expanded the desktop entry. 

Also I removed the original way to get the icon from a link since deadbeef already contains the icon in its directory and linked that instead.

### DIFF
--- a/programs/x86_64/deadbeef-devel
+++ b/programs/x86_64/deadbeef-devel
@@ -1,0 +1,119 @@
+#!/bin/sh
+
+APP=deadbeef-devel
+SITE="https://deadbeef.sourceforge.io"
+
+# CREATE THE FOLDER
+mkdir /opt/$APP
+cd /opt/$APP
+
+# ADD THE REMOVER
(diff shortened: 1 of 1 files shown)


ivan-hc:
Sorry, but the way you did it, it will overwrite the normal deadbeef in some points (for example in /usr/local/bin)

I suggest to keef the value in "$APP". See the variants of Firefox, for example.

Samueru-sama:
> Sorry, but the way you did it, it will overwrite the normal deadbeef in some points (for example in /usr/local/bin)
> 
> I suggest to keef the value in "$APP". See the variants of Firefox, for example.

Done

PR #366, opened 2024-04-08
Small fixes to deadbeef and deadbeef-devel

Samueru-sama:
Alright these changes put the desktop entries of both applications up to date, they also have the `exec=appname %F` which was missing from the original script. Not having the %F prevents deadbeef from opening files thru xdg-open. 

I also reverted that original change to deadbeef-devel to use the built in icon from deadbeef, because turns out that icon has a very low quality lol.


ivan-hc:
Just a tip, you don't need to made the symlink executable, in $PATH

PR #375, opened 2024-04-09
Add AppImages of DeaDBeeF and Ryujinx. 

Samueru-sama:
Adds these to the repo: 

https://github.com/Samueru-sama/DeaDBeef-AppImage
https://github.com/Samueru-sama/Ryujinx-AppImage

These scripts work, and they also have a working update mechanism. However I noticed that the `Exec=` in the .desktop entry of both scripts contains the name of the symlink in `$PATH` instead of the full path to the appimage. I don't see a problem with that since they have unique names but let me know otherwise.  (I really don't know how appman replaces the `Exec=` with the full path to the appimage instead)

### DIFF
--- a/programs/x86_64/deadbeef-devel
+++ b/programs/x86_64/deadbeef-devel
@@ -1,11 +1,11 @@
 #!/bin/sh
 
-APP=deadbeef-devel-appimage
-REPO="Samueru-sama/DeaDBeef-AppImage"
+APP=deadbeef-devel
+SITE="https://deadbeef.sourceforge.io"
 
-# CREATE DIRECTORIES
-if [ -z "$APP" ]; then exit 1; fi
-mkdir /opt/$APP && cd /opt/$APP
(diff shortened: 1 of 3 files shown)


Samueru-sama:
@ivan-hc Did you change anything in the scripts that I didn't notice? I now tested installing them thru appman and now my desktop entries do contain the full path as `Exec=/path/to/appimage` when before it was just `Exec=$APP`?

ivan-hc:
nope, the install.am module has a patch that, if used with AppMan, changes the Exec entry in the .desktop file to redirect it to the file directly. This wont happen in AM, instead.

Samueru-sama:
> nope, the install.am module has a patch that, if used with AppMan, changes the Exec entry in the .desktop file to redirect it to the file directly. This wont happen in AM, instead.

Oh, sorry about that. Yeah I was so confused by that. 😅

PR #384, opened 2024-04-12
Fix amdgpu_top

### LINKED ISSUE #381: amdgpu_top issue with the desktop entries.
context: https://github.com/Umio-Yasuno/amdgpu_top/issues/77

Basically we need to `--appimage-extract assets/*.desktop` for amdgpu_top and ignore the desktop entry that is in the top level of squashfs-root. 

I tried to do it but it didn't work so now I'm here 😅

Samueru-sama:
Hey turns out the changes you made didn't actually fix the issue 👀 As it is still installing only one desktop entry instead of the two it has. 

Also amdgpu_top has a dummy empty icon, it should not be used, the desktop entries by default use a generic system resources icon instead. 

Fixes #381 for real this time lol.


Samueru-sama:
Also added deadbeef-stable-appimage to the list of apps.

Issue #383, opened 2024-04-12
Include packages versions


zen0bit:
Nice to see which packages versions are available, updated, updating with..

What you think?🤔

ivan-hc:
Are you talking about `-f`?

Samueru-sama:
This is going to be a lot of work because the version file in each package is usually the full url download link.

I think it has to be done eventually though.

ivan-hc:
> This is going to be a lot of work because the version file in each package is usually the full url download link.
> 
> I think it has to be done eventually though.

its not just this, not all installed apps have this "version" file, for example, Firefox (the original one).

Another option would be tho check the version from the binary itself... but not fo all it works.

If you try this on Avidemux, for example, the app has no options at all, just the ones to extract the AppImage, elsewere the AppImage simply starts with its default GUI

Samueru-sama:
@ivan-hc In that case the best approach would be to add a new file for this, for example `actualversion` that will contain the version of the application, and little by little we start adding it. 

And once we have enough packages with `actualversion` file AM can start displaying it along side the package name. 

What do you think? I can start doing that to the apps I've added once you give the go ahead.

Samueru-sama:
For example for the packages that use github for download, the version is usually the second field in the `$version` url. 

So `pkgversion=$(echo $version | awk -F / '{print $(NF-1)}'` would do it for most packages.

ivan-hc:
> @ivan-hc In that case the best approach would be to add a new file for this, say for example `actualversion` that will contain the version of the application, and little by little we start adding it.
> 
> And once we have enough packages with `actualversion` file AM can start displaying it along side the package name.
> 
> What do you think? I can start doing that to the apps I've added once you give the go ahead.

I like the idea... but I think AM is doing too much.

Furthermore, AM was built to always install the latest version of the app available from the source. If the package turns out to be old, there are 2 reasons:
1. AM hasn't updated in a long time;
2. the source developer has not added an official AppImage, or any portable format that AM can handle.

For point 2, the choice would make more sense. It would mean reminding the user to go and "wand" the developer upstream to have him update the AppImage (and he will surely tell you "NO! Use Flatpak!"). Other than that, I don't see much reason to put that information in "`-f`".

ivan-hc:
> For example for the packages that use github for download, the version is usually the second field in the `$version` url.
> 
> So `pkgversion=$(echo $version | awk -F / '{print $(NF-1)}'` would do it for those packages.

[...]

PR #398, opened 2024-04-16
Changes to AM-SAMPLE-AppImage

Samueru-sama:
The first change is combining the `mv` and `--appimage-extract` commands. That way mv will move the .desktop file only if a file was actually extracted with --appimage-extract. 

The second change simplifies the patching in the .desktop file. I created this test demonstration that shows what it does: 

```
#!/bin/sh

APP=TESTDEMO

sed -i "s#Exec=[^ ]*#Exec=$APP#g" ./$APP.desktop
sed -i "s#Icon=.*#Icon=/opt/$APP/icons/$APP#g" ./$APP.desktop
```

And ran it the librewolf.desktop file thru, which this is how it looks before changing it: 

```
[Desktop Entry]
Name=LibreWolf
Exec=librewolf %u
Icon=librewolf
Type=Application
MimeType=text/html;text/xml;application/xhtml+xml;x-scheme-handler/http;x-scheme-handler/https;application/x-xpinstall;application/pdf;application/json;
StartupWMClass=LibreWolf
Categories=Network;WebBrowser;
StartupNotify=true
Terminal=false
X-MultipleArgs=false
Keywords=Internet;WWW;Browser;Web;Explorer;
Actions=new-window;new-private-window;profilemanager;
X-AppImage-Version=177

[Desktop Action new-window]
Name=Open a New Window
Exec=librewolf %u

[Desktop Action new-private-window]
Name=Open a New Private Window
Exec=librewolf --private-window %u

[Desktop Action profilemanager]
Name=Open the Profile Manager
Exec=librewolf --ProfileManager %u
```

And this is after running the change: 

```
[Desktop Entry]
Name=LibreWolf
Exec=TESTDEMO %u
Icon=/opt/TESTDEMO/icons/TESTDEMO
Type=Application
MimeType=text/html;text/xml;application/xhtml+xml;x-scheme-handler/http;x-scheme-handler/https;application/x-xpinstall;application/pdf;application/json;
StartupWMClass=LibreWolf
Categories=Network;WebBrowser;
StartupNotify=true
Terminal=false
X-MultipleArgs=false
Keywords=Internet;WWW;Browser;Web;Explorer;
Actions=new-window;new-private-window;profilemanager;
X-AppImage-Version=177

[Desktop Action new-window]
Name=Open a New Window
Exec=TESTDEMO %u

[Desktop Action new-private-window]
Name=Open a New Private Window
Exec=TESTDEMO --private-window %u

[Desktop Action profilemanager]
Name=Open the Profile Manager
Exec=TESTDEMO --ProfileManager %u

```

And just like that, every instance of Exec= and Icon= was changed to the desired string and it preserved the original launch flags.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -8,6 +8,7 @@ Before reporting a problem, be it a bug, design or others, we assume you have ma
 1. [the README.md](https://github.com/ivan-hc/AM/blob/main/README.md) does not cover your problem
 2. the problem has not been reported in the [issue tracker](https://github.com/ivan-hc/AM/issues)
 3. the problem is not related to the installed apps but to the [installation scripts](https://github.com/ivan-hc/AM/tree/main/programs)
+4. the "AM"/"AppMan" command line interface [is updated to the latest version](https://github.com/ivan-hc/AM/releases/latest)
 
 If all apply, then please consider opening a [new issue](https://github.com/ivan-hc/AM/issues).
 
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -75,12 +75,10 @@ chmod a+x /opt/$APP/AM-updater
 # LAUNCHER & ICON
 app=$(echo $APP | cut -c -3)
 cd /opt/$APP
-./$APP --appimage-extract *.desktop 1>/dev/null
-./$APP --appimage-extract share/applications/*.desktop 1>/dev/null
-./$APP --appimage-extract usr/share/applications/*.desktop 1>/dev/null
-mv squashfs-root/*.desktop ./$APP.desktop
-mv squashfs-root/share/applications/*.desktop ./$APP.desktop
-mv squashfs-root/usr/share/applications/*.desktop ./$APP.desktop
+./$APP --appimage-extract *.desktop 1>/dev/null && mv squashfs-root/*.desktop ./$APP.desktop
+./$APP --appimage-extract share/applications/*.desktop 1>/dev/null && mv squashfs-root/share/applications/*.desktop ./$APP.desktop
+./$APP --appimage-extract usr/share/applications/*.desktop 1>/dev/null && mv squashfs-root/usr/share/applications/*.desktop ./$APP.desktop
+
 if [ ! -e ./$APP.desktop ]; then 
 	rm ./$APP.desktop; ./$APP --appimage-extract usr/share/applications/*$app*.desktop 
 	mv squashfs-root/usr/share/applications/*.desktop ./$APP.desktop
@@ -89,13 +87,9 @@ if [ ! -e ./$APP.desktop ]; then
 	rm ./$APP.desktop; ./$APP --appimage-extract share/applications/*$app*.desktop 1>/dev/null
 	mv squashfs-root/share/applications/*.desktop ./$APP.desktop
 fi
-CHANGEEXEC=$(cat ./$APP.desktop | grep Exec= | tr ' ' '\n' | tr '=' '\n' | tr '/' '\n' | grep -i $app | head -1)
-sed -i "s#$CHANGEEXEC#$APP#g" ./$APP.desktop
-sed -i "s#AppRun#$APP#g" ./$APP.desktop
-sed -i "s#Exec=/bin/#Exec=#g" ./$APP.desktop
-sed -i "s#Exec=/usr/bin/#Exec=#g" ./$APP.desktop
-CHANGEICON=$(cat ./$APP.desktop | grep Icon= | head -1)
-sed -i "s#$CHANGEICON#Icon=/opt/$APP/icons/$APP#g" ./$APP.desktop
+
+sed -i "s#Exec=[^ ]*#Exec=$APP#g" ./$APP.desktop
+sed -i "s#Icon=.*#Icon=/opt/$APP/icons/$APP#g" ./$APP.desktop
 
 mkdir icons
 mv $(./$APP --appimage-extract *.png) ./icons/$APP 2>/dev/null


Samueru-sama:
Also why don't we just do `--appimage-extract .DirIcon` to get the icon in most cases?

.DirIcon can either be the actual icon, which on my appimages that is like 70% of the time. Or it can be a symlink to the actual icon. 

If it turns out to be a symlink, then we already know what the next --appimage-extract has to be to ge the icon by doing a `readlink ./squashfs-root/.DirIcon`

ivan-hc:
There is a issue, if we use && every time that we try to extract the launcher, and the launcher is just a brocken symlink or a dumb file, the next one will overwrite the previous one :thinking:

Samueru-sama:
> There is a issue, if we use && every time that we try to extract the launcher, and the launcher is just a brocken symlink or a dumb file, the next one will overwrite the previous one 🤔

Doesn't that happen already in the way it was done before? 

```
./$APP --appimage-extract *.desktop 1>/dev/null && mv squashfs-root/*.desktop ./$APP.desktop
./$APP --appimage-extract share/applications/*.desktop 1>/dev/null && mv squashfs-root/share/applications/*.desktop ./$APP.desktop
./$APP --appimage-extract usr/share/applications/*.desktop 1>/dev/null && mv squashfs-root/usr/share/applications/*.desktop ./$APP.desktop
```

here mv will overwrite ./$APP.desktop if there is a broken symlink or dumb file. 

Either way the order is what is most important here, notice that we first try `*.desktop ` then `share/applications/*.desktop` then `usr/share/applications/*.desktop` it would be extremely rare for `usr/share/applications/*.desktop` to be a symlink that would overwrite `*.desktop`. 

Now if I were doing this backwards, then yeah that would be a problem.

ivan-hc:
Please, try this with `avidemux` at https://github.com/mean00/avidemux2 and other "strange" github apps

Samueru-sama:
> Please, try this with `avidemux` at https://github.com/mean00/avidemux2 and other "strange" github apps

Tested it and it worked: 

image

The behaviour is not really that different than from the older template, it is just more efficient. So I really doubt you will found a case where the older template works but this one does not. 

>Also, complete this template with the $APP function you did in your custom templates, in order to solve the conflict between mine and yours templates... but without scale the lines, for example "version=" must be present at lines 19 and 39 (to prevent updating also the template.am module, see where sed -n appear in [template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am)).

what? xd

ivan-hc:
> what? xd

you understood correctly, I intend to entrust you with the drafting of the new templates

Samueru-sama:
> > what? xd
> 
> you understood correctly, I intend to entrust you with the drafting of the new templates

Sorry but I did not, what do you mean by conflict? but without scale the lines?

ivan-hc:
For "conflicts" I intend that function if "$APP" is empty... but maybe try to "exit" only the current working script, to prevent exiting the whole process of installation if other scripts are in the waiting list (i.e. by installing multiple apps at once).

And "without scaling lines", if you can, keep "version=" at lines 19 and 39... but if you can't, I'll correct [template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am), if needed. We must try to use the same template for all AppImages, using the option `-t` and the template number Zero.

Samueru-sama:
> For "conflicts" I intend that function if "$APP" is empty... but maybe try to "exit" only the current working script, to prevent exiting the whole process of installation if other scripts are in the waiting list (i.e. by installing multiple apps at once).

You mean this? `if [ -z "$APP" ]; then exit 1; fi` I just added anyway lol. (I thought you didn't want it from the last time). 

> And "without scaling lines", if you can, keep "version=" at lines 19 and 39... but if you can't, I'll correct [template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am), if needed. We must try to use the same template for all AppImages, using the option `-t` and the template number Zero.

Right now `version=` is on line 20?

ivan-hc:
OK, now change
```
mkdir /opt/$APP
cd /opt/$APP
```
to
```
mkdir /opt/$APP && cd /opt/$APP
```
so we can put the "version=" references to the original lines 19 and 39

Samueru-sama:
Ready I think

ivan-hc:
perfect

ivan-hc:
@Samueru-sama if you feel confortable, you can change these templates the way you want, in respect of the [template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am) module's rules

if we need to redraw the installation scripts we need to workthe easier way as possible

we can also try to implement infos about the version, the way @zen0bit intend in its [issue](https://github.com/ivan-hc/AM/issues/383)

Samueru-sama:
> @Samueru-sama if you feel confortable, you can change these templates the way you want, in respect of the [template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am) module's rules
> 
> if we need to redraw the installation scripts we need to workthe easier way as possible
> 
> we can also try to implement infos about the version, the way @zen0bit intend in its [issue](https://github.com/ivan-hc/AM/issues/383)

I will see if I can optimize the icon extraction part like I mentioned here: 

https://github.com/ivan-hc/AM/pull/398#issuecomment-2058998844

Then yeah I will look at adding the `pkgersion` thing.

Samueru-sama:
Btw @ivan-hc I'm taking a look again at the sample script and I need to know what does this do? 

```
app=$(echo $APP | cut -c -3)
if [ ! -e ./$APP.desktop ]; then 
	rm ./$APP.desktop; ./$APP --appimage-extract usr/share/applications/*$app*.desktop 
	mv squashfs-root/usr/share/applications/*.desktop ./$APP.desktop
fi
if [ ! -e ./$APP.desktop ]; then 
	rm ./$APP.desktop; ./$APP --appimage-extract share/applications/*$app*.desktop 1>/dev/null
	mv squashfs-root/share/applications/*.desktop ./$APP.desktop
fi
```

What appimage needs this? I don't understand how this is needed since the regular `--appimage-extract usr/share/applications/*.desktop ` does the same thing as `--appimage-extract usr/share/applications/*$app*.desktop`?

ivan-hc:
The variable "$app" only takes the .desktop file which name contain at least the firrst three letters of the name of the app (for example "mpv" or GNOME applications  that have names like org.gnome.appname.desktop

Sometime AppImages have multiple .desktop files.

This is an attempt to find the right .desktop file when we have multiple ones.

Samueru-sama:
> The variable "$app" only takes the .desktop file which name contain at least the firrst three letters of the name of the app (for example "mpv" or GNOME applications that have names like org.gnome.appname.desktop
> 
> Sometime AppImages have multiple .desktop files.
> 
> This is an attempt to find the right .desktop file when we have multiple ones.

Ok but regular `--appimage-extract usr/share/applications/*.desktop` will also extract those examples you gave no problems since it has the `*.desktop` wildcard. `*.desktop` will always match more files than `*$app*.desktop `

Multiple .desktop files can indeed be a problem, but that part of the script is not going to fix it because it first checks for `[ ! -e ./$APP.desktop ]` and that is always going to be true unless it is possible to make an appimage that doesn't have a top level .desktop file, which it may be a symlink but `[ ! -e ./$APP.desktop ]` doesn't check if it is a symlink so that part of the script will never even run.

ivan-hc:
@Samueru-sama yesterday I've done a test with "vlc-cmatomic", from https://github.com/cmatomic/VLCplayer-AppImage and I'v used the command
```
am -t vlc-unofficial
```
no icon was shown.

A reason because I had that scheme in the templates was also related to this old AppImage.

However, if I run the command
```
am -t vlc-cmatomic
```
to create the script, the icon is installed correctly but only because the icon "vlc-cmatomic.png" is already stored on portable-linux-apps.github.com

Samueru-sama:
> @Samueru-sama yesterday I've done a test with "vlc-cmatomic", from https://github.com/cmatomic/VLCplayer-AppImage and I'v used the command
> 
> ```
> am -t vlc-unofficial
> ```
> 
> no icon was shown.
> 
> A reason because I had that scheme in the templates was also related to this old AppImage. The icon here is just a symlink to the one stored deeper in the AppImage.
> 
> However, if I run the command
> 
> ```
> am -t vlc-cmatomic
> ```
> 
> to create the script, the icon is installed correctly but only because the icon "vlc-cmatomic.png" is already stored on portable-linux-apps.github.com

What's the issue? the new template should pull the icon from the appimage no problem since .DirIcon is an image. I just tested doing an --appimage-extract to confirm that.

ivan-hc:
try to install the script using `-i` (not as "vlc-cmatomic", use another name, for example "vlc-unofficial")

ivan-hc:
however, this is not a problem since the apps are stored in the database and the icons on the catalogue

Samueru-sama:
> try to install the script using `-i` (not as "vlc-cmatomic", use another name, for example "vlc-unofficial")

Alright I did this: 

`am -t vlc-unofficial` and gave it the link to the github. It created this: https://pastebin.com/MntijknA

then I ran `am -i ./vlc-unofficial` and installed the appimage, with a working icon.

Issue #397, opened 2024-04-16
zsh completion?


Ranko-Saotome:
I don't know if you take feature requests but I noticed you had bash completion think you can add completion for the zsh

ivan-hc:
I'm not familiar with ZSH at all

Samueru-sama:
> I'm not familiar with ZSH at all

I will take a look at this later in the day, I normally don't need this because zsh does completion from your zsh history as well. 

Something worth mentioning so I don't forget is that I need to check for `$ZDOTDIR` variable before patching `.zshrc`.

Samueru-sama:
@ivan-hc Alright I came out with something, tested and it works, I will make a new PR with these changes once you decide on what to do with the currently open PR #398 😅

ivan-hc:
> @ivan-hc Alright I came up with something, tested and it works, I will make a new PR with these changes once you decide on what to do with the currently open PR #398 😅

OK, let talk at that PR before we go on

ivan-hc:
With this commit we also have a new release https://github.com/ivan-hc/AM/pull/401

AM 6.5.1

Ranko-Saotome:
thanks for adding!

PR #400, opened 2024-04-16
Add zsh completion

Samueru-sama:
I don't know if there is a way to fix the lack of space in lines 438, 439, 440, because otherwise the space gets piped into `.zshrc. `

And if I try to individually `echo "" >>` each line to `.zshrc` there is the problem that if the last line in zshrc does not end in an empty line, it will put the text next to the existing text for the first line, breaking it.

Fixes https://github.com/ivan-hc/AM/issues/397

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -422,6 +422,21 @@ case "$1" in
 		$SUDOCOMMAND mv "$COMPLETIONFILE" "$COMPLETIONPATH/" 2> /dev/null
 		echo " Completion has been enabled!"
 	fi
+	if [ -f "$HOME/.zshrc" ] || [ -f "$ZDOTDIR/.zshrc" ]; then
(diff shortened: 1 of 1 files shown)


ivan-hc:
at line 425 you mention a "$ZDOTDIR", but where is the value?

ivan-hc:
> at line 425 you mention a "$ZDOTDIR", but where is the value?

also, I've tested it in Shellcheck and does not reports issues at line 425... is this some kind of known environment variable available system wide?

Samueru-sama:
@ivan-hc I've noticed that the space was added back in.

The reason it was originally like this: 

image

Was because otherwise if .zshrc ends on a line that contains text instead of an empty line, it will cause this issue and break line 46 in like in this test run of `--enable-completion` I just did: 

image

PR #415, opened 2024-04-18
add use_functions for aguments + small style fix

zen0bit:
I am not sure what's your default editor config (using spaces somewhere but mostly tabs)
Maybe we can set in scripts?

something like `# vim:tabstop=4:shiftwidth=4:expandtab`?

It should go on last line of script

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -519,3 +519,5 @@ case "$1" in
 	'version'|'-v'|'--version')                      echo "6.6.1";;
 	*)                                               exec "$AMCLIPATH" ;;
 esac
+
+# vim:tabstop=4:shiftwidth=4:expandtab
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -13,9 +13,9 @@ else
 fi
 
 if [ -n "$XDG_DATA_HOME" ]; then
-        DATADIR="$XDG_DATA_HOME"
+	DATADIR="$XDG_DATA_HOME"
 else
-        DATADIR="$HOME/.local/share"
+	DATADIR="$HOME/.local/share"
 fi
 
 function _am(){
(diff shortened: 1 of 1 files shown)


zen0bit:
How you debug your script?

I can't use `bash -x` on it...

ivan-hc:
nothing works, apart the single options

Istantanea_2024-04-18_11-16-31

ivan-hc:
> How you debug your script?
> 
> I can't use `bash -x` on it...

by testing the options one by one, on my PC

zen0bit:
if i want use bash -x am -h

no matter what
ask me to create new config for appman
I think AppMan/am should be totaly included. I see only difference in path and sudo
Both could be taked care in installation procces of apps itself
so am will run without sudo until sudo is rezlly needed than ask for it if not in appman mode

Samueru-sama:
Why does DATADIR have two tab spaces instead of just one?

image

zen0bit:
I forgot pass the arguments :stuck_out_tongue_winking_eye:

zen0bit:
Or you have different configuration? tell us...

zen0bit:
Program behavior has not changed

ivan-hc:
@zen0bit all works well, apart two things:
- your beloved `-f` now has options, so add a "$@" to made them work;
- this is a refactoring, you don't need to change the version, at least not with a "dot" (max 6.6-1)

The way I intend new versions:
- `N` is a release with radical and deep changes (as they were AM and AppMan merged in one, v5, and modularization, v6);
- `N.N` brings with it new option or an essential feature that has been improved (for example, last 6.6 is because updates are in parallel);
- `N.N.N` iprovements to the single options.

The three above can be annunced as "releases".

In this case your PR is nothing but a simple refactoring, so `N.N-N` its enough.

ivan-hc:
@zen0bit I'll fix the above for you, don't worry

zen0bit:
> @zen0bit I'll fix the above for you, don't worry

Ialso found, but you are quicker :+1:

PR #434, opened 2024-04-19
Fix btop and update libreoffice to new template.

Samueru-sama:
btop was symlinking a directory to my $PATH instead of the binary, I updated it and fixed it (it is likely that the other versions of btop also need fixing btw). 

Also updated libreoffice to the new template.


ivan-hc:
@Samueru-sama the "libreoffice" script if one of those I'm proud the more, thank you for having care of it

PS: are you aware that Type2 AppImages that use .zsync files to be updated, can still be updated if converted in Type3?

Istantanea_2024-04-19_15-31-14

and in the case of Libreoffice, we earn 25+ MB of disk space

PR #446, opened 2024-04-20
Use more functions

zen0bit:


ivan-hc:
@zen0bit I like the fact that `sudo` does not recognizes the command `appman` and that `am -i` does not require `sudo`

Istantanea_2024-04-21_10-38-30

this solves a big issue. Thank you

PR #457, opened 2024-04-21
FIrst batch of updated apps. 

Samueru-sama:


Samueru-sama:
Tell me how it goes so I can do the rest.

ivan-hc:
its all OK, but just add a 2>/dev/null in front of deadbeef abd deadbeef-devel downloads of .desktop and icon to prevent this

Istantanea_2024-04-21_22-42-45

ivan-hc:
I did this change yesterday

Samueru-sama:
> its all OK, but just add a 2>/dev/null in front of deadbeef abd deadbeef-devel downloads of .desktop and icon to prevent this
> 
> Istantanea_2024-04-21_22-42-45

But adding 2>/dev/null won't get rid of those 2 extra progress bars? Or are you ok with having multiple progress bars?

ivan-hc:
WAIT A MINUTE!

I forgot to remove the silencer to the installation command

Istantanea_2024-04-21_22-48-43

ivan-hc:
in my opinion and in my experience, you may left one command per line, so I can definitelly solve the issue by adding a filter with "grep" to each line starting with one of the commands that we see above

ivan-hc:
so get rid of that damn && and let each command have its own line, after this you've nothing more to do... we have nothing more to do

Samueru-sama:
> in my opinion and in my experience, you may left one command per line, so I can definitelly solve the issue by adding a filter with "grep" to each line starting with one of the commands that we see above

image

I really don't know why it shows text in your end. 

Not allowing && in scripts sounds like a terrible hack. I don't agree with that.

Samueru-sama:
> so get rid of that damn && and let each command have its own line, after this you've nothing more to do... we have nothing more to do

HEY WAIT I THINK I KNOW WHY IT HAPPENS.

image

It's a bashism lol. 

image

`&>/dev/null` has to be changed for `>/dev/null 2>&1`

ivan-hc:
> Not allowing && in scripts sounds like a terrible hack. I don't agree with that.

as far as I know, "&&" is just to save lines of code to merge them into one. How much does it cost you to dedicate a command to each line?

Samueru-sama:
> > Not allowing && in scripts sounds like a terrible hack. I don't agree with that.
> 
> as far as I know, "&&" is just to save lines of code to merge them into one. How much does it cost you to dedicate a command to each line?

No, `&&` only allows the next command to run if the previous command didn't fail. Which yeah in this case it matters little because --appimage-extract doesnt' exit with non 0 when it finds nothing. But for other stuff in can be critical and needed. 

I found the issue anyway, it is a bashism xd. 

Just quickly change `&>/dev/null` for `>/dev/null 2>&1` and test deadbeef again, that should fix it.

ivan-hc:
You follow the rules too much, when a bit of anarchy? I use these scripts from 3 years and never has all the issues you have "daydreamed" in the last month. Is it possible that the only linux system that strictly follows these security rules is installed on your PC?

Samueru-sama:
> You follow the rules too much, when a bit of anarchy? I use these scripts from 3 years and never has all the issues you have "daydreamed" in the last month. Is it possible that the only linux system that strictly follows these security rules is installed on your PC?

What you want is actually imposing more rules, I was baffled when the other day adding `" "` to a path breaks appman because you had a sed function that needed a space in the path. 

the more flexible the scripts are the better it will be in the future for everyone.

Samueru-sama:
> You follow the rules too much, when a bit of anarchy? I use these scripts from 3 years and never has all the issues you have "daydreamed" in the last month. Is it possible that the only linux system that strictly follows these security rules is installed on your PC?

Also the reason I wasn't able to replicate the issue is because on archlinux `sh` is a symlink to `bash` instead of `dash` like on debian btw.

ivan-hc:
I need you restore the template and the deadbeefs from main to dev, again.

I've finally solved everything and I want to release this new installation ui

PR #458, opened 2024-04-21
Update AM-SAMPLE-AppImage to be POSIX compliant 

Samueru-sama:
Test

image

Now it should be fully silent when using the dash shell (which is /bin/sh in debian).


Samueru-sama:
If you still get the sed error it might be because there is no .desktop file being extracted like I noticed yesterday btw. 

I didn't silence sed because that command should not fail and should notify when it does as indicates that the `.desktop` is missing.

PR #467, opened 2024-04-22
Sneaky bug and fix krita.

Samueru-sama:
Ok this fixes a sneaky bug in deadbeef-devel-appimage.

Also updated krita to the new template, see how good this new template is that the icon works now 😇

### DIFF
--- a/programs/x86_64/deadbeef-devel-appimage
+++ b/programs/x86_64/deadbeef-devel-appimage
@@ -33,7 +33,7 @@ cat >> /opt/"$APP"/AM-updater << 'EOF'
 APP=deadbeef-devel-appimage
 REPO="Samueru-sama/DeaDBeef-AppImage"
 version0=$(cat /opt/$APP/version)
-version=$(wget -q https://api.github.com/repos/$REPO/releases -O - | grep browser_download_url | grep -i appimage | cut -d '"' -f 4 | head -1)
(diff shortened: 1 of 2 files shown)


Samueru-sama:
Btw @ivan-hc is the last part of the script:

```
# MESSAGE
echo "
 $APP is provided by $SITE
"
```

Something that needs to be added to every new appimage? because it is not part of the new template.

ivan-hc:
this is an obsolete part used in old releases (I'm talking about 1.x, 2.x ... 3.x). Its totally incompatible with what we have now (in particular since this morning, with the 6.6.2 release

ivan-hc:
its task was to appear when the script was finished... now its no more so

if you see, there is also a piece "change permissions", it is needed to run the script as a standalone one, to have permissions.

All of this is removed by default using the patch in the install.am module

PS: please don't ask me to add the site at the end of each process. The option `-a` exists for this.

Samueru-sama:
> PS: please don't ask me to add the site at the end of each process. The option `-a` exists for this.

No no, I don't want that, just that I didn't know what to do when updating scripts that had that. 

From now on I will delete that part when updating a script.

ivan-hc:
what an obsolete database

Samueru-sama:
> what an obsolete database

If you ever decide to do a full rewrite of the database let me know, because some stuff has to be decided beforehand. 

For example I don't understand why the `.desktop` get installed by default on `/usr/share/applications` instead of `/usr/local/share/applications` like how the symlink is already in `/usr/local/bin` for example.

PR #477, opened 2024-04-23
Security: Specified the use of "AM" on multi-account systems

ivan-hc:
On systems with multiple accounts, only the user who installed "AM" and the root user will be able to use "AM" and all its options. Apps installed with "AM" will also be updateable and manageable by the system administrator.

Non-privileged users will instead be advised to contact the administrator to change permissions on the /opt/am directory or will be advised to use "AppMan" as an alternative.

Istantanea_2024-04-23_12-42-11

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -2,7 +2,7 @@
 
 DIR="$( cd "$( dirname "$0" )" && pwd )"
 CLI=$(basename "$0")
-AMVERSION="6.6.2"
+AMVERSION="6.6.2-1"
 
 SCRIPTDIR="$(if command -v xdg-user-dir &>/dev/null; then xdg-user-dir DESKTOP; else echo $HOME; fi)"
 
@@ -119,6 +119,20 @@ export AMCATALOGUEICONS="https://raw.githubusercontent.com/Portable-Linux-Apps/P
 # Determine system architecture and current user
 export arch="$HOSTTYPE"
 export currentuser="$USER"
+if [ "$AMCLI" == am ]; then
+	ampathowner=$(ls -ld /opt/am | awk 'NR==1 {print $3}')
+	if [ "$ampathowner" != "$currentuser" ]; then
+		echo ' 💀 ERROR: you are not the owner of /opt/am, so you cannot use "AM"'
+		echo '--------------------------------------------------------------------------'
+		echo ' The user that have installed "AM" on this system is "'"$ampathowner"'"'
+		echo ' you are only allowed to use the installed apps, not to manage them.'
+		echo '--------------------------------------------------------------------------'
+		echo ' Please, contact "'"$ampathowner"'" to change the ownership of /opt/am or'
+		echo ' consider using "AppMan" instead, see https://github.com/ivan-hc/AppMan'
+		echo '--------------------------------------------------------------------------'
+		exit
+	fi
+fi
 
 if [[ -f "$AMPATH"/new-repo.txt ]]; then
 	AMREPO=$(<"$AMPATH/new-repo.txt") # Utilizing command substitution with < for more efficient file reading
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ Initially the two projects traveled in parallel to each other, until version 5, 
 #### To be more detailed, here is an overview of how apps are installed by "AM" and "AppMan"
 
 Where `$PROGRAM` is the application we're going to install:
-- "AM" (ie the `am` command) installs programs and works at system level (i.e. for all the users). "AM" requires the `sudo` privileges but only to install and remove the app, all the other commands can be executed as a normal user. This allows multiple users of the same system to be able to use the same installed applications. This is what an installation script installs with "AM":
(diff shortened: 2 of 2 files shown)


Samueru-sama:
Looks ok but I wonder how permissions get handled with native packages for multi user systems, because as far as I know this isn't an issue in that case.

ivan-hc:
I've done my tests on a VM of Debian with 3 accounts:
- the main one, ivan, the one that have installed "AM"
- the other two (guest and guestroot) have the same permissions and are not in the "sudoers" (also I've tested them in sudo, root and ivan groups)

### Tests
- I've installed anydesk (appimage) and firefox (autoupdatable binary), the latter has also been manually replaced with an older version
- from the guest* accounts I get the message I've added, so I was unable to use "AM" on them
- both apps can be used from the guest* accounts
- firefox, if used from the guest* accounts, is unable to be updated, I get a notification from it about the availability of a newer version to be downloaded manually
- back to "ivan", "AM" works normally

ivan-hc:
another test I did (in the VM and on my host system) is to run "AM" as root user, and the result was out of any prevision:
- from "ivan" I've done `su` and entered the root shell
- all commands are working as they should with "AM"
- I'm exited from the soot shell and I was still able to have my permissions as "ivan" by editing/run commands etcetera... in all directories of "AM", also the ones of the installed apps

Samueru-sama:
What happens if AM doesn't do the `chown -R $currentuser /opt/$APP` after installing an app?

ivan-hc:
I've also removed the filter and added the user "guestroot" to sudoers, "AM" has no RW permissions.

ivan-hc:
> What happens if AM doesn't do the `chown -R $currentuser /opt/$APP` after installing an app?

can't update it or do anything in that directory (without "sudo")

Samueru-sama:
> > What happens if AM doesn't do the `chown -R $currentuser /opt/$APP` after installing an app?
> 
> can't update it or do anything in that directory (without "sudo")

Then it is not that different than pacman or apt where you need to use sudo to update the system applications. But every user can still launch those applications. 

Assuming this multi-user permission goes away by never doing the `chown -R $currentuser /opt/$APP` then I think this is the proper solution. AppMan is already there for people that can't use sudo, and AM already needs sudo to install new packages anyway.

ivan-hc:
> > > What happens if AM doesn't do the `chown -R $currentuser /opt/$APP` after installing an app?
> > 
> > 
> > can't update it or do anything in that directory (without "sudo")
> 
> Then it is not that different than pacman or apt where you need to use sudo to update the system applications. But every user can still launch those applications.
> 
> Assuming this multi-user permission goes away by never doing the `chown -R $currentuser /opt/$APP` then I think this is the proper solution. AppMan is already there for people that can't use sudo, and AM already needs sudo to install new packages anyway.

It's correct. AM uses sudo only to install and remove the apps (or enable/disable bash-completion).

Suppose someone wants to install AM on many pc into a school, the students are allowed only to use the app, they can't install anything else, or do updates... this is a task of the administrator.

ivan-hc:
And with or without this change, the only one able to manage "AM" and all the installed apps is always the owner of /opt/am

This PR oly adds a method to show a message if the user that is using "AM" is not the one that hav installe it... or have installed any other app with it.

ivan-hc:
We can workaround it by changing the permissions only to the users with a group, but I've already tried to do chmod -R 0777 without success.

ivan-hc:
@Samueru-sama so, you say it is better I leave this or I should do something else? What you suggest?

Samueru-sama:
> > > > What happens if AM doesn't do the `chown -R $currentuser /opt/$APP` after installing an app?
> > > 
> > > 
> > > can't update it or do anything in that directory (without "sudo")
> > 
> > 
> > Then it is not that different than pacman or apt where you need to use sudo to update the system applications. But every user can still launch those applications.
> > Assuming this multi-user permission goes away by never doing the `chown -R $currentuser /opt/$APP` then I think this is the proper solution. AppMan is already there for people that can't use sudo, and AM already needs sudo to install new packages anyway.
> 
> It's correct. AM uses sudo only to install and remove the apps (or enable/disable bash-completion).
> 
> Suppose someone wants to install AM on many pc into a school, the students are allowed only to use the app, they can't install anything else, or do updates... this is a task of the administrator.

Sounds like more trouble just to give the ability of non privileged users to update apps tbh.

>so, you say it is better I leave this or I should do something else? What you suggest?

I would just get rid of all the chown and require sudo when updating.

ivan-hc:
> I would just get rid of all the chown and require sudo when updating.

so "AM" become a package manager like the others if we put "sudo" everywhere. Its not the goal of having it.

Samueru-sama:
Also @ivan-hc you can still let users update without doing the chown, You can set up rules so that certain commands don't need sudo to be run, I use for my cpu-scheduler scripts for example. Much simpler and doesn't have this issue.

ivan-hc:
@Samueru-sama the fact is that to install "AM" and apps trhough it you need "sudo", so everything is installed have no RW permissions in /opt/*/

ivan-hc:
@Samueru-sama this seems to be enough to add all permissions to all into a directory owned by the administrator
```
currentuser=$(who | awk '{print $1}')
chown -R $currentuser /opt/am 2> /dev/null
find /opt/am -exec chmod 777 {} \; 2> /dev/null
```
"$currentuser" is still needed to the system administrator not to use "sudo"

I've also added something similar in install.am and worked, I was able to install apps in both the privileged users, and not in the unprivileged one

ivan-hc:
This instead is enough to give ownership to root but RW to everyone:
```
chown -R "$USER" /opt/am 2> /dev/null
find /opt/am -exec chmod 777 {} \; 2> /dev/null
```

PR #497, opened 2024-04-26
Small improvement and more cleanup EDIT: FIX APP-MANAGER

Samueru-sama:
'$LOGNAME' does the same thing as the who | awk etc that was being used before. It also doesn't change when using sudo like whoami does: 

```
~/ echo "$LOGNAME"     
samuel
~/ sudo echo "$LOGNAME"
samuel
~/ whoami              
samuel
~/ sudo whoami         
root
```

https://lists.ubuntu.com/archives/ubuntu-users/2015-July/281503.html

'$USER' also works but isn't part of POSIX. 

---------------------

Cleaned up the aarch64 install scripts. 

The i686 scripts also need to be fixed, but I'm still on the camp that it is a waste of time to maintain those 😆

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -182,7 +182,7 @@ function _am_security_check() {
 	# Prevent non-privileged users from using certain options
 	function _user_not_in_sudoers() {
 		if [ "$AMCLI" == am ]; then
-			if ! groups "$USER" | grep -qw sudo; then
+			if ! groups "$USER" | grep -qw root; then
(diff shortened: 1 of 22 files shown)


Samueru-sama:
@ivan-hc The `function _user_not_in_sudoers` was still broken 💀

I fixed it already, the grep was checking for sudo instead of root.

ivan-hc:
thank you @Samueru-sama

PR #525, opened 2024-04-28
Update INSTALL to a wget2 compatible grep

### LINKED ISSUE #496: Fedora 40 KDE
Hi,
After the migration from Fedora 39 to 40, AM does not install correctly. When I run `am -h` or any other command, it says `modules` are missing.
What am I overlooking?

Samueru-sama:
Test: this method works with both wget and wget2

image

Also made sure that it installs am using both wget and wget2. Fixes #496 for good.


ivan-hc:
Let talk at https://github.com/ivan-hc/AM/discussions/524

PR #528, opened 2024-04-29
Use "jq" to handle JSON files / check if wget 1.x is in use

ivan-hc:
This change will help handle the transition to "wget2" from "wget" version 1.x

Fix https://github.com/ivan-hc/AM/discussions/524

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -156,7 +156,7 @@ function _am_security_check() {
 	function _am_dependences_check() {
 		# Check for essential commands required by the application
 		missing_deps=()
-		for name in cat chmod chown curl grep sed wget; do
+		for name in cat chmod chown curl grep jq sed wget; do
 			if ! command -v "$name" &>/dev/null; then
 				printf "\nERROR: command \"%s\" not found, strongly needed by \"%s\"!\n" "$name" "$AMCLI"
 			missing_deps+=("$name")
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="6.6.3.1"
+AMVERSION="6.6.4"
 
 # Determine the name of this script and its working directory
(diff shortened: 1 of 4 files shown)


ivan-hc:
@Samueru-sama nothing else?

Samueru-sama:
> @Samueru-sama nothing else?

The only detail is that you are still parsing json data instead of using jq to give you the links directly, however it is unlikely that it will be break but keep that in mind.

ivan-hc:
@Samueru-sama Regarding patches for "wget", the options of "wget" version 1.x and "wget2" are very different from each other.

"wget2" uses a loading bar by default, while "wget" has many more options, which allow us to customize the interface, for example using "dots" (which "wget2" no longer does). Such options, if added to "wget2", does not recognize them, and interrupts the download miserably.

These are two completely different programs, one is not the counterpart of the other, but a "substitute". When updating, a pre-existing program is generally enriched with new features. Among the listed options, only "-q" and "--no-verbose" have in common (and the latter acts differently), while "--progress" in "wget2" is only "bar", "dot " is not yet supported.

Finally, "jq" is only used for api.github.com, some portable apps, for example "firefox", also include a "-O -", and would show an error message from there.

Unfortunately, there are a lot of scripts, especially those that work with api.github.com, and editing them all would be a pain.

There is also the fact that each script can contain unique keywords to select the exact file.

As for the additional commands, in this case, it's just an "aesthetic" problem.

For the rest, the downloads all worked, without any problems.

Sure, adding a dependency like "jq" hurts... but in the past I've worked around many more dependencies, making them optional (zip, tar, ar, zsync... even xterm). In this case we are talking about a particularly useful tool for what we need to do, in view of a future adoption of wget2 (hoping it is not too premature, as I explained on [Reddit](https://www.reddit.com/r/Fedora/comments/1cf5jpu/switching_from_wget_to_wget2_suddenly_in_fedora/)).

PR #529, opened 2024-04-29
add jq

Samueru-sama:
👀


Samueru-sama:
@ivan-hc Remember that `function _am_security_check` needs to allow for jq to be installed. Something like this crude example can be put right before the function runs: 

```
if ! command -v jq &>/dev/null; then
	echo "dependency jq not detected on the system, want to use am to install jq?"
	am -i jq
fi
```

ivan-hc:
it would be enough to patch the installation scripts by adding your `sed 's/[()",{}]/ /g; s/ /\n/g'` instead

in that case the dependence is no more needed

however I've failed in implementing it via sed

PR #538, opened 2024-05-02
Cleanup

zen0bit:
Not everything cleaned up yet

I am doing just what's easy to do..

Rest will come later...

leave it here for me a will be constantly improving this.

We can merge this?

I will be doing changes "on the go"

## Feel free to test this


nazdridoy:
@zen0bit since you are working on refactoring. i've a few suggestions for you.

#!/shebang
- declare globalVariables
- define functions (any and all function)
  - make these so that these can be unit tested.
  - DO NOT define function inside another function
    - if it HAS TO BE a nested function define it outside of the parent function (above) and call it.
    - if child functions need  any local variable from their parent functions, first declare local variable and then call the child function
  

main(){

- prioritize case statement (its faster because of hash table) 
- if..else are ok but put most likely case at top and break out of clause if one pass or fail your check. (if you think carefully you can save significant run time)
- for / while loop: avoid as much as possible (these are crazy slow)
- DO NOT define function inside another function.
}


@ivan-hc please try to follow this guideline for future module. if you have any confusion, feel free to ask.

zen0bit:
> if you have any confusion, feel free to ask.

What's purpose of case with only one option? (Seen a lot in AM)
(I prefer iffi for that)

nazdridoy:
you should not use case for everything...

APP-MANAGER
```diff
@@ -14,6 +14,13 @@ CLI=$(basename "$0")
 arch="$HOSTTYPE"
 currentuser="$USER"
 
+# Colors
+RED='\033[0;31m'; LightBlue='\033[1;34m'; Green='\033[0;32m'
+
+web_get() {
```
ivan-hc, line 20: This is a bash script, this line should start with "function"

APP-MANAGER
```diff
@@ -44,24 +48,6 @@ function _no_sudo() {
 	fi
 }
 
-_no_sudo
```
ivan-hc, line 47: if you can, try to leave the function as one of the first and standalone, so whatever the user want to do, him can never use sudo in the command.

As usual, many people still use sudo for everything. This function, if starts at the really begin, prevents this syntax.

modules/install.am
```diff
@@ -1,320 +1,381 @@
...
-								echo -e ' ⚠️ WARNING: '$(echo '"'$arg'"' | tr a-z A-Z)' MAY NOT BE UPDATABLE WITHOUT "ZSYNC".\n\n THE APP WILL STILL BE INSTALLED, BUT INSTALLING "ZSYNC" IS RECOMMENDED!\n'
-							fi
-						fi
-					fi
-
-					# If you use the --latest option, add the "latest" flag to scripts that point to a very common "release"
-					if grep -q 'force-latest' $AMPATH/.cache/install-args; then
-						sed -i 's#releases -O -#releases/latest -O -#g' ./$arg
-					fi
+web_get() {
```
ivan-hc, line 3: See https://github.com/ivan-hc/AM/pull/538/files#r1588605326

Also, being this function already implemented in APP-MANAGER, it will be called also into the modules. You don't need to put it here too.

modules/sync.am
```diff
@@ -1,74 +1,64 @@
 #!/usr/bin/env bash
 
+web_get() {
```
ivan-hc, line 3: see https://github.com/ivan-hc/AM/pull/538/files#r1588607815

APP-MANAGER
zen0bit: I aware of this. Some bash require it some versions not
Not really sure which, but better to use that everywhere

APP-MANAGER
zen0bit: I am not started yet...

modules/install.am
zen0bit: Note yet exported from APP_MANAGER...

zen0bit:
I not done anything yet just moved functions and created web_get

But that's not even begining

Samueru-sama:
> I not done anything yet just moved functions and created web_get
> 
> But that's not even begining

Can you try to make some of the modules use sh instead of bash?

zen0bit:
> > I not done anything yet just moved functions and created web_get
> > But that's not even begining
> 
> Can you try to make some of the modules use sh instead of bash?

Never scripted with sh in mind
So don't know what's about.

I know that is differ between bash and sh but...
What i can't use under sh?
Don't know can try
I like learn new things

But don't know where to start...
What's will be limitating factor...

zen0bit:
Here is my source of inspiration:
https://github.com/vitalets/github-trending-repos/issues/35

Samueru-sama:
> > > I not done anything yet just moved functions and created web_get
> > > But that's not even begining
> > 
> > 
> > Can you try to make some of the modules use sh instead of bash?
> 
> Never scripted with sh in mind So don't know what's about.
> 
> I know that is differ between bash and sh but... What i can't use under sh? Don't know can try I like learn new things
> 
> But don't know where to start... What's will be limitating factor...
> 
> But i still don't know bash properly so I think I shouldn't mess my bash knowledge with sh yet...

The only time I found myself needing to use bash instead of sh was with the apprun of my android-tools appimage, as I needed to filter "$@" after "$1" and with bash it is as simple as using`"${@:2}"`. Besides that I've been able to do everything with sh instead. And I don't think am or its modules do something that needs the example I just gave. 🤔

I asked you because iirc I've seen you before use `>/dev/null 2>&1` which is POSIX instead of the `&>/dev/null` that is used with bash hehe.

zen0bit:
> I asked you because iirc I've seen you before use `>/dev/null 2>&1` which is POSIX instead of the `&>/dev/null` that is used with bash hehe.

😀 If is like that I can try. sh is more common than bash?

That /dev/null is more copy paste problem. Since I don't know what use where🤫

PS: But I will try, just to see how it will goes...

Samueru-sama:
> > I asked you because iirc I've seen you before use `>/dev/null 2>&1` which is POSIX instead of the `&>/dev/null` that is used with bash hehe.
> 
> 😀 If is like that I can try. sh is more common than bash?
> 
> That /dev/null is more copy paste problem. Since I don't know what use where🤫
> 
> Oh and that POSIX standart what's that really is? Why should I stick/or not with it? 🤷 PS: But I will try, just to see how it will goes...

Normally I try to make my scripts posix compliant unless I need a specific feature from bash. The only advantage is that the scripts are technically more universal and also considerably faster (which doesn't matter much as am isn't slow as is lol).

ivan-hc:
> Normally I try to make my scripts posix compliant unless I need a specific feature from bash. The only advantage is that the scripts are technically more universal and also considerably faster when dash is the sh (which doesn't matter much as am isn't slow as is lol).

At the beginning AM was written in SH, but then I needed some functions and commands (which I don't remember now) that only worked in BASH, and hence the conversion... it seems like AM v2 or v3, I don't remember.

But it must be said that I wasn't as familiar with scripts as I am now, so in this PR you might be able to find solutions that I didn't find.

For now, the main differences that I found and that come to mind are that functions in BASH must start with `function`, not in SH... then `echo -e` in BASH is not necessary in SH , `echo` is enough. I can't think of anything else right now.

ivan-hc:
@zen0bit why "function web_get" is also in the modules? If it is in APP-MAGARER, its enough to be called by the modules

zen0bit:
@Ivan look here:
[`f6c25a4` (#538)](https://github.com/ivan-hc/AM/pull/538/commits/f6c25a417e006158afaebb00250fb90ddc2d11af)

ivan-hc:
> @ivan look here: [`f6c25a4` (#538)](https://github.com/ivan-hc/AM/pull/538/commits/f6c25a417e006158afaebb00250fb90ddc2d11af)

this function belong to @nazdridoy , it is needed to check if the github key is used or if its needed a new one

nazdridoy:
> @ivan look here: [`f6c25a4` (#538)](https://github.com/ivan-hc/AM/pull/538/commits/f6c25a417e006158afaebb00250fb90ddc2d11af)

> # Is this used??

it should be.

ivan-hc:
@zen0bit here https://github.com/ivan-hc/AM/pull/548#issuecomment-2094203317 you asked me what are the modules that can edit/change the installations scripts. Well, instal.am, and partially deownload.am. The latter in fact only downloads the scripts... but contains an option "--convert" that mades them compatible with a local installation for AppMan.

Since we removed the $currentuser variable from all the installation scripts (thankd to @Samueru-sama ) this is the only way we have to install the applications to made them installable as standalone ones, in the $HOME. I had to change it to convert the scripts for a system wide installation, also to made them compatible with systems including RO filesystems (see immutable ones and those with BTRFS)... but since we have lost the $currentuser variable, now they can be only installed with AM system-wide. So yes, in download.am this feature is deprecated. The only module that edits the installation scripts is install.am, nothing else.

ivan-hc:
@zen0bit is this ready? Should I merge to https://github.com/ivan-hc/AM/tree/refactoring?

ivan-hc (commented):
in this moment its a bit difficult to navigate in the web ui due to the big amount of files.

I think I should sync  "refactoring" with "main", and then you must push again modules, INSTALL and APP-MANAGER... but I don't want you lost your commits.

its there a more convenient and easy way to do so?

ivan-hc:
I don't know what this PR still lacks, we can discover this by testing it, no other ways

PR #549, opened 2024-05-04
Update APP-MANAGER

Samueru-sama:
This simple trick does the same thing as the old if statements that were being used before. It is also POSIX so it won't be an issue in the future just in case you wonder.

@ivan-hc `$CACHEDIR` is not being used right now but it will be used in the future, I'm waiting for zenobit to finish the changes they are doing so I can start on that.

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -45,7 +45,7 @@ for module in $MODULES; do
 done
 
 # ENABLE NON-ROOT PERMISSIONS TO THE MAIN DIRECTORY FOR THE CURRENT USER
-chown -R "$currentuser" /opt/am 2> /dev/null
+chown -R $currentuser /opt/am 2> /dev/null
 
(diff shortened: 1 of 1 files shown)


ivan-hc:
I would suggest to let AppMan use $CACHEDIR too. I'd like to remove the $AMPATH/.cache directory in the next future.

Let @zen0bit doing its work first and we will see

Samueru-sama:
> I would suggest to let AppMan use $CACHEDIR too. I'd like to remove the $AMPATH/.cache directory in the next future.
> 
> Let @zen0bit doing its work first and we will see

Ok @zen0bit do not turn what I'm deleting into a function as it will not be used anymore, just in case.

zen0bit:
I think I am already done with APP-MANAGER...

PR #551, opened 2024-05-04
Dev

ivan-hc:


ivan-hc:
New issue: INSTALL can't allow the user having privileges... I've not seen yet, but its quite possible that its because $currentuser has quotes

ivan-hc:
issue 3: I can't remove am with `am -R am`

ivan-hc:
Now I started testing the modules one by one, but in first I have enabled the "dev" repository with `am --devmode-enable` and then used "sync"... this come from "main" and obviously works.

Then I've removed the /opt/am/betatester file and started the command `am --devmode-enable` again... the file has been created... but the message appears like this

Istantanea_2024-05-04_20-50-20

its not a problem, we can fix this... I only hope all other messages appear as they should

ivan-hc:
For this PR I want to alternate with you with the tests, otherwise they will never end.

Proceed in order:
```
sudo ./INSTALL
am --devmode-enable
am -s
am --devmode-disable
am --devmode-enable
```
and from there you will do all the testing reporting bugs and fixes as you tackle new problems.

@zen0bit @nazdridoy @Samueru-sama  Have a good trip.

PS: unfortunately this is the only alternative branch from which we can download modules besides "main", in the future I will certainly extend this possibility to all branches... but for now, we will work with what we have available.

zen0bit:
i am not yet added arguments anywehere yet (excepet) web_get

Going to look to it...

INSTALL
```diff
@@ -1,5 +1,9 @@
 #!/bin/sh
 
+web_get() {
+	wget -q
+}
+
```
zen0bit, line 6: 1st problem

INSTALL not check dependencies, will try run installation even without curl/wget

zen0bit:
As you can maybe figured out
I made fresh VM for testing...
Nothing is installed except xfce and some dev apps which i downloaded

Since cached this first bug right on start.

Thinking that will be better make test environment reproducible.
So I will push also CI with tests...

So Test will be easily reproducible to anyone out there.
We can trigger this action easily any time...

I will do the CI part, but anyone could try make TEST app script template.

TEST app template
Should contain everything what is same in all(most) app scripts
Will install just itself

But after any action (from TEST app) will show PASS/FAIL message

I hope that my inensions are clear...
Any questions?

INSTALL
ivan-hc: you need to add "$@" to expand the arguments, as I did at https://github.com/ivan-hc/AM/pull/551/commits/a05f4bc34cacbc49738446995f17aa61d504f88d

nazdridoy:
> I will push also CI with tests...

CI auto test every PRs is a crying need..... but for that to work properly every script must be written so these can be unit tested. functions need to be written in a way that they are not heavily dependent on any thing and can can be tested with test data. 


@ivan-hc  i propose adding a testFile directory with test template for every function. if any new function is written at some point in future that PRs also has to include a test file to test those function. in summery every function must be and will be auto tested before these can be merged with main.

if you or @zen0bit or any other want to give it a go, go ahead. otherwise i'll  write first batch of test files next month... these will compliment CI test.

PR #554, opened 2024-05-05
CI

zen0bit:
To see tests results on working AM


zen0bit:
This could be merged..

Hello world will be changed to my script bit later...

ivan-hc:
To test all apps on each push is boring. May this be scheduled once at week instead... or scheduled for each alphabetic letter?

With this approach, you will start always from A and for each push you will never reach Z and keep having security alerts about failed jobs.

ivan-hc:
Also, this PR keeps bombing me of notifications. Should I turn off notifications on PS and security alerts then?

zen0bit:
> Also, this PR keeps bombing me of notifications. Should I turn off notifications on PR and security alerts then?

I am done here for now..

zen0bit:
> Also, this PR keeps bombing me of notifications. Should I turn off notifications on PR and security alerts then?



> To test all apps on each push is boring. May this be scheduled once at week instead... or scheduled for each alphabetic letter?
> 
> With this approach, you will start always from A and for each push you will never reach Z and keep having security alerts about failed jobs.

Just demo version

Can do whatever you like I think...

Samueru-sama:
> In appsuite is test for 1 random app...

This is amazing, will come very handy!

zen0bit:
My idea is:
Run test on app if app is changed (triggered by change itself)
Run all test once (maybe week)
If AM changes, trigger Appsuite (test all arguments, with random app)
Maybe we can later split tests by modules? To trigger just for changed module...

Something like this?

ivan-hc:
> My idea is: Run test on app if app is changed (triggered by change itself) Run all test once (maybe week) If AM changes, trigger Appsuite (test all arguments, with random app) Maybe we can later split tests by modules? To trigger just for changed module...
> 
> Something like this?

OK, its better

ivan-hc:
For me is better if tests are done on changes related to one or more scripts in /programs/x86_64, but only on those that are been changed, on push.

For now you can start sunning one job for all apps and take note of the failed ones.

Then, change the workflow in a way that it can check only the ones we are gonna change.

A full check of all installation scripts should be monthly, so we have more space and time to test the ones that must be reviewed.

zen0bit:
exactly my plan

Artifacts with some logs will be saved (as zipped folder) if job not fail...

Will see after this finish...

ivan-hc:
About tests on all apps... maybe is better to wait bug reports

A full check may include a lot of time, also due to some apps like voicevox that is several gigabites of size... or all kdegames apps (about 40-41) that are one app... but for the latter, a list already exists in install.am
```
	ARGS=$(tail -n +2 $AMPATH/.cache/install-args)
	METAPACKAGES="kdegames kdeutils libreoffice nodejs platform-tools"

	for arg in $ARGS; do

		function _install(){
			echo ""
			case $arg in

			'--debug'|'--force-latest')	return 0;;
			*) 
				function _install_arg(){
					chmod a+x ./$arg

					# Check if the installation script is a metapackage (example one of the 40+ kdegames scripts)
					for metapackage in $(echo "$METAPACKAGES"); do
						if grep -q "$metapackage" ./$arg 2> /dev/null; then
							if test -d $APPSPATH/$metapackage/tmp; then
								$SUDOCOMMAND $APPSPATH/$metapackage/remove 2> /dev/null
							elif test -d $APPSPATH/$metapackage; then
								echo ' ◆ '$(echo '"'$arg'"' | tr a-z A-Z)' IS PART OF "'$(echo "$metapackage" | tr a-z A-Z)'", ALREADY INSTALLED'; return 0
							fi
						fi
					done
```

ivan-hc:
@zen0bit I have a suggestion to speed up testing of apps to install: boycott scripts containing the "api.github.com" keyword.

As long as the scripts are linked to an app hosted on github.com, we are 99% sure that this will work. At most, if something goes wrong, we should wait for user feedback.

The real obstacle are apps that aren't hosted on Github. Run a check only on those scripts.

Currently the scripts containing the keyword "api.github.com" are 1707.

So the scripts left are 279. Check only them instead.

zen0bit:
> @zen0bit I have a suggestion to speed up testing of apps to install: boycott scripts containing the "api.github.com" keyword.
> 
> As long as the scripts are linked to an app hosted on github.com, we are 99% sure that this will work. At most, if something goes wrong, we should wait for user feedback.
> 
> The real obstacle are apps that aren't hosted on Github. Run a check only on those scripts.
> 
> Currently the scripts containing the keyword "api.github.com" are 1707.
> 
> So the scripts left are 279. Check only them instead.

Can you list them somewhere?

PS: Maybe include in AM somewhere list only (not from gh?) To not have to check all if not from gh..

ivan-hc:
> > @zen0bit I have a suggestion to speed up testing of apps to install: boycott scripts containing the "api.github.com" keyword.
> > As long as the scripts are linked to an app hosted on github.com, we are 99% sure that this will work. At most, if something goes wrong, we should wait for user feedback.
> > The real obstacle are apps that aren't hosted on Github. Run a check only on those scripts.
> > Currently the scripts containing the keyword "api.github.com" are 1707.
> > So the scripts left are 279. Check only them instead.
> 
> Can you list them somewhere?
> 
> PS: Maybe include in AM somewhere list only (not from gh?) To not have to check all if not from gh..

You may setup this in the workflow run directly.

For example:
- download the content of /programs/x86_64
- use an "if" condition to remove all files containing the keyword "api.github.com"
- run the command `am -i $(ls ./programs/x86_64/ | tr '\n' ' ')` so you can install all the apps at once

zen0bit:
Continue in #565

PR #581, opened 2024-05-20
Gron time

Samueru-sama:
It's a shame jq ended up winning these two 🥲


ivan-hc:
think that I was thinking of get rid of api.github.com... or at least I'd like to implement something similar to torsocks or a sort of VPN to increase API rate limit... see https://github.com/ivan-hc/AM/discussions/578

PR #588, opened 2024-05-22
Improve speed of -a / about, -l / list and -f / files

ivan-hc:
The function to check the versions of installed applications takes a long time. If the "version-args" file already exists, you don't need to do it all over again. You might as well just create it:
- when apps are installed;
- when updates are made;
- when the file is not present.

In the latter case, a slowdown may occur the first time one of the affected options is launched, but all subsequent times only the existing file will be relied on.

### DIFF
--- a/modules/files.am
+++ b/modules/files.am
@@ -82,6 +82,7 @@ function _files() {
 
 if test -f "$AMPATH"/.cache/remove-args; then
 	rm -f "$AMPATH"/.cache/files-args*
+	rm -f "$AMPATH"/.cache/remove-args
 fi
 if [ "$2" = "--less" ]; then
 	cd "$APPSPATH" &&


ivan-hc:
The function to check the application type also takes a long time. The file listing all installed apps will also be retained, but will be regenerated if any applications have previously been removed.

Samueru-sama (commented):
Couldn't this have been a single `rm -f "$AMPATH"/.cache/*-args*` ?

ivan-hc:
or `rm -f "$AMPATH"/.cache/*`

ivan-hc:
I can also add this in remove.am

ivan-hc:
...the more time passes, the more I forget how many details have been studied in this application to work as best as possible, @Samueru-sama , proving that there is no refactoring, mine or anyone else's, that can compensate for the work I did to make it what it is now.

in remove.am a file named "unavailable-args" is created to list all applications that you have tried to remove and that instead arenot available.

if we try to remove the content of "$AMPATH"/.cache at the end of the process, we have a message like this:

Istantanea_2024-05-23_01-02-22 png

while instead it should look like this:

Istantanea_2024-05-23_01-03-25 png

so its better if we leave as it is now.

PS: on purpose, I also solved the problem of the table title, which disappeared after I made the changes to "AM" this morning. I hadn't even noticed.

Samueru-sama:
> or `rm -f "$AMPATH"/.cache/*`

Yeah didn't mention that one because they will remove everything in cache including something that might be needed idk. 

`*-args*` will instead remove all files that match that expression.

ivan-hc:
> > or `rm -f "$AMPATH"/.cache/*`
> 
> Yeah didn't mention that one because they will remove everything in cache including something that might be needed idk.
> 
> `*-args*` will instead remove all files that match that expression.

:point_down: 

unavailable-args

:point_up:

PR #598, opened 2024-05-25
Update mpv

Samueru-sama:
🫡

### DIFF
--- a/programs/x86_64/mpv
+++ b/programs/x86_64/mpv
@@ -1,114 +1,92 @@
 #!/bin/sh
 
 APP=mpv
-REPO="ivan-hc/MPV-appimage"
+SITE1="Samueru-sama/mpv-AppImage"
+SITE2="ivan-hc/MPV-appimage"
 
-# CREATE THE FOLDER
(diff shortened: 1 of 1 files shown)


Samueru-sama:
Btw @ivan-hc take a look at running linux deploy on top of the already built archimages before running appimagetool on them, linuxdeploy uses something call `strip` on the libraries that reduces the size of the appimage. It might reduce it a lot in this case.

ivan-hc:
> Btw @ivan-hc take a look at running linux deploy on top of the already built archimages before running appimagetool on them, linuxdeploy uses something call `strip` on the libraries that reduces the size of the appimage. It might reduce it a lot in this case.

thanks but the archimage is a different kind of appimage, it is a minimal Arch Linux container (JuNest) with all the libraries needed for the app. If I use linuxdeploy, it will get only the files of the app, and the libraries rely only on the GLIBC version of Arch Linux... so the app is no more suitable for old distributions. In brief, Archimage is a portable container and it needs its own system files and libraries to work (bubblewrap, glibc... pulseaudio...).

Issue #612, opened 2024-06-01
AM asking for a directory to store local apps. 


musicfunfan:
So i updated am today. And i get this massage on the terminal. 
![screenshot](https://github.com/ivan-hc/AM/assets/86926443/cb6206de-0bfb-48e0-b415-ec003e956da6)
I do not understand what is the purpose of this file. I just Added in the file path ``$HOME/Applications`` but i do not know. What is the purpose of this directory?

Samueru-sama:
Omg not again lol, we had this issue a week ago or so iirc

musicfunfan:
So this is a bug ? Because i do not use ``appman``. I use [topgrade](https://github.com/topgrade-rs/topgrade) and i do not know what command that use to update AM. 
According to [this](https://github.com/topgrade-rs/topgrade/blob/main/src/steps/os/linux.rs#L580).
![screenshot](https://github.com/ivan-hc/AM/assets/86926443/3c1b1273-64e9-480c-bb7c-fc590b8403b1)
The command topgrade runs for am is ``am -u`` so... I thing this is a general problem. 
This is what i have on my ``$HOME/Applicatioons`` Directory. 
![screenshot](https://github.com/ivan-hc/AM/assets/86926443/dc9deaef-c4a5-4a4b-9bf8-eaa07dcf8f26)
But i do not thing that i ever had appman installed. So this is a bug?

Samueru-sama:
Yes it is a bug, appman and am are the same thing, just different functions inside the APP-MANAGER script.

musicfunfan:
Ok, i was just looking if i did something wrong. Because ,most of the time... i was doing something wrong 😉. Any way. Thank you for the feedback...  I thing the appimages updated so more or less AM still working.

ivan-hc:
Please, can you remove the directory `$HOME/.config/appman` and update "AM" again?

I need to understand if this is a problem that affects who switches from from a previous release to 6.8.4 only or it still happens.



PS: the behaviour you faced is the same I've seen when trying to update "AM" with `am -u` or `am -s`, it seems that `-s` does not like to be included as a command in external modules, as it is in update.am... you've said that this is happened with topgrade.

ivan-hc:
I've just updated everything using topgrade and everything works without problems... so this bug is related to a previous version switching to the latest one. New users of "AM" should not face this issue.

ivan-hc:
[...]

Issue #600, opened 2024-05-25
Panic!-Please, write the name or the path of your custom application's folder? 


valtoree:
Hi
Sorry I dont understand, I just install apps either via apt, flatpak and now am -i to install apps, so what do I add here?
Sorry if dumb question.

ivan-hc:
You can read the list of the available apps with `am -l` and if you want to search an appname using keywords, try (for example) `am -q video editor`

To install an application is the same thing, write the lowercased name of the app you want (if it is in the list), for example:
```
am -i firefox brave-appimage chromium avidemux kdenlive
```
this will install these five apps (three web browsers and two video editors).

What is the app you're looking for?

Samueru-sama:
> Hi Sorry I dont understand, I just install apps either via apt, flatpak and now am -i to install apps, so what do I add here? Sorry if dumb question.

Looks like you are using AppMan and it is asking you where you want your user applications to be installed. 

Just type the name of the directory in home where you want the appimages to be installed. Most people use `~/Applications`

valtoree:
Thanks for quick reply, i usually do a 'am -u' everyday. So no app install just update and got prompted this.

ivan-hc:
are you talking about the `--user` option or the "AppMan Mode" or "appman"?

The prompt helps you to customize the path where you want to install all the apps in your $HOME. In "AM" its called "AppMan Mode" and its how "appman" works to made non privileged users installing apps in their $HOME

ivan-hc:
See https://github.com/ivan-hc/AM#how-to-use-am-in-non-privileged-mode-like-appman

valtoree:
I'm taking about @ivan-hc ur github am install app. Not touched appman. No idea what that is.

valtoree:
I've just installed ur app, as per github, "wget https://raw.githubusercontent.com/ivan-hc/AM/main/INSTALL && chmod a+x ./INSTALL && sudo ./INSTALL", then started installing apps like 'am -i appname'. Never had to write a path to custom app folders

valtoree:
Do I find where am has installed the apps and point the folder there?

valtoree:
Watched video it seems update script has enabled --user perhaps?

ivan-hc:
The apps are installed in /opt

See the spoiler at https://github.com/ivan-hc/AM#differences-between-am-and-appman

ivan-hc:
> Watched video it seems update script has enabled --user perhaps?

[...]

PR #613, opened 2024-06-01
Include modules update.am, sync.am and clean.am in APP-MANAGER / Merge 4 modules into one / Added a header to all modules, to briefly describe them

ivan-hc:
This update includes the two modules dedicated to CLI updates, apps and modules themselves within APP-MANAGER and the consequent reduction of modules from 16 to 14.

These are essential commands that should be executed at the end of each update process, not directly from the external module. It is therefore right that they return to where they were born, i.e. in the main CLI, as before the transition of "AM" from monolithic to modular.


ivan-hc:
Updating from the main CLI should fix furter issues like this https://github.com/ivan-hc/AM/issues/612 and this https://github.com/ivan-hc/AM/issues/600

ivan-hc:
The last commit allows the main CLI to automatically remove the obsolete modules at first update/sync

ivan-hc:
The sync.am and update.am modules are obsolete, they will still be present in the repository as long as old versions of "AM"/AppMan prior to this 6.8.5 are still in circulation

ivan-hc:
New module "management.am" that includes the following options:
- 	'-b' or 'backup'
- 	'-o' or 'overwrite'
- 	'-R'
- 	'-r' or 'remove'
- 	'--rollback'
- 	'lock'
- 	'unlock'
- 	'nolibfuse'

now we have 4 modules less to download. All the abow options are the ones to manage the installed apps.

PR #631, opened 2024-06-06
Monitor changes in the size of your installed apps!

ivan-hc:
Now information about the size, type, and version of each app will be listed in three different files before being merged under a single file that will show information about individual installed apps.

This is to constantly update the file that lists the sizes of the applications. They will thus be monitored constantly.

The file relating to the type of app will remain unchanged, so as to avoid slowdowns, the same goes for the file that lists the versions (with the exception of self-updating applications, for example Firefox and Thunderbird).

In the example, here we have "chromium" installed with "AppMan":

- the old dimensions are indicated in the blue rectangles;
- the large green rectangle in the center indicates the AppImage launched from the terminal and which uses a .home directory previously created with the option `-H`, so as not to disperse the dotfiles in the user's $HOME;
- in the red rectangles there is a change in size after using the app.
 
 
Istantanea_2024-06-06_21-53-09

May users always keep an eye on their installed apps and have absolute control over their "AM"/"AppMan" experience!

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="6.10.1"
+AMVERSION="6.10.2"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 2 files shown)


Samueru-sama:
Ivan you fucking read my mind, I was thinking about this today lol

ivan-hc:
This happens when you become passionate about an "improvised" project, by following it, you end up identifying with its creator.

PR #633, opened 2024-06-08
Optimizations to the files module (and display total size as well).

Samueru-sama:
I know the title sends chills down your spine @ivan-hc 😆 but I did test this it with **appman** and it works: 

image

-------------------------------

**However** I'm not sure if the spaces that I added for `Total size` is the right way of doing this, as I don't know if when using a different font this messes the order.

### DIFF
--- a/modules/files.am
+++ b/modules/files.am
@@ -1,14 +1,13 @@
-#!/usr/bin/env bash
+#!/bin/sh
 
 ##########################################################################
 # THIS MODULE ALLOWS YOU TO SEE THE MAIN INFORMATION ON THE INSTALLED APPS
 ##########################################################################
 
-function _files_header() {
+_files_header() {
 	echo ""
-	echo $(echo "YOU HAVE INSTALLED "
-	cd "$APPSPATH" &&
-	find . -type f -name 'remove' 2>/dev/null | sed -r 's|/[^/]+$||' | sort | uniq | wc -l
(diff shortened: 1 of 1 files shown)


ivan-hc:
> You can also make it display the total size in the first paragraph instead if you think that way is better.

Nope, its OK at the footer. Thanks

ivan-hc:
Why have you converted it in SHELL?

Samueru-sama:
> Why have you converted it in SHELL?

What do you mean? That I changed `bash` for `sh`? That's something that I hope all of am could eventually be converted to.

I normally only use bash when there is a feature that can only be used in bash, like array filtering, since that wasn't being used here I changed it to sh. 

When I tested the script I also tested that change and made sure that dash is indeed my sh.

ivan-hc:
This is amazing, thank you.

What if I change
```
echo "--------------------------------------------------------------------------------"
echo "- Total Size:							      $(du -sh "$APPSPATH" | awk '{print $1}' | sed 's/.$/   &/; s/$/iB/')"
echo ""
```
with
```
echo -e "\n TOTAL SIZE: $(du -sh "$APPSPATH" | awk '{print $1}' | sed 's/.$/ &/; s/$/iB/') of disk space in use\n"
```
to have

Istantanea_2024-06-08_23-03-21


?

ivan-hc:
Don't worry, I'll do it. Since it's your idea, I just wanted to propose this change of style.

PR #637, opened 2024-06-09
Optimize database and improve "query"

ivan-hc:


Samueru-sama:
Looks good.

ivan-hc:
@Samueru-sama about colors, you forgot to add the flag
```
			grep -i -E "$(echo "$@" | cut -f3- -d ' ' | tr -s ' ' '|')" "$AMPATH/$arch-apps" --color=always | _clean_lists_and_queries
			grep -i -E "$(echo "$@" | cut -f3- -d ' ' | tr -s ' ' '|')" "$AMPATH/libs-list" --color=always | _clean_lists_and_queries
```
to have this

Istantanea_2024-06-09_17-44-29 png

however, its strange that we have not a ◆ , and also Brave does not appear in chromium

ivan-hc:
EDIT: by removing " | _clean_lists_and_queries" it appears ok

Istantanea_2024-06-09_17-47-08 png

however, the flaf seems not to work for generic keywords

Samueru-sama:
Oh the --color flag needs to go after the path. I tried putting it with the other flags and it didn't work. 

Brave does appear in chromium, the full paragraph is this: 

`◆ chromium : Web Browser. Open source base of Google Chrome, Edge and Brave.`

The thing is that you have a function that truncates the text later on.

ivan-hc:
I cut the lines so as not to exceed 80 characters in length, I don't like seeing words truncated. If there was the possibility to move forward it would be better.

Samueru-sama:
Also I just noticed that you removed the entire ` _clean_lists_and_queries` function. Is that not needed anymore? I wasn't sure why sed would try to remove URLs from the package list.

ivan-hc:
> I wasn't sure why sed would try to remove URLs from the package list.

Oh! I forgot! We have third party repositories that work like this! They build the file to read with "-a" directly from the list!

See https://github.com/ivan-hc/neodb

Samueru-sama:
Something that can be done is to first run the function and then grep the result, however that essentially means that grep will never be able to match whatever text was removed after the 80 character limit. 

The other solution is to rewrite the apps that exceed that 80 character limit, shouldn't be that many lol. 

Or maybe raising the character limit to 100?

ivan-hc:
Or to remove that limit

ivan-hc:
Just tested, and now works (I had to be offline for this):

Istantanea_2024-06-09_18-11-38 png

ivan-hc:
@Samueru-sama as regards the colors for the keywords, it is not important, as the search terms must be "those", of course!

The fact of using them at least in listing the packages is already a great improvement.

ivan-hc:
> Btw hold making a whole new release announcement as I will try to convert all the other modules to use `/bin/sh`

@Samueru-sama  I'm not comfortable with SHELL, it doesn't have all the features that BASH has. That's why all my projects went to BASH. Let's say I almost hated SHELL. I switched AM/AppMan to BASH for a reason... I think from v3... I don't remember when.

ivan-hc:
I think this release is OK, can I publish it? Or is there something you want to add?

Samueru-sama:
> > Btw hold making a whole new release announcement as I will try to convert all the other modules to use `/bin/sh`
> 
> @Samueru-sama I'm not comfortable with SHELL, it doesn't have all the features that BASH has. That's why all my projects went to BASH. Let's say I almost hated SHELL. I switched AM/AppMan to BASH for a reason... I think from v3... I don't remember when.

WIll see, so far it has been easy to convert back to shell. 

Something that cannot be done easily with shell is array filtering, like I do in the `AppRun` of my `android-tools-appimage`. 

>@Samueru-sama I think this release is OK, can I publish it? Or is there something you want to add?

Yes you can, but what I meant by holding is not writing an entire lost post detailing the changes, as there is more to come soon lol (but if you like to do it go ahead).

ivan-hc:
> Yes you can, but what I meant by holding is not writing an entire lost post detailing the changes, as there is more to come soon lol (but if you like to do it go ahead).

Oh well, these are gradual changes. The more we do, the fewer changes to the code there will be and the easier I will be able to intervene in case of problems, like tonight, when I had to solve this https://github.com/ivan-hc/AM/pull/636

PS: I'm happy to have a collaborator who wants to contribute with innovation and not distort/confuse things, unlike someone else in the past. Thank you for these improvements you made. I really needed this kind of support. It is good for both my work with "AM" and my psychological well-being.

Samueru-sama:
@ivan-hc Looks like something broke with appman. 😭

I noticed that when I updated, for some reason hyperfine kept updating every time I ran `appman -u`

Then I ran the AM-updater script manually, and it worked and said that no update was needed like it should 🤔

So I removed hyperfine and now I can't install it again. Tested removing htop and I can't installed it either. Looks like it fails to download the file now.

However suyu works. It seems to only affect github for some reason.

ivan-hc:
@Samueru-sama Maybe you have reached the githug api call limit. Here I've no issues

Istantanea_2024-06-09_20-16-20 png

ivan-hc:
Wait about 1 hour.

I know this is a pain in the ass, I've opened a discussion for this https://github.com/ivan-hc/AM/discussions/578

Samueru-sama:
> Wait about 1 hour.
> 
> I know this is a pain in the ass, I've opened a discussion for this #578

Yeah looks like that was the issue as it fixed itself randomly. 🙃

What is weird is that I didn't download anything since I woke up from github. Only updated right now.

Samueru-sama:
@ivan-hc Btw I just updated the android-tools appimage to use sh: 

This is what I needed to change: 

All instances of `"${@:2}"` were replaced by a `$ARGS` variable, which is this: `ARGS="$(echo "$@" | cut -f2- -d ' ')"`

`if [[ "$yn" == [yY]* ]]; then` was replaced by `if echo "$yn" | grep -i '^y' >/dev/null 2>&1; then` Another way would be a case statement but I think what I did is simpler lol.

PR #639, opened 2024-06-09
Update AM-SAMPLE-AppImage

Samueru-sama:
Hey @ivan-hc I noticed you were updating the WINE appimages, I assume it was because of what I told you about hyperfine having a check before downloading.

However we are still using the older template in the repo, the template that I was talking about is something that I was experimenting with personally, I haven't done the PR because I don't know if it works ok with appimages that have a `zsync` file. So I'm making this PR with those changes included (+ some more).

**I HAVE NOT TESTED THIS YET!** (I mean this exact template, but most of the changes here have been tested in several apps before).

-------------------------------------------------------------------------

The main questions are: 

* Does the way I changed the zsync work? The only appimage that I use that has a zsync file is libreoffice, and that one uses a whole different script. **Also should the AM-updater script exit 1 if the zsync download fails or should it then try to manually download the file? (that wasn't being done before btw).**

**Also I think there was a mistake before**, there were was a `rm -R -f "/opt/$APP/*zs-old" "/opt/$APP/*.part"` but I that will not work, because the wildcard `*` inside quotes `""` doesn't expand as far as I know lol. Sorry if it was me that made this change before.

* The other questions is, are there appimages that don't have a top level .desktop file or symlink?

I made this change: 

```
./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
if [ -L ./"$APP".desktop ]; then
	LINKPATH=$(readlink ./"$APP".desktop)
	./"$APP" --appimage-extract "$LINKPATH" && mv ./squashfs-root/"$LINKPATH" ./"$APP".desktop
fi
```

So that we don't have to try several different .desktop locations like we were doing before with the Icon, however this will not work if it is possible for an appimage to not have a top level `.desktop` file, I've seen some that have a symlink and that works here, but if there is no symlink or .desktop it will not work. I don't know if it is possible for one to exist. 

* Finally does `$version` still need to be on line 19?

### DIFF
--- a/modules/template.am
+++ b/modules/template.am
@@ -314,7 +314,7 @@ while [ -n "$1" ]; do
 								mv ./am-scripts/$arch/$arg ./am-scripts/$arch/$arg.old
 								sed -n '1,14p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
-								sed -n '15,34p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
+								sed -n '16,34p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
 								sed -n '36,71p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								rm -f ./am-scripts/$arch/$arg.old;;
--- a/modules/template.am
+++ b/modules/template.am
@@ -312,9 +312,9 @@ while [ -n "$1" ]; do
 							case "$DOWNLOADURL" in
 							*)
 								mv ./am-scripts/$arch/$arg ./am-scripts/$arch/$arg.old
-								sed -n '1,18p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
+								sed -n '1,14p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
-								sed -n '20,34p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
+								sed -n '15,34p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
 								sed -n '36,71p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
 								rm -f ./am-scripts/$arch/$arg.old;;
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -39,7 +39,7 @@ SITE="REPLACETHIS"
 version0=$(cat "/opt/$APP/version")
 version=$(FUNCTION)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
-if [ "$version" != "$version0" ] || test -f "/opt/$APP/*.zsync"; then
+if [ "$version" != "$version0" ] || test -f /opt/"$APP"/*.zsync; then
 	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
 	wget "$version.zsync" 2>/dev/null || { wget "$version" || exit 1; }
 	# Use tar fx ./*tar* here for example in this line in case a compressed file is downloaded.
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -39,7 +39,7 @@ SITE="REPLACETHIS"
 version0=$(cat "/opt/$APP/version")
 version=$(FUNCTION)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
-if test -f "/opt/$APP/*.zsync"; then
+if test -f /opt/"$APP"/*.zsync; then
 	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
 	wget "$version.zsync" || exit 1
 	cd ..
@@ -57,7 +57,6 @@ elif [ "$version" != "$version0" ]; then
 
 	cd ..
 	mv --backup=t ./tmp/*mage ./"$APP" || exit 1
-	mv --backup=t ./tmp/* ./"$APP"
 	chmod a+x "/opt/$APP/$APP"
 	echo "$version" > ./version
 	notify-send "$APP is updated!"
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -54,6 +54,7 @@ elif [ "$version" != "$version0" ]; then
 	notify-send "A new version of $APP is available, please wait"
 	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
 	wget "$version" || exit 1
+
 	cd ..
 	mv --backup=t ./tmp/*mage ./"$APP" || exit 1
 	mv --backup=t ./tmp/* ./"$APP"
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -39,26 +39,15 @@ SITE="REPLACETHIS"
 version0=$(cat "/opt/$APP/version")
 version=$(FUNCTION)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
-if test -f /opt/"$APP"/*.zsync; then
+if [ "$version" != "$version0" ] || test -f "/opt/$APP/*.zsync"; then
 	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
-	wget "$version.zsync" || exit 1
+	wget "$version.zsync" 2>/dev/null || { wget "$version" || exit 1; }
 	cd ..
-	mv ./tmp/*.zsync ./"$APP".zsync
-	zsync "/opt/$APP/$APP.zsync"
-	rm -R -f ./*zs-old ./*.part ./tmp
-	chmod a+x "/opt/$APP/$APP"
-	echo "$version" > /opt/$APP/version
-	notify-send "$APP is updated!"
-	exit 0
-elif [ "$version" != "$version0" ]; then
-	notify-send "A new version of $APP is available, please wait"
-	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
-	wget "$version" || exit 1
-
-	cd ..
-	mv --backup=t ./tmp/*mage ./"$APP" || exit 1
-	chmod a+x "/opt/$APP/$APP"
+	mv ./tmp/*.zsync ./"$APP".zsync 2>/dev/null || mv --backup=t ./tmp/*mage ./"$APP"
+	zsync "/opt/$APP/$APP.zsync" 2>/dev/null
+	chmod a+x "/opt/$APP/$APP" || exit 1
 	echo "$version" > ./version
+	rm -R -f ./*zs-old ./*.part ./tmp ./*~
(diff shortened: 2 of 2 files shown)


ivan-hc:
I've updated the WINEs because the appimageupdatetool, when you check the version, it uses appimageupdatetool to check the version from the internet: 
- if you're offline, the version is "unknown"
- also, this kind of check is slower if compared with the check on the .zsync file

however, I had issues when trying to update the AppImage using zsync, so I had to change the AM-updater by including only the zsync part, and without `if/else/fi`.

Now everything works as expected.

ivan-hc:
You forgot spaces before and after line 57

Also, is `1>/dev/null` enough to silence errors in `mv`?

Also, why you added messages like "update not needed"? Its a responsability of the CLI.

ivan-hc:
A bit of patience, your questions are a lot... let me organize the answer :smile:

Samueru-sama:
> You forgot spaces before and after line 57

Adding them right now

> Also, is `1>/dev/null` enough to silence errors in `mv`?

I don't think that's the kind of error that should be silenced, if `DirIcon` fails to be moved, you will have a broken .desktop without icon in your applications menu. 

iirc mv doesn't produce an output when it is susceful, in that case the `1>/dev/null` better be removed all together I think

> Also, why you added messages like "update not needed"? Its a responsability of the CLI.

That was there before. What I changed is the conditions that make it appear. That is before `update not needed` was given if `version` and `0version` were the same and the script exited afterwards. 

Now instead, if `version` and `0version` are NOT the same, then it updates and `exits 0` that way the update message isn't given. 

That update message was always there anyway. And I really don't want to get rid of it as part of my testing involves running the `AM-updater` manually making sure that all parts of it work and that message helps me a lot lol.

Samueru-sama:
Btw I only added one space before the `cd ..` as I understand that space is for a possible `tar fx`. 

However what is the space after needed for? It isn't in the other cd .. where there is a comment for tar.

ivan-hc:
> Hey @ivan-hc I noticed you were updating the WINE appimages, I assume it was because of what I told you about hyperfine having a check before downloading.

Nope, my answer is at first message https://github.com/ivan-hc/AM/pull/639#issuecomment-2156824771

> However we are still using the older template in the repo, the template that I was talking about is something that I was experimenting with personally, I haven't done the PR because I don't know if it works ok with appimages that have a `zsync` file. So I'm making this PR with those changes included (+ some more).
> 
> **I HAVE NOT TESTED THIS YET!** (I mean this exact template, but most of the changes here have been tested in several apps before).

OK. Take your time. Head. Don't make rash choices.

> The main questions are:
> 
>     * Does the way I changed the zsync work?

I don't know

> The only appimage that I use that has a zsync file is libreoffice, and that one uses a whole different script.

I use it too and updated sometime, no issues here.

> Also should the AM-updater script exit 1 if the zsync download fails or should it then try to manually download the file? (that wasn't being done before btw).

I always try, regardless. Maybe the developer will decide to add a .zsync file if he hadn't done so before. If he does this, when you upgrade, while running "-s" you will get a message saying that the script has changed. Once you have read this message, if you want, you can reinstall it and try the new features.
 
> Also I think there was a mistake before, there were was a `rm -R -f "/opt/$APP/*zs-old" "/opt/$APP/*.part"` but I that will not work, because the wildcard `*` inside quotes `""` doesn't expand as far as I know lol. Sorry if it was me that made this change before.

At this point I'm not sure if we have done some wrong changes. I'm sure that all brave* and firafox* AppImages are been updated with your template. You should check all scripts containing your syntax at this point.

> The other questions is, are there appimages that don't have a top level .desktop file or symlink?

you say inside the Appimage? All appImages have at least this structure:
```
app.dexktop
app.png/svg
AppRun
```

this is enough, for example, if the app is the AppRun script itself. But the .desktop file must be in place.

You say if all Appimages have a .destop file and icon to be integrated in the menu? All they have this, but sometime its useless.

For example, all "wine" ive updated before, they are command line utilities. La launcher is needed there.

> I made this change:
> 
> ```
> ./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
> if [ -L ./"$APP".desktop ]; then
> 	LINKPATH=$(readlink ./"$APP".desktop)
> 	./"$APP" --appimage-extract "$LINKPATH" && mv ./squashfs-root/"$LINKPATH" ./"$APP".desktop
> fi
> ```
> 
> So that we don't have to try several different .desktop locations like we were doing before with the Icon, however this will not work if it is possible for an appimage to not have a top level `.desktop` file, I've seen some that have a symlink and that works here, but if there is no symlink or .desktop it will not work. I don't know if it is possible for one to exist.

sometime the Appimages have a stange .desktop file without entries... just a text file. I don't remember where I've seen this. I think on the `vlc-cmatomic` or `vlc-icflorescu`, or maybe in `avidemux` (but I use `debian-testing-avidemux`).
 
>  Finally does `$version` still need to be on line 19?

19 and 40.

All depends on the changes we do to "[template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am)".

In the end, we are the only two that are costantly uploading apps.

ivan-hc:
> Btw I only added one space before the `cd ..` as I understand that space is for a possible `tar fx`.
> 
> However what is the space after needed for? It isn't in the other cd .. where there is a comment for tar.

I don't remember where, but I had to use it before.

Well, if the situation comes up, nothing prevents us of changing th script manually, for one app

Samueru-sama:
> I don't know

kekw, I guess time to hunt for a simple appimage that is compatible with the template that has a zsync. 

> > Also should the AM-updater script exit 1 if the zsync download fails or should it then try to manually download the file? (that wasn't being done before btw).
> 
> I always try, regardless. Maybe the developer will decide to add a .zsync file if he hadn't done so before. If he does this, when you upgrade, while running "-s" you will get a message saying that the script has changed. Once you have read this message, if you want, you can reinstall it and try the new features.

What I mean is that right now (and iirc always) if there is a `zsync` file the `AM-updater` script will _always_ use it for updating, even if the `zsync` download fails. 

This can be changed so that if the `zsync` download fails, it tries to download the whole appimage instead. What do you think?

> > Also I think there was a mistake before, there were was a `rm -R -f "/opt/$APP/*zs-old" "/opt/$APP/*.part"` but I that will not work, because the wildcard `*` inside quotes `""` doesn't expand as far as I know lol. Sorry if it was me that made this change before.
> 
> At this point I'm not sure if we have done some wrong changes. I'm sure that all brave* and firafox* AppImages are been updated with your template. You should check all scripts containing your syntax at this point.

All my appimages have been updating fine, however only `libreoffice` has a zsync file.

> you say inside the Appimage? All appImages have at least this structure:

Yes

> this is enough, for example, if the app is the AppRun script itself. But the .desktop file must be in place.
> You say if all Appimages have a .destop file and icon to be integrated in the menu? All they have this, but sometime its useless.
> For example, all "wine" ive updated before, they are command line utilities. La launcher is needed there. 
> sometime the Appimages have a stange .desktop file without entries... just a text file. I don't remember where I've seen this. I think on the `vlc-cmatomic` or `vlc-icflorescu`, or maybe in `avidemux` (but I use `debian-testing-avidemux`).

Oh in that case I no problem, something similar can happen with `.DirIcon` being a dummy file that is empty. For those rare cases the script can be changed manually to fix that.

There is also a posibility for `.DirIcon` to be a symlink to another symlink, so far only two appimages have shown this, one was `lime3ds` and the other was `scrcpy` or something like it, for those I simply repeated the test for a symlink.

> 19 and 40.
> 
> All depends on the changes we do to "[template.am](https://github.com/ivan-hc/AM/blob/main/modules/template.am)".
> 
> In the end, we are the only two that are costantly uploading apps.

Can that range be lowered? Because I don't know if you have noticed, the template right now has this to be able to meet that: 

image

And I don't know, it doesn't look very good 😆

Samueru-sama:
@ivan-hc Btw updating libreoffice to a type3 appimages breaks the zsync update: 

image

ivan-hc:
> @ivan-hc Btw updating libreoffice to a type3 appimages breaks the zsync update:

now I understand why I updated a few times and there were slightly different versions of the version declared in "-f" and the one in the internal menu of the application :eyes:

ivan-hc:
So I should change "nolibfuse.am"

Samueru-sama:
> So I should change "nolibfuse.am"

I think in this case the whole updater script has be reworked. 

And `nolibfuse` removes the `.zsync` file, that way it defaults to updating by downloading the whole appimage and then converting it back to type3

ivan-hc:
> What I mean is that right now (and iirc always) if there is a `zsync` file the `AM-updater` script will _always_ use it for updating, even if the `zsync` download fails.
> 
> This can be changed so that if the `zsync` download fails, it tries to download the whole appimage instead. What do you think?

:+1: 

> Can that range be lowered? Because I don't know if you have noticed, the template right now has this to be able to meet that:
> 
> And I don't know, it doesn't look very good 😆

the script only serves two things: installing the app (via "-i") and notifying you that it has changed (via "-s").

If you think it needs to be changed, in the next PR you also need to change the relevant line in "template.am"
```
							case "$DOWNLOADURL" in
							*)
								mv ./am-scripts/$arch/$arg ./am-scripts/$arch/$arg.old
								sed -n '1,18p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
								sed -n '20,39p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
								echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/$arch/$arg
								sed -n '41,132p' ./am-scripts/$arch/$arg.old >> ./am-scripts/$arch/$arg
								rm -f ./am-scripts/$arch/$arg.old;;
							esac
```
for example, to move it to line 18 you need to change "18p" with "17p" and "20,39p" with "19,39p".

In the function abovethe lines are between 18 and 20 and 39 and 41... while 132 is approximate. It serves to avoid cutting what is in the original file.

To insert the command into the "version" variable, I have to copy only the parts that interest me from the original script and leave empty those lines that I need to fill (19 and 40).

ivan-hc:
> > So I should change "nolibfuse.am"
> 
> I think in this case the whole updater script has be reworked to be able to update without zsync.
> 
> And `nolibfuse` removes the `.zsync` file, that way the updater defaults to updating by downloading the whole appimage and then converting it back to type3

this is a good idea... but we need to put a message for this. Libreoffice is big, its different to update it using zsync or to download it again. With zsync its faster.

Samueru-sama:
> 👍

Alright that was done already, check the changes see how the new `AM-updater` is made in the template. 

>To insert the command into the "version" variable, I have to copy only the parts that interest me from the original script and leave empty those lines that I need to fill (19 and 40).

Ok, I will try this tomorrow because it is 10 pm already lol 

>this is a good idea... but we need to put a message for this. Libreoffice is big, its different to update it using zsync or to download it again. With zsync its faster.

I would just change the description of `nolibfuse` to indicate that zsync updates get disabled. 

And maybe also add a `echo "This will disable zsync"` when `nolibfuse` is used.

ivan-hc:
@Samueru-sama consider your PR a supplement to the one I just closed. We should also update scripts that use this new syntax.

Possibly, in "nolibfuse" we should add an additional check, to check whether AM-updaters support version comparison as an alternative, be it mine, in old versions... be it yours.

For example, "libreoffice" does not have this type of support, also because its URL is uniform, has no version numbers and never changes... and all "wines" are already Type3. They therefore do not need this change.

As for me, I'll have to settle for keeping Libreoffice Type2, with all its additional 30MB :disappointed:

ivan-hc:
@Samueru-sama you're wrong here https://github.com/ivan-hc/AM/pull/639/commits/b84df044b58cb5e901a14ca16b36730ef7ae929f

you've changed the option 2, while Appimages are managed by option 0

Samueru-sama:
> @Samueru-sama you're wrong here [b84df04](https://github.com/ivan-hc/AM/commit/b84df044b58cb5e901a14ca16b36730ef7ae929f)
> 
> you've changed the option 2, while Appimages are managed by option 0
> 
> option 2 is the one for the archives

Fuck me, I did a Control+F on the site and I saw that there were only 3 `sed -n` in the entire module wtf. 

I will close this PR and make a new one, because there was a lot of errors in the commit history already before this lol.

PR #646, opened 2024-06-10
Dev

Samueru-sama:
This change makes is obvious when the zsync update fails. 

Lets wait to see what happens at cpu-x before deciding on removing the file all together lol. 

Btw this new template is pretty much ready to be merged into main, the only thing is that the line that downloads the `.zsync` file can be disabled in the meantime.

### DIFF
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -35,7 +35,7 @@ SITE="REPLACETHIS"
 version0=$(cat "/opt/$APP/version")
 version=$(FUNCTION)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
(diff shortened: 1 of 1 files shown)


ivan-hc:
> Btw this new template is pretty much ready to be merged into main, the only thing is that the line that downloads the `.zsync` file can be disabled in the meantime.

We can't know if the developer will plan to add a .zsync file in the future, the approach used so far is to "take a guess"

Samueru-sama:
> > Btw this new template is pretty much ready to be merged into main, the only thing is that the line that downloads the `.zsync` file can be disabled in the meantime.
> 
> We can't know if the developer will plan to add a .zsync file in the future, the approach used so far is to "take a guess"

WIth the way the template is written, if a zsync file is added in the future, it will actually download it as well. 

The problem is that since the AM-updater defaults to trying `zsync` first and if that `zsync` is broken it basically breaks the entire update for that application.

PR #647, opened 2024-06-10
Update template.am and AppImage's installation script template / fix some invasive workflows

ivan-hc:


ivan-hc:
@Samueru-sama However, I want to remind you that, in case an icon is not intercepted in the AppImage, the "install.am" module contains a function to download icon from https://portable-linux-apps.github.io , in case the "icons" directory is empty.

ivan-hc:
Just tested brave-appimage and then converted to type3

Also if the .zsync file was removed, it is downloaded again.

In my previous changes, the .zsync file had to be removed, so that, the update can be done with comparison of versions.

Your new template breaks this.

ivan-hc:
Also, you've removed notification when update has been detected. Command `notify-send "A new version of $APP is available, please wait"`

Samueru-sama:
> Also, you've removed notification when update has been detected. Command `notify-send "A new version of $APP is available, please wait"`

Oh, didn't notice that.

Btw is this notification really needed? because on my end when I run `am -u` because I have over 30 applications, the notifications just fill the screen all at once and are not really readable as result.

ivan-hc:
> > Also if the .zsync file was removed, it is downloaded again.
> 
> Yeah you have to disable the line that downloads the `.zsync` file like I mentioned it before.

Nope, if someone uses nolibfuse, the check should be automatic.

Also, have you tested this using the option `--rollback`? Its broken too using Anydesk, for example, that has not a .zsync file

ivan-hc:
> Btw is this notification really needed? because on my end when I run `am -u` because I have over 30 applications, the notifications just fill the screen all at once and are not really readable as result.

Do all 30 update at once? Impossible.

The notification, in my case, is convenient, since I use few AppImages.

Samueru-sama:
>Nope, if someone uses nolibfuse, the check should be automatic.

Oh in that case I think what needs to be done is adding the same `[ -e *zsync ]` before download the zsync file. If the file gets deleted it wont be downloaded automatically again.

> > Btw is this notification really needed? because on my end when I run `am -u` because I have over 30 applications, the notifications just fill the screen all at once and are not really readable as result.
> 
> Do all 30 update at once? Impossible.
> 
> The notification, in my case, is convenient, since I use few AppImages.

Since we have parallel updates this happens very often, granted I need to wait for several days to get the explosion of notifications. 

I will record it btw.

ivan-hc:
The --rollback option takes into account the existence of the "AM-updater" file. If a .zsync file does not exist, it will make a copy and patch the AM-updater file, to upgrade/downgrade the app.

ivan-hc:
> > Also, have you tested this using the option --rollback? Its broken too using Anydesk, for example, that has not a .zsync file
> 
> Wanna know something? I didn't know `--rollback` existed lol

My fault, I wrote too many interconnected options. “AM” is too cool :sunglasses:

ivan-hc:
You better fix this situation ASAP, we could have a ton of PR on new apps (I wish we did) and we would all have apps that can't downgrade.

Samueru-sama:
https://streamable.com/rtzljw

>You better fix this situation ASAP, we could have a ton of PR on new apps (I wish we did) and we would all have apps that can't downgrade.

Define ASAP, I can fix it today (hopefully). If anything only the updater part of the template needs to be reverted to the older method.

ivan-hc:
Don't worry, I'll try to adapt the function to the changes

Samueru-sama:
> Don't worry, I'll try to adapt the function to the changes

If the function disables the update, I think the easiest solution would be a sed that replaces the `set -u` in the `AM-updater` script for something like this: 

`echo "Updates disabled & exit 0`

That way the script doesn't continue. 

And if you want to disable the rollback have am replace that back for `set -u`

PR #650, opened 2024-06-11
Temporary disable zsync downloads

Samueru-sama:
We already ran into two appimages with broken zsync, those are `brave` and `cpu-x`

This disables it until we figure out a way to fix those issues. In the meantime we should only focus on getting zsync to work on big applications like libreoffice (which it does work already).

### DIFF
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -14,7 +14,7 @@ chmod a+x "/opt/$APP/remove"
 # DOWNLOAD AND PREPARE THE APP, $version is also used for updates
 version=$(FUNCTION)
 wget "$version" || exit 1
-#wget "$version.zsync" 2> /dev/null
+#wget "$version.zsync" 2> /dev/null # Comment out this line if you want to use zsync, warning that more often than not it is broken
 echo "$version" > /opt/$APP/version
 # Use tar fx ./*tar* here for example in this line in case a compressed file is downloaded.
 cd ..
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -14,7 +14,7 @@ chmod a+x "/opt/$APP/remove"
 # DOWNLOAD AND PREPARE THE APP, $version is also used for updates
 version=$(FUNCTION)
 wget "$version" || exit 1
-wget "$version.zsync" 2> /dev/null
+#wget "$version.zsync" 2> /dev/null
 echo "$version" > /opt/$APP/version
 # Use tar fx ./*tar* here for example in this line in case a compressed file is downloaded.
 cd ..


ivan-hc:
You have disabled it during the installation.

We should not care if the developer have a brocken .zsync file, him should fix it.

See https://github.com/ivan-hc/AM?tab=readme-ov-file#an-application-does-not-work-is-old-and-unsupported

The better thing we can do is to add a check to validate the .zsync file, and then allow the user to disable it.

I'll take care of it when I've finisced with --rollback.

Samueru-sama:
> We should not care if the developer have a brocken .zsync file, him should fix it.

I think it is a very bad idea to have this enabled by default. It was never working before (with the exception of libreoffice), and now we are only seeing problems with it.

Keep in mind that a broken zsync means that the app never updates.

Most appimages are very small (less than 100 MiB) I don't think it is worth this hassle. Only on big appimages one should try to get the zsync to work.

Also I think this is a bigger issue with `linuxdeploy` will see what happens with cpu-x

ivan-hc:
@Samueru-sama If they fix it in the future, we would have to come back here and change the behavior again.

I repeat, "AM" only has to distribute the apps, it doesn't have to also take care of the mistakes of those who created them. Let us not burden ourselves with responsibilities that do not concern us.

We are just here to create a centralized bridge for user and developer to communicate quickly.

Samueru-sama:
> @Samueru-sama If they fix it in the future, we would have to come back here and change the behavior again.
> 
> I repeat, "AM" only has to distribute the apps, it doesn't have to also take care of the mistakes of those who created them. Let us not burden ourselves with responsibilities that do not concern us.
> 
> We are just here to create a centralized bridge for user and developer to communicate quickly.

Ivan zsync wasn't working before because the check for the zsync file was broken. None of the appimages that were made in the last few months have a working zsync to begin with. 

Enabling this means that zsync will now be tried instead of updating the regular way it was done before, and we've already seen that it is broken. 

Also fixing back would be a single sed in the database that replaces all instances of `#wget "$version.zsync"` for `wget "$version.zsync"`

I really don't like this idea that one shouldn't bother to fix this, reminds me of archlinux when it removed the glibc patch **which made more people switch to flatpak to use steam** because flatpak takes care of that issue while arch doesn't think they should enable a one line of code in their repo.

ivan-hc:
> Ivan zsync wasn't working before because the check for the zsync file was broken. None of the appimages that were made in the last few months have a working zsync to begin with.
> 
> Enabling this means that zsync will now be tried instead of updating the regular way it was done before, and we've already seen that it is broken.
> 
> Also fixing back would be a single sed in the database that replaces all instances of `#wget "$version.zsync"` for `wget "$version.zsync"`
> 
> I really don't like this idea that one shouldn't bother to fix this, reminds me of archlinux when it removed the glibc patch **which made more people switch to flatpak to use steam** because flatpak takes care of that issue while arch doesn't think they should enable a one line of code in their repo.

In this case... I agree with you.

Samueru-sama:
> > Ivan zsync wasn't working before because the check for the zsync file was broken. None of the appimages that were made in the last few months have a working zsync to begin with.
> > Enabling this means that zsync will now be tried instead of updating the regular way it was done before, and we've already seen that it is broken.
> > Also fixing back would be a single sed in the database that replaces all instances of `#wget "$version.zsync"` for `wget "$version.zsync"`
> > I really don't like this idea that one shouldn't bother to fix this, reminds me of archlinux when it removed the glibc patch **which made more people switch to flatpak to use steam** because flatpak takes care of that issue while arch doesn't think they should enable a one line of code in their repo.
> 
> In this case... I agree with you.

Btw just in case this change makes so that the appimages update like they have been doing before. 

If we run into an appimage that has a fixed url and only uses a zsync to update, like `cpu-x` we don't have other choice other than enabling `zsync` for that specific app and hope that it works.

Samueru-sama:
https://github.com/TheTumultuousUnicornOfDarkness/CPU-X/issues/331

The issue is that we need to use appimageupdatetool instead. 

However using appimageupdatetool opens up a GUI and it seems there is no TUI only option.

ivan-hc:
> However using appimageupdatetool opens up a GUI and it seems there is no TUI only option.

this is wrong. I've recently replaced the installation script for wine* for this reason. See the last changes in that commit https://github.com/ivan-hc/AM/commit/1cb233b48fc9a7872ee89ae25c47776053d6051e the CLI version was https://github.com/ivan-hc/AppImageUpdate/releases/download/continuous/appimageupdatetool-x86_64.AppImage

ivan-hc:
are you using the mine or the one from https://github.com/AppImageCommunity/AppImageUpdate/releases ?

also there are two, one of 33mb and another one of 45mb, the big one is the gui, mine is based on the CLI, the small one

Samueru-sama:
> are you using the mine or the one from https://github.com/AppImageCommunity/AppImageUpdate/releases ?
> 
> also there are two, one of 33mb and another one of 45mb, the big one is the gui, mine is based on the CLI, the small one

Oh! no I did you use yours, didn't know it existed. 

Why did you forked it btw?

ivan-hc:
I've forked it to made it a Type3 one

ivan-hc:
Also, appimageupdatetool is slow when you try to check the version of a program.

Samueru-sama:
Well, in that case looks like temporary solution will be more than temporary 😆

cpu-x has versioned releases as well, I will fix the script later in the day.

ivan-hc:
NOTE, in the new --rollback option I'm working on, to disable the .zsync behaviour, I temporarilly rename it as "$APP".zsync.old before I rename it normally.

This is why I think that to solve the .zsync issue is something the developer should care on.

Samueru-sama:
> NOTE, in the new --rollback option I'm working on, to disable the .zsync behaviour, I temporarilly rename it as "$APP".zsync.old before I rename it normally.
> 
> This is why I think that to solve the .zsync issue is something the developer should care on.

There is no issue, the "issue" is that some appimages need appimageupdatool to update.

ivan-hc:
Also, since we can use both behaviours for the update (zsync and comparison), an integrity check on that file is what we need.

In case its brokem, then we need to remove it, and maybe report this to the user.

ivan-hc:
Its better I start closing this PR https://github.com/ivan-hc/AM/pull/651 for the new release, before we start studying how to solve the .zsync thing.

PR #652, opened 2024-06-11
Merge pull request #649 from ivan-hc/dev

ivan-hc:
Fix new AppImage template

### DIFF
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -37,11 +37,12 @@ version=$(FUNCTION)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
 if [ "$version" != "$version0" ] || [ -e /opt/"$APP"/*.zsync ]; then
 	mkdir "/opt/$APP/tmp" && cd "/opt/$APP/tmp" || exit 1
-	wget "$version.zsync" 2>/dev/null || { wget "$version" || exit 1; }
+	[ -e /opt/"$APP"/*.zsync ] || notify-send "A new version of $APP is available, please wait"
+	[ -e /opt/"$APP"/*.zsync ] && wget "$version.zsync" 2>/dev/null || { wget "$version" || exit 1; }
 	# Use tar fx ./*tar* here for example in this line in case a compressed file is downloaded.
 	cd ..
 	mv ./tmp/*.zsync ./"$APP".zsync 2>/dev/null || mv --backup=t ./tmp/*mage ./"$APP"
-	[ -e /opt/"$APP"/*.zsync] && { zsync "/opt/$APP/$APP.zsync" || notify-send -u critical "zsync failed to update $APP"; }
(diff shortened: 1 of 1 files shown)


Samueru-sama:
Btw what do you mean by integrity checks? (I just read the release notes). 

The new template already sends a notify-send when zsync fails to update the appimage. And that gets tested when a new template is made, I think that's enough of a integrity check, if zsync fails to update the appimage we disable the zsync download.

ivan-hc:
This only happens with newer scripts... but what about the older ones?

This is why we need an integrity chek. It must be included in -u or -s.... or at least we should convert all the installation scripts to this new template.

ivan-hc:
or maybe included in `-i`

ivan-hc:
we need a plan for this change

Samueru-sama:
> or at least we should convert all the installation scripts to this new template.

They have to be converted because a lot of them have broken checks to use zsync already.

Samueru-sama:
This whole new template ended up being something that sidetracked me, because I was actually working on optimizing and converting the modules to use `sh`, and I noticed you were updating apps to the existing template and I thought you were doing it to get the check for empty url that wasnt in the template back then 😅

Samueru-sama:
The only thing is that I think all the recent appimages that I've added don't have the zsync check, because I have no plans on adding zsync to my appimages. Those don't need to be updated.

Samueru-sama:
FInally @ivan-hc if you want to start fixing the scripts, I think you have to start with the biggest apps first, as those are the ones where it makes sense to use zsync.

ivan-hc:
> and I noticed you were updating apps to the previous template and I thought you were doing it to get the check for empty url that wasnt in the template back then 😅

???

Nope, new --rolback also supports new templates.

All it does is to clone AM-updater as AM-rollback, and then:
- made $version0 unusable;
- temporarily rename the .zsync files as .old not to made them detected
- replace all lines containing the "wget" command with the "wget" command created with the function

so the comparison action can act to replace the installed app with what you've choose

ivan-hc:
> FInally @ivan-hc if you want to start fixing the scripts, I think you have to start with the biggest apps first, as those are the ones where it makes sense to use zsync.

A big work has been already done in the years, and all bug fixes have been made after user's feedback.

I'd like to keep have these scripts for now, and to change them just in case. I can't focus my whole life on "AM".

Samueru-sama:
> ???

I updated the template because remember the issue that I had that I hit the API limit? 

The issue cause the `AM-updater` to always update, because turns out that if `version` is empty it also updates the app lol. 

I already had a pseudo new template that I was using in the apps that I was adding that had the check for empty `version` but it wasn't in the template. 

After I mentioned that I saw that you updated the wine appimages, and I thought you did it to get that check for empty `version` which that wasn't the reason you did it 😅

ivan-hc:
> After I mentioned that I saw that you updated the wine appimages, and I thought you did it to get that check for empty `version` which that wasn't the reason you did it 😅

I'm sorry I broke your heart :rofl:  . It wasn't for you that I made that change.

PR #656, opened 2024-06-12
Update aisap

Samueru-sama:
Oh boy this thing looks amazing. 

The sandbox can be as simple as replacing the symlink of the application for a wrapper script that runs `aisap [OPTIONS] [APP]`.

### DIFF
--- a/programs/x86_64/aisap
+++ b/programs/x86_64/aisap
@@ -1,114 +1,75 @@
 #!/bin/sh
 
+# AM INSTALL SCRIPT VERSION 3. 
+set -u
 APP=aisap
-REPO="mgord9518/aisap"
+SITE="mgord9518/aisap"
 
-# CREATE THE FOLDER
-mkdir /opt/$APP
-cd /opt/$APP
+# CREATE DIRECTORIES AND ADD REMOVER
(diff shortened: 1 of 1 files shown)


ivan-hc:
Let try aisap ASAP

Samueru-sama:
> Let try aisap ASAP

I already opened an issue at their repo asking some questions. 

It is far easier to use than firejail. But there are some stuff I still don't understand. 

The app already has a set of default permissions for the most common applications, however I'm still not sure on how to overwrite this. 

I think having a flat permission that allows all apps to dead `$XDG_CONFIG_HOME` and `~/Downloads` is a good default.

Samueru-sama:
@ivan-hc Now I'm using brave with sandbox. It only has access to my `~/Downloads` dir. 

When you run the app with aisap it makes a fake home next to the app appimage. Inside that fakehome I copied the following dirs from `XDG_CONFIG_HOME` and `XDG_DATA_HOME`: 

```
gtk3-3.0
fontconfig
kdeglobals
icons
themes
```

And now brave has full integration with my desktop. It uses my theme and cursor, etc. 

Hopefully this can be made more flexible and I don't need to have a bunch of different fakehomes with repeated files all over, maybe one fakehome for sandboxed apps would be better.

PR #667, opened 2024-06-15
Update README.md

Samueru-sama:
Since this is tool is used by the `nolibfuse` option.


ivan-hc:
So appimagetool must be removed as reference

PR #674, opened 2024-06-18
Lots of improvements

Samueru-sama:
mgord9518 fixed the issue with some apps not getting their fake home 🎉 I haven't thanked then yet, I will right after I write this. 

New improvements to sandboxes. 

* Now it better tries to give access to the sandboxed app original config directory.
For example gimp names its `$XDG_CONFIG_HOME` directory `GIMP` in all caps, to fix that issue I added this check in the sandbox script: 

`APPCONF=$( ls "$CONFIGDIR" | grep -i "$APPNAME" | head -1 )`

Which will correctly find the name of the directory we want. 

If that directory doesn't exist, it defaults to using the name of the binary. `--add-file "$CONFIGDIR/${APPCONF:-$APPNAME}":rw \`

* **Now `sandboxes.am` asks the user which user directories it should give access to:** 

image

By default access to `~/Downloads ~/Documents ~/Pictures ~/Videos` will be denied, unless the user says yes to each respective question. If nothing is entered then it assumes that you said no in each question (You can also type N, n, No, etc).

**I will likely expand this to also allow access to a specific directory that the user writes its path to.**

* Added more access to `dconf` and `dbus` by default, now when you sandbox `cpu-x` you don't have a broken theme (Well not in the case of my PC, but this is because I don't have a `~/.config` directory and aisap assumes that it is located there 😅, **but for most users it should not break the theme**).

image

### DIFF
--- a/modules/sandboxes.am
+++ b/modules/sandboxes.am
@@ -100,6 +100,11 @@ case "$1" in
 		DATADIR="${XDG_DATA_HOME:-$HOME/.local/share}"
 		CONFIGDIR="${XDG_CONFIG_HOME:-$HOME/.config}"
 		CACHEDIR="${XDG_CACHE_HOME:-$HOME/.cache}"
+
+		# Try to find the right name of the app xdg directories, as sometimes it is not the same as $APPNAME
+		APPDATA=$( ls "$DATADIR" | grep -i "$APPNAME" | head -1 )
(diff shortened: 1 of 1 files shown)


Samueru-sama:
Showing the changes with gimp, I only gave it access to `~/Pictures` and `~/Downloads`.

image

PR #681, opened 2024-06-19
Changed a lot of things

Samueru-sama:
I put a check inside the sandbox script and module that if for some reason one of the xdg user dir variables equals `$HOME` it will empty  the variable:

```
if [ "$XDG_DOWNLOAD_DIR" = "$HOME" ]; then XDG_DOWNLOAD_DIR=""; fi
if [ "$XDG_MUSIC_DIR" = "$HOME" ]; then XDG_MUSIC_DIR=""; fi
if [ "$XDG_PICTURES_DIR" = "$HOME" ]; then XDG_PICTURES_DIR=""; fi
if [ "$XDG_VIDEOS_DIR" = "$HOME" ]; then XDG_VIDEOS_DIR=""; fi
if [ "$XDG_DOCUMENTS_DIR" = "$HOME" ]; then XDG_DOCUMENTS_DIR=""; fi
```

This causes both, sandbox script and the sandbox module to use the fallback value of each variable.

Here for example I defined `XDG_DOWNLOAD_DIR=$HOME` in my `xdg-user-dirs.dirs` 

image

As result the script will instead use the default value of `~/Downloads`. 

----------------------------------

The good thing of this approach is that it prevents the issue that the user might allow access to `~/Downloads` but then later on their xdg-user-dirs for some reason starts defining `XDG_DOWNLOAD_DIR=$HOME` without them knowing. 

----------------------------------

I also changed the order of operations so that sudo is only needed once. This also means that the moving of the script and the removing of exec permissions for the appimage won't take place until the very end. That way if the user control+c half way it doesn't result in changes to the appimage. 

-----------------------------------

I removed the several `cat HEREDOC` you were using before and changed it back for the older method. 

The problem with the multiple  `cat HEREDOC` is that if the user says no to any of the questions, we have to have the `--rm-file` line in the script, because remember that `aisap` has an internal set of rules that have to be overwritten. So we cannot simply not add the line if they say no. 

**What I mean is that if the `-rm-file "${XDG_DOWNLOAD_DIR:-~/Downloads}"` is not added to the script, aisap might still allow access to it depending on the application.**

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -30,8 +30,7 @@ function _clean_amcachedir() {
 }
 
 APPMANCONFIG="$CONFIGDIR/appman"
-
-export SCRIPTDIR="$(if command -v xdg-user-dir &>/dev/null; then xdg-user-dir DESKTOP; else echo $HOME; fi)"
+export SCRIPTDIR="$(xdg-user-dir DESKTOP 2>/dev/null || echo "$HOME")"
 
 # Determine sytem package manager
 function _system_package_manager_check() {
--- a/modules/sandboxes.am
+++ b/modules/sandboxes.am
@@ -62,12 +62,6 @@ case "$1" in
 		APPIMAGEPATH="$APPSPATH/$2"
 		APPIMAGE="$APPIMAGEPATH/$2"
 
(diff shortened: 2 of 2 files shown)


Samueru-sama:
Btw we have to apply this trick of using as little sudo as possible for AM. As that will make it much easier to later on add doas support.

PR #684, opened 2024-06-20
Update AM-SAMPLE-Archive

Samueru-sama:
I've done 0 tests on this.

This template assumes that the `.desktop` and `icon` are missing from the downloaded `$version`,and have to be gotten thru another download, as that is very common with this kind of applications.

### DIFF
--- a/templates/AM-SAMPLE-Archive
+++ b/templates/AM-SAMPLE-Archive
@@ -1,66 +1,61 @@
 #!/bin/sh
 
+# AM INSTALL SCRIPT VERSION 3. 
+set -u
 APP=SAMPLE
 SITE="REPLACETHIS"
 
-# CREATE THE FOLDER
-mkdir /opt/$APP
-cd /opt/$APP
+# CREATE DIRECTORIES AND ADD REMOVER
+[ -n "$APP" ] && mkdir -p "/opt/$APP/tmp" "/opt/$APP/icons" && cd "/opt/$APP/tmp" || exit 1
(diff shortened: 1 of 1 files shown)


ivan-hc:
```
DESKTOP=$(FUNCTION)
ICON=$(FUNCTION)
```
???

ivan-hc:
@Samueru-sama before thinking about the .desktop file and the icon, I was thinking about how to solve this, let's say that an archive contains another subfolder:
```
mv --backup=t ./tmp/* ./
```

ivan-hc:
I think I'll solve the .desktop issue using an HEREDOC in template.am

ivan-hc:
In the meantime I merge, so I can adapt the option 2 to your template

Samueru-sama:
> This is great, I like the way you manage tar/zip archives
> 
> but...
> 
> ```
> DESKTOP=$(FUNCTION)
> ICON=$(FUNCTION)
> ```
> 
> ???

https://github.com/ivan-hc/AM/blob/main/programs/x86_64/speedcrunch

A lot of portable and cli applications often don't shit the .desktop or icon inside.

However they usually have those in their repo, so I instad get them with wget instead of making the .desktop.

>@Samueru-sama before thinking about the .desktop file and the icon, I was thinking about how to solve this, let's say that an archive contains only another subfolder containing all the files:

You could put a check for a directory in the top level of `tmp` and thta way the mv has an extra step.

Issue #683, opened 2024-06-20
Apps 


GeoBoi254:
TLock
https://github.com/eklairs/tlock/releases

Clipse
https://github.com/savedra1/clipse/releases

Onefetch
https://github.com/o2sh/onefetch/releases

Cpufetch
https://github.com/Dr-Noob/cpufetch/releases

Pfetch-rs
https://github.com/Gobidev/pfetch-rs/releases

Pop
https://github.com/charmbracelet/pop/releases

Smassh
https://github.com/kraanzu/smassh/releases

Dooit
https://github.com/kraanzu/dooit/releases

Samueru-sama:
@ivan-hc Do you want me to send the new template for CLI applications before starting this?

Because last time I didn't have the md files.

ivan-hc:
Don't worry, if you have not time, I can compile .md files by myself, these are github apps, and I can generate a dumb option 0 for AppImages for them, to have the .md file.

I'm curious about your new template.

Samueru-sama:
>I'm curious about your new template.

It's what I've been using for the recent non appimage applications, it is a stripped version of the appimage template.

ivan-hc:
ok

ivan-hc:
I think I should start improving this https://github.com/ivan-hc/AM/blob/main/templates/AM-SAMPLE-Archive

Option 2

ivan-hc:
maybe by adding inbuilt commands `if`/`fi` to use tar/unzip... and maybe the option itself should have a .desktop file Y, N

ivan-hc:
all installation scripts are on my PC right now, using the new upcoming template

ivan-hc:
I was looking for the number of apps available on Flathub, and noticed that until 3 months ago they were 2400, while the verified ones were 1000.

We have some metascripts in the database, but on 2059 installation scripts available, we have about 2050 apps (including the 40 of kdenlive and the 4-5 kdeutils). If we merge metascripts, we have 2000+ apps. Unofficial or official... the source code is always availableby following the URLs in `-a` or by downloading the script with `-d`. I don't know how to read ostree.

But the idea that we have 2000+ apps we have uploaded for fun, as enthusiasts... and the 2400+ uploaded by a great community... with only 1000 apps verified... made me smile :muscle:

Samueru-sama:
> I was looking for the number of apps available on Flathub, and noticed that until 3 months ago they were 2400, while the verified ones were 1000.
> 
[...]

PR #687, opened 2024-06-20
Improve "template.am" to speed up the creation of installation scripts

ivan-hc:


ivan-hc:
@Samueru-sama in the meantime I've done a code refactoring for this module andnow both scripts for AppImages and Archives (the latter is the one you've worked on in the PR) are at same level, so functions to add download URLs from non-github sources are the same.

I've not worked on .desktop files yet, but I hope to release this for the night, since we an issue opened for this https://github.com/ivan-hc/AM/issues/683

It would be great if our template for Archives and generic programs is able to do things as authomatic as the AppImage's one already do.

ivan-hc:
@Samueru-sama thanks, I've used your tip to add also 7z and deb among the supported packages.

ivan-hc:
Maybe I can put a question to optionally create a .desktop file, and that is. Maybe I could set "N"  as default.

ivan-hc:
Since you said that you will improve the command to move stuff on top directory, I've removed it among the options.

All is left is to made the .desktop file optional and custom.

Samueru-sama:
What could be done is add a function in templates that generates a very simple generic .desktop for the application and then it appends it to the generated template with the `>>` operator. 

Btw is the there reason you added this with a `;` separator? 

`[ -e ./*tar* ] && tar fx ./*tar* && rm -f ./*tar*; [ -e ./*zip ] && unzip -qq ./*zip 1>/dev/null && rm -f ./*zip`

Instead of just: 
```
[ -e ./*tar* ] && tar fx ./*tar* && rm -f ./*tar*
[ -e ./*zip ] && unzip -qq ./*zip 1>/dev/null && rm -f ./*zip
```
?

And finally I have one big request that please I beg you. 

Remove this: 

`[ -e ./*deb ] && ar x ./*deb && tar fx ./data.tar.xz && rm -f ./*tar* ./*deb`

It really doesn't look good 😅 We should only try to use deb packages as last resort and not add it to the template.

Please!

---------------------------------------

Later in the night I will try to make something that detects that there is a lone directory after the extract and that way the mv moves what's inside the directory.

ivan-hc:
About .deb packages, you're right, I should get rid of them... or maybe let people use them in option 1, so they can create AppImages from them.

I'll remove it and then I'll scale the line.

The reason because I did it is this
```
function _template_if_hosted_elsewhere() {
	sed -i "s#REPLACETHIS#$RESPONSE#g" ./am-scripts/"$arch"/"$arg"
	# ADD A DESCRIPTION FOR THIS APPLICATION
	read -r -ep " ◆ ADD A BRIEF DESCRIPTION FOR THIS APPLICATION $(echo -e '\n : ')" COMMENT
	case "$COMMENT" in
	*)
		# CREATE A WEBPAGE FOR https://portable-linux-apps.github.io CONTAINING ALL THE INFO ABOUT THIS APP
		echo "# $(echo "$arg" | tr '[:lower:]' '[:upper:]')" >> ./am-scripts/portable-linux-apps.github.io/apps/"$arg".md
		echo -e "\n $COMMENT\n\n SITE: $RESPONSE\n\n | [Applications](https://portable-linux-apps.github.io/apps.html) | [Home](https://portable-linux-apps.github.io)\n | --- | --- |" >> ./am-scripts/portable-linux-apps.github.io/apps/"$arg".md &&
		# CREATE A NEW LINE FOR THE APPLICATION'S LIST
		echo "◆ $arg : $COMMENT" >> ./am-scripts/list
		;;
	esac
	echo "-----------------------------------------------------------------------"
	# IF YOU CAN, USE A ONE-LINE COMMAND TO DOWNLOAD THE PROGRAM
	read -r -ep " USE A ONE-LINE SHELL COMMAND TO CHECK ONLY THE URL TO THE PROGRAM $(echo -e '\n\n if the URL is fixed, simply add the "echo" command at the beginning\n\n :') " DOWNLOADURL
	case "$DOWNLOADURL" in
	*)
		mv ./am-scripts/"$arch"/"$arg" ./am-scripts/"$arch"/"$arg".old
		sed -n '1,14p' ./am-scripts/"$arch"/"$arg".old >> ./am-scripts/"$arch"/"$arg"
		echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/"$arch"/"$arg"
		sed -n '16,35p' ./am-scripts/"$arch"/"$arg".old >> ./am-scripts/"$arch"/"$arg"
		echo 'version=$('$DOWNLOADURL')' >> ./am-scripts/"$arch"/"$arg"
		sed -n '37,132p' ./am-scripts/"$arch"/"$arg".old >> ./am-scripts/"$arch"/"$arg"
		rm -f ./am-scripts/"$arch"/"$arg".old
		;;
	esac
}
```
this function is compatible with both your templates in option 0 and 2

ivan-hc:
@Samueru-sama I only have one issue when I need to move the files from ./tmp to ./, if the file is only one, it should be renamed as "$APP", so we need to change `mv ./tmp/* ./` and `mv --backup=t ./tmp/* ./` into another one line command. Any idea?

ivan-hc:
> @Samueru-sama I only have one issue when I need to move the files from ./tmp to ./, if the file is only one, it should be renamed as "$APP", so we need to change `mv ./tmp/* ./` and `mv --backup=t ./tmp/* ./` into another one line command. Any idea?

@Samueru-sama help me with this "mv" thing too

Samueru-sama:
> > @Samueru-sama I only have one issue when I need to move the files from ./tmp to ./, if the file is only one, it should be renamed as "$APP", so we need to change `mv ./tmp/* ./` and `mv --backup=t ./tmp/* ./` into another one line command. Any idea?
> 
> @Samueru-sama help me with this "mv" thing too

I will, but do I have to do it like right now? Because it will take a bit longer for me to come up with a solution for the mv. 

My idea was to do a `ls | wc --l` in tmp. 

Then if that number is `1` (remember that now we remove the tar, zip, etc file after decompressing) and that file is a direectory, then the mv will do this: `mv ./tmp/*/* ./`

If the file is not a directory, then the mv does this `mv ./tmp/* ./$APP`

ivan-hc:
@Samueru-sama just because I'm a total neubie, I don't know much about `printf` so I always use `echo`.

I was asking you if the syntax of my commit was correct https://github.com/ivan-hc/AM/pull/687/commits/d7ebf16fafe249c1192a4ad2be3a161b44685dd1 but I've seen that you did the same, so now I know hot to use it. Thanks.

In brief, strange commands with multiple escapes need `printf`, on the contrary, normal lines need `printf '\n%s'`, isn't it?

Samueru-sama:
> In brief, strange commands with multiple escapes need `printf`, on the contrary, normal lines need `printf '\n%s'`, isn't it?

printf is a lot like awk, we first define what we want to do, and then we give the string (the text). 

`\n` means that it will be printed on a new line. 

`%s` is the string (the text). 

Sometimes you can also put those inside the text like it is done with echo, sometimes it doesn't work and I have to separate them, I did this a lot in the sandboxes module recently.

ivan-hc:
> I will, but do I have to do it like right now?

its not urgent, today we were able to create our installation scripts quickly... and you too have noticed that some commands had to be `./tmp/* ./"$APP"` instead of `./tmp/* ./`.

This release is more important for us that maintain this repository and are the ones that upload the apps.

So to add this check or not is not important, all we need to do is to upload the installation scripts quickly and when they are archives or scripts, we need to test them still quickier and without extracting them one by one each time.

If there is something I hate the most is to upload archives -_-

Today we have a new quick way to do so, thanks to.

This is enough for today... tomorrow I will find other ways to exploit you :rofl:

Samueru-sama:
I replied on reddit. 

I can help you with this today, like in a few hours, I need to eat xd.

Samueru-sama:
By the way Ivan can you change the way the URL gets generated for the new appimage templates so that it no longer needs the `jq`? 

Something like this: 

`sed 's/[()",{}]/ /g; s/ /\n/g' | grep -oi 'https.*$APP.*appimage$' | head -1`

Should automatically match most appimages. And if we don't want the continuos built, it gets changed to this: 

`sed 's/[()",{}]/ /g; s/ /\n/g' | grep -oi 'https.*$APP.*appimage$' | grep -v continuos | head -1`

ivan-hc:
Do a PR and let we see the changes together.

Samueru-sama:
> Do a PR and let we see the changes together.

I tried to change that but the resulting template still had the old url filtering 🥲

I'm the in the middle of adding some small changes to sandboxes, nothing big btw.

ivan-hc:
> By the way Ivan can you change the way the URL gets generated for the new appimage templates so that it no longer needs the `jq`?

what part of the code hurts you? I've not understand

PR #691, opened 2024-06-21
Update AM-SAMPLE-Archive

Samueru-sama:
Alright I think this is what you wanted. 

I noticed you didn't remove the check for a deb file in the updater. 

You are very lucky that the `if [ -d ./tmp/* ]` fails when there is a directory **and** multiple files, it only goes  thru when there is a lone directory. 

Edit: Also I changed the check fora tar file to this `[ -e ./*tar.* ]` with a dot that way we avoid a possible false positive.

### DIFF
--- a/templates/AM-SAMPLE-Archive
+++ b/templates/AM-SAMPLE-Archive
@@ -15,10 +15,10 @@ chmod a+x "/opt/$APP/remove"
 version=$(FUNCTION)
 wget "$version" || exit 1
 [ -e ./*7z ] && 7z x ./*7z && rm -f ./*7z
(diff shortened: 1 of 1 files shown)


ivan-hc:
> I noticed you didn't remove the check for a deb file in the updater.

:eyes: I hadn't noticed. Thanks!
 
> You are very lucky that the `if [ -d ./tmp/* ]` fails when there is a directory **and** multiple files, it only goes thru when there is a lone directory.

we're lucky I haven't fallen asleep yet :smile: 

> Edit: Also I changed the check fora tar file to this `[ -e ./*tar.* ]` with a dot that way we avoid a possible false positive.

Thanks! Thanks! Thanks!

Issue #688, opened 2024-06-21
Cannot remove apps


GeoBoi254:
I went to uninstall some programs and i gave me an error. It seems the remove script is not formatted properly.

Example:
nvtop
```
#!/bin/sh
rm -R -f /home/geo/.local/share/applications/AM-nvtop.desktop /home/geo/Applications/nvtop /home/geo/.local/bin/nvtop
```

cpufetch
`#!/bin/sh\nset -e\nrm -f /home/geo/.local/bin/cpufetch\nrm -R -f /home/geo/Applications/cpufetch`

Samueru-sama:
What error did you get? 

Also is nvtop or cpufetch that has the problem?

GeoBoi254:
> What error did you get?
> 
> Also is nvtop or cpufetch that has the problem?

nvtop uninstalls properly but cpufetch doesnt
Heres the error
`/home/geo/Applications/appman/modules/management.am: line 36: /home/geo/Applications/cpufetch/remove: cannot execute: required file not found`

ivan-hc:
> cpufetch `#!/bin/sh\nset -e\nrm -f /home/geo/.local/bin/cpufetch\nrm -R -f /home/geo/Applications/cpufetch`

I've installed it before I've uploaded it, with AppMan... strange

Samueru-sama:
Oh I see the problem now. 

`#!/bin/sh\nset -e\nrm -f /home/geo/.local/bin/cpufetch\nrm -R -f /home/geo/Applications/cpufetch`

This will create a remove file that should look like this: 

```
#!/bin/sh
set -e
/home/geo/.local/bin/cpufetch
rm -R -f /home/geo/Applications/cpufetch

```

Very likely there is something wrong in the install script that caused `/home/geo/.local/bin/cpufetch` to not be created. As result the remover will not continue, that's intended.

ivan-hc:
I cannot reproduce the issue

![Istantanea_2024-06-21_03-42-17 png](https://github.com/ivan-hc/AM/assets/88724353/36a3aa82-dbd2-4999-a990-87c6827c21f4)

GeoBoi254:
@ivan-hc try removing it

Samueru-sama:
Just tested the script as well: 

![image](https://github.com/ivan-hc/AM/assets/36420837/5acc19d3-3506-47aa-9a6c-bcf5ca4d65ea)

[...]

PR #695, opened 2024-06-21
👀

Samueru-sama:
* Now sandboxes asks if you want to configure access to directories, By default hitting enter without anything will say Yes, but if you say no it will skip the several questions for each directory, essentially denying the access to them.

Very useful for when you want to sandbox a CLI appliation that often doesn't need access to any of these directories. 

image

* Added `~/Desktop` and `~/Games` to the list of directories. 

`XDG_GAMES_DIR` is technically not part of the spec but it is commonly used and [requested](https://xdg.freedesktop.narkive.com/5opetQ4G/recognizing-games-dir). And `aisap` has ~/Games as a common rule as well, so with this method we can't deny it if the user wants to. 

I actually had a `~/Games` directory before I discovered aisap lol.

* Added to option to allow a specified directory

image

**I also tested this with both `bash` and `dash` as sh so I don't run into another surprise.** 

~~I may later add a check that prints a big warning if the user gives access to a spooky location~~ 😆

EDIT: Done!

### DIFF
--- a/modules/sandboxes.am
+++ b/modules/sandboxes.am
@@ -72,22 +72,25 @@ case "$1" in
 		# Check if we are using AM or AppMan
 		printf '\n%s\n' " Making aisap script for \"$(echo "$AMCLI" | tr a-z A-Z)\"..."
 
+		[ -z "$AMCACHEDIR" ] && exit 1
 		rm -Rf "$AMCACHEDIR/sandbox-scripts"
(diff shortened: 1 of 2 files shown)


Samueru-sama:
Also I forgot, I updated the nvtop script thinking that was the issue at #688

Threw it here anyway lol

PR #697, opened 2024-06-21
Sneaky

Samueru-sama:


Samueru-sama:
Btw @ivan-hc I made this script that builds and installs `fuse 2.9.9` in `/usr/local`: 

```
#!/bin/sh

# AM INSTALL SCRIPT VERSION 3. 
set -u
APP=libfuse2
SITE="https://github.com/libfuse/libfuse/releases/download/fuse-2.9.9/fuse-2.9.9.tar.gz"

echo "This script is meant for distros that don't allow fuse2 and fuse3 to be installed together, like debian and forks"

sleep 1

if ! command -v autoconf 1>/dev/null; then
	echo "You need autoconf to be installed for this script to work"
	echo "We will try to install it"
	sudo apt install autoconf
fi

# CREATE DIRECTORIES AND ADD REMOVER
[ -n "$APP" ] && mkdir -p "/opt/$APP" && cd "/opt/$APP" && wget $SITE || exit 1
tar fx ./*tar.* && rm -f ./*tar.* && cd fuse-2* && ./configure --disable-example --enable-lib && sudo make install && cd .. || exit 1
echo "2.9.9" > ./version

cat >> ./remove << 'EOF'
#!/bin/sh
APP=libfuse2
cd /opt/libfuse2/fuse-2* && sudo make uninstall && rm -rf /opt/libfuse2
EOF
chmod a+x ./remove
```

What do think about this replacing the debian package?

ivan-hc:
How many dependences other than "make" are needed for this?

Samueru-sama:
> How many dependences other than "make" are needed for this?

I think that is the only one, iirc autoconf should pull make when you install it. 

make, autoconf, etc are very basic dependencies that often come installed by default.

ivan-hc:
> autoconf

I see 
Istantanea_2024-06-22_01-13-24 png

Samueru-sama:
> autoconf

now install autoconf and try again.

ivan-hc:
Exists a way to debloat Fedora? To update using DNF is a pain, expecially in virtual machine (I've 550 RPM packages to be updated).

Samueru-sama:
> Exists a way to debloat Fedora? To update using DNF is a pain, expecially in virtual machine (I've 550 RPM packages to be updated).

Does fedora not allow libfuse2 and libfuse3 to be installed together?

I really don't know the answer to your question, never had good experiences with fedora, last time I tried it was in 2018 or so.

ivan-hc:
I need a compiler too

Istantanea_2024-06-22_01-17-29 png

ivan-hc:
however, its enough to have only the library in /usr/local/lib, no other files, to made AppImages type2 work on any distro, regardless if you have fuse3 installed

ivan-hc:
this is why my script only extracts the .deb pachage from Debian

Samueru-sama:
> I need a compiler too
> 
> Istantanea_2024-06-22_01-17-29 png

```
if ! command -v autoconf 1>/dev/null || ! command -v gcc 1>/dev/null; then
	echo "You need autoconf and gcc installed to be installed for this script to work"
	exit 1
fi
```

ivan-hc:
@Samueru-sama at this point its better to download a precompiled ibrary, as we do with Appimages, don't you?

Samueru-sama:
> @Samueru-sama at this point its better to download a precompiled ibrary, as we do with Appimages, don't you?

Installing libraries on the system that may not have been compiled for is a bad idea.

I had an issue with cmake before that turned out to be caused by a library that I had in my home when I built polybar, this is why now polybar is an appimage instead. 

Also extracting them for a deb package doesn't look good.

ivan-hc:
but works

ivan-hc:
also, what GNU/Linux disto is as secure as Debian?

Samueru-sama:
> but works

Ivan you were recently telling me about having AM reviewed, don't you think it looks really bad if the reviewer were to noticed  that you are extracting libraries from a deb package? And more importantly, this can also be installed on a non debian distro, who knows what kind of broken behavior this can cause. 

fuse is also used by `ntfs-3g` and `mtpfs`, breaking those can cause a lot of trouble for the user.

> also, what GNU/Linux disto is as secure as Debian?

What?

ivan-hc:
consider that the two distributions that are politically enemies of AppImage adoption are Ubuntu and Fedora:
- Ubutnu is based on Debian, so a library from Debian is 100% compatible if installed so;
- Fedora, I've always used my script to install it, never faced issues and all AppImage are working well.

This is the exact thing we would do if we want to manually install a package from  a distro to another, with the only difference that all we need is one file (and a symlink to it), libfuse2

ivan-hc:
And this is the exact thing that many AUR packages will do on Arch Linux: to extract files form .deb packages. I've seen them.

ivan-hc:
With the differenca that the AUR packages will extract files from third-party .deb packages, while we use an official one, from Debian, the universal operating system.

Why an unofficial AUR package would be more legitimate in doing this and we not?

Samueru-sama:
Okay, hopefully I'm just too paranoid and this never ends up biting us. 

>Why an unofficial AUR package would be more legitimate in doing this and we not?

The AUR is know for being broken and arch and arch-based distros have big disclaimers that they don't officially support it.

I no longer use the aur.

ivan-hc:
> The AUR is know for being broken and arch and arch-based distros have big disclaimers that they don't officially support it.
> 
> I no longer use the aur.

and this is why I don't use Arch as my main distro but only in VM: Debian has much more programs in its official repositories than the ones listed in core, extra, community and multilib.

However, if the AUR exists is because there are many people that will use it.

That said, I doubt that an Ubuntu/Fedora user will not try to install `libfuse2`/`fuse-libs` from APT/DNF first and will try out our installation script... however this is just an help. An additional help. A workaround. Not the defaoult choice.

Samueru-sama:
What debian package do you need that is not available as an arch package? 

Arch bundles their packages, for example libreoffice is one big package instead of being several ones. Usually what is available in debian you also have it available in arch.

ivan-hc:
The only package that Arch Linux supports officially and that Debian hasn't that I need is Avidemux.

However, I'm glad I can build it from both Debian and Arch https://github.com/ivan-hc/Avidemux-unofficial-appimage

Samueru-sama:
`rofi-emoji` is also available in arch as an official package but not in debian.

ivan-hc:
I don't know how much is mainstream "rofi-emoji", I still use
```
:) xD \m/ METAL \m/ <3 U_U u.u O_o
```
and so on for my emojy


PS: lol (cit.)

ivan-hc:
emojis will not made my videos shorter -_-

Samueru-sama:
Jokes aside `rofi-emoji` is an emoji selector that I call by typing `super+o` that's why you see me using emojis here: 

image

ivan-hc:
there are those who still open the browser looking for emojis to copy/paste (me), but the problem is that there are still too many people who are convinced that you have to look for AppImages on random sites to get them... and that's it a real security problem. Would you trust Appimagehub.com? I do not.

ivan-hc:
in the meantime I've just found an issue
```
# Check if you are installing an app or a library
function _if_library() {
	echo -e "◆ $(echo '"'$arg'"' | tr a-z A-Z): starting installation script" &&
	if grep -q "/usr/local/lib" ./$arg; then
		echo -e '\n ⚠️ This script installs a system library in /usr/local/lib, cancel it if:\n'
		echo ' - it is already installed in the system;'
		echo -e ' - its already present in your repositories (so install it from there).\n'
		read -p " Do you wish to continue (N,y)?" yn
		case "$yn" in
		'y'|'Y')	;;
		'n'|'N'|*)	echo -e "\n INSTALLATION ABORTED!"; return 0;;
		esac	
	fi
}
```
how should I interrupt this process and skip to the next argument?

Module install.am (I've done the change to made N the default choice.

Samueru-sama:
Also is there a reason you made the library thing a separate function in AM instead of just being a regular package script? 

You can put the questions inside the script like I did the check for autoconf.

ivan-hc:
You know? I was thinking to implement your function in the install module as a separated funftion, so that it will be built into the module.

I'm going doing the refactoring in the meantime

Samueru-sama:
> You know? I was thinking to implement your function in the install module as a separated funftion, so that it will be built into the module.
> 
> I'm going doing the refactoring in the meantime

I think it is better for it to be a installation script and that is it. It doesn't need to be its own function in am.

I also don't see a point in having those libs functions in am, what you are doing to install the debian package can also be just a install script. 

Just my opinion, since installing libraries is not something that am was made to do (since we target portable applications and appimages).

ivan-hc:
Because libraries, to work globally, need to be in /usr/lib or /usr/local/lib, not in ~/.local/lib

This is why they are different.

Samueru-sama:
> ~/.local/lib

You can make it work in `~/.local/lib` by adding that location to `$LD_LIBRARY_PATH` similar to how `~/.local/bin` gets added to `$PATH`.

ivan-hc:
PS: just noticed that also this function in install.am keeps installing packages not needed

```
function _if_metapackage() {
	for metapackage in $METAPACKAGES; do
		if grep -q "$metapackage" ./"$arg" 2> /dev/null; then
			if test -d "$APPSPATH"/"$metapackage"/tmp; then
				$SUDOCOMMAND "$APPSPATH"/"$metapackage"/remove 2> /dev/null
			elif test -d "$APPSPATH"/"$metapackage"; then
				echo ' ◆ '"$(echo "\"$arg\"" | tr '[:lower:]' '[:upper:]')"' IS PART OF "'"$(echo "$metapackage" | tr '[:lower:]' '[:upper:]')"'", ALREADY INSTALLED'; break
			fi
		fi
	done
}
```

I've understand that all this is because I've converted everything in functions, so `return 0` does not work anymore

ivan-hc:
> > ~/.local/lib
> 
> You can make it work in `~/.local/lib` by adding that location to `$LD_LIBRARY_PATH` similar to how `~/.local/bin` gets added to `$PATH`.

the libfuse2 installation script ends with `sudo ldconfig` to made applications recognize the library once installed

ivan-hc:
to understand what I'm sayng, try to use this exact command
```
appman -i platform-tools adb
```
this should install only the first, while thye second installs the same first reference, so it should exit the function and go to the next argument.

Its not so.

ivan-hc:
platform-tools result installed twice

Samueru-sama:
Talking about bugs, guess what: 

image

IT CAME BACK 😭

ivan-hc:
it seems that "return 0" only works for its own function... so I've tested the module by implementing the four functions into the main one and all works now

ivan-hc:
try now

Samueru-sama:
> try now

Something weird happened that when I updated I had an explosion of notifications of `A new version of $APP`...  even though only one app actually updated.

ivan-hc:
Let continue talking at https://github.com/ivan-hc/AM/pull/700

PR #698, opened 2024-06-22
Small changes to sandboxes

Samueru-sama:
Now you can also unsandbox the application with am by running `am --disable-sandbox $APP` instead of `$APP --disable-sandbox`

image

Changed the way the directories get listed when they are asked. 

Now when `XDG_DOCUMENTS_DIR` is defined, instead of am printing the entire path to it, it just says `~/Documenti` for example. 

I changed this because in your video you had `~/Games` printed while all other directories had their full path. So this way we have consistency. 

image

Also I added a `sleep 0.5` before the question to allow access to specific directory is given, because it has happened to me several times that I accidentally skip that question when I just want that one and not the others.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -540,6 +540,7 @@ function _generate_options_list() {
 		--less
 		--pkg
 		--rollback
+		--disable-sandbox
 		--sandbox
 		--silent
 		--system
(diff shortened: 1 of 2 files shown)


Samueru-sama:
Please Ivan don't make whole video about this 😅, this is just a small change.

PR #700, opened 2024-06-22
Fix "install.am" + more "sandboxing" improvements

ivan-hc:


Samueru-sama:
I'm getting this every time I updated: 

https://streamable.com/oxmnrk

Also now I have a bunch of empty `tmp` directories in each application.

Samueru-sama:
Wait nvm, I think I hit the github rate limit again.

ivan-hc:
Are you sure you've not reached the github API call limit?

Samueru-sama:
> Are you sure you've not reached the github API call limit?

Yeah it was that, what happens is that I still have a bunch of old scripts that don't have the check againts empty `$version`

ivan-hc:
how many apps have you installed?

Samueru-sama:
> how many apps have you installed?

38

ivan-hc:
you need to have a github key then or you will always have the issue

Samueru-sama:
> you need to have a github key then or you will always have the issue

`{"message":"API rate limit exceeded for ip-address. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://docs.github.com/rest/overview/resources-in-the-rest-api#rate-limiting"}`

I think you can put a check when `am -u` is ran that it checks if github responds with `API rate limit exceeded` and that way it doesn't continue with trying to update.

Samueru-sama:
I will be updating some of my installed apps to the new template, I think I will also merge all the deabeef into a script like we did we mpv, that way I don't have apps with long names like `deadbeef-devel-appimage`

ivan-hc:
> I will be updating some of my installed apps to the new template, I think I will also merge all the deabeef into a script like we did with mpv, that way I don't have apps with long names like `deadbeef-devel-appimage`

When you do this, remove the reference from the list, so I can quickly update the website

Issue #701, opened 2024-06-22
Apps


GeoBoi254:
sptlrx
https://github.com/raitonoberu/sptlrx/releases

lexido
https://github.com/micr0-dev/lexido/releases

vhs
https://github.com/charmbracelet/vhs/releases

glow
https://github.com/charmbracelet/glow/releases

freeze
https://github.com/charmbracelet/freeze/releases

soft-serve
https://github.com/charmbracelet/soft-serve/releases

typioca
https://github.com/bloznelis/typioca/releases

Samueru-sama:
@ivan-hc I think the jq thing I was talking about in the template has to be fixed before we add more apps.

Samueru-sama:
Ugh that lexido is not xdg base dir compliant

ivan-hc:
Can someone of you take track of the programs marked as "other" in `am -f`? I'd like to fix this.

Samueru-sama:
> Can someone of you take track of the programs marked as "other" in `am -f`? I'd like to fix this.

here are mine: 

```
◆  dolphin-emu           |  unknown       |  appimage-type3     |  64    MiB
◆  ppsspp                |  unknown       |  appimage-type3     |  26    MiB
◆  fastfetch             |  unknown       |  binary/executable  |  23    MiB
◆  handlr                |  unknown       |  binary/executable  |  3.9   MiB
◆  lf                    |  unknown       |  other              |  3.5   MiB
◆  gron.awk              |  unknown       |  script             |  984   KiB
◆  i3lock-color          |  unknown       |  launcher           |  180   KiB
```

@GeoBoi254 can you show your output of `appman -f`?

GeoBoi254:
```
◆  thorium-avx2    |  124.0.6367.218  |  appimage-type2     |  258   MiB
◆  ventoy          |  1.0.99          |  binary/executable  |  23    MiB
◆  smassh          |  3.1.3           |  binary/executable  |  19    MiB
◆  platform-tools  |  35.0.1          |  set/tools          |  17    MiB
◆  gping           |  1.16.1          |  other              |  5.5   MiB
◆  pfetch-rs       |  2.9.2           |  other              |  3.9   MiB
◆  cpufetch        |  1.05            |  other              |  1.1   MiB
◆  nvtop           |  3.1.0           |  appimage-type2     |  452   KiB
```

Samueru-sama:
> ```
[...]

PR #702, opened 2024-06-22
Unified deadbeef, updates some apps and added more applications, changed the template

Samueru-sama:
That was tricky but it worked out well.


Samueru-sama:
I updated the archive because I discovered that there is a 3rd possibility for the mv. 

* The first is that if it is a lone directory.

* The second is if it is a lone file.

* The 3rd one (this one) is when there is multiple files without directory, that causes `if [ -d ./tmp/* ]` to fail with a too many arguments error!. 

I changed the else condition to this, which takes into account that there can be multiple files without directory in tmp: 

`./tmp/* ./"$APP" 2>/dev/null || mv ./tmp/* ./`

Samueru-sama:
Closes #701

Samueru-sama:
@ivan-hc I actually removed the old [deadbeef](https://github.com/ivan-hc/AM/pull/702/commits/4b540e23440a445d865e37872fe7f4d0ee3dfe33) but looks like I copied the wrong list when adding the other apps that caused them to be back.

ivan-hc:
@Samueru-sama don't worry, I've solved :+1:

Issue #706, opened 2024-06-22
Option to skip _patch_bashrc_and_profile #705


Samueru-sama:
I would love an option to have the `APP-MANAGER` script skip the function `_patch_bashrc_and_profile` where the script writes the following to `~/.bashrc` (and other dotfiles):

```shell
export PATH=$(echo $PATH | tr ":" "\n" | grep -v "/.local/bin" | tr "\n" ":" | sed s/.$//)
export PATH=$PATH:$HOME/.local/bin
```

Personally, I don't like anything touching my bash dotfiles and prefer to manage my `PATH` manually via my own dotfiles.

Any option to disable that would be great. I don't know if you'd prefer an environment variable to control it like `[[ -z "${APPMAN_SKIP_PATCH_BASHRC_AND_PROFILE:-} ]] && _patch_bashrc_and_profile` or a value in the `appman-config` config file.

---

Unrelated to this, I just want to say I love `appman`. I had ignored AppImages due to the lack of a package management command line tool. Thank you so much for this. This plus `AppImaGen` and `ArchImage` are great. I very much appreciate it.

_Originally posted by @rvenutolo in https://github.com/ivan-hc/AM/discussions/705_

Samueru-sama:
I think the default behavior should be like this: 

```
if ! echo "$PATH" | grep .local/bin; then
    function_to_patch
fi

```

Also do we have to patch the bashrc? I'm a zsh user, on zsh you only have to add `$HOME/.local/bin` to `zprofile`, adding it to zshrc would cause  `$HOME/.local/bin` to repeat in PATH the second time you open a terminal window.

Samueru-sama:
Also worth mentioning that adding `$HOME/.local/bin` to PATH is job of the distribution as detailed in the XDG Base Dir specification: 

[`User-specific executable files may be stored in $HOME/.local/bin. Distributions should ensure this directory shows up in the UNIX $PATH environment variable, at an appropriate place.` ](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html)

So yeah I think we should just give a warning and ask the user if they want to patch their `.profile`/`.bash_profile`.

ivan-hc:
@Samueru-sama I think you are absolutely right.

I would like to stop patching configuration files, and I would like to delegate everything to the README. It's not "AppMan" who should handle them.

I want to disable the function, or at least I would like an option that can solve the problems.

Samueru-sama:
[...]

PR #707, opened 2024-06-22
Now "AppMan" prompts you to enable ~/.local/bin in $PATH, if not in place

Samueru-sama:
I did some limited testing, but I'm not a bash user so let me know if I need to change anything. 

I removed the patching of the `.bashrc` because as far as I know it isn't needed and can actually cause problems like I mentioned in the issue.

closes #705 #706

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="6.14-1"
+AMVERSION="6.14-2"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -78,8 +78,8 @@ function _am() {
 #######################################
 
 function _patch_bashrc_and_profile() {
-	if ! echo "$PATH" | grep "$BINDIR"; then
-		read -p "Warming: \"${XDG_BIN_HOME:-$HOME/.local/bin}\" is not in PATH, would you like appman to patch your .profile or .bash_profile? Y/n): " yn
+	if ! echo "$PATH" | grep -q "$BINDIR"; then
+		read -rp "Warming: \"${XDG_BIN_HOME:-$HOME/.local/bin}\" is not in PATH, would you like appman to patch your .profile or .bash_profile? (Y/n): " yn
 		if ! echo "$yn" | grep -i '^n' >/dev/null 2>&1; then
 			if test -f "$HOME/.profile"; then
 				printf "export PATH=$PATH:$BINDIR" >> "$HOME/.profile"
@@ -92,6 +92,7 @@ function _patch_bashrc_and_profile() {
 		fi
 	fi
 }
+
 
 function _appman_initialize() {
 	APPSDIR=$1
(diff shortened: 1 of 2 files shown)


Samueru-sama:
wait my read isn't working lol

ivan-hc:
I'd like to remove completelly this patch, $PATH order is something that the user should choose.

Samueru-sama:
> I'd like to remove completelly this patch, $PATH order is something that the user should choose.

We still need the warning, because otherwise if $HOME/.local/bin isn't in path that would result in appimages that can't be launched from the terminal. 

What I did will echo a warning with the optional question to patch the files. However right now the behaviour is that appman will not continue (`exit 1`) if the user says no.

Samueru-sama:
Oh wait I changed something I shouldn't have changed

ivan-hc:
@Samueru-sama we have a guide for AppMan https://github.com/ivan-hc/AM#how-to-install-appman

the behaviour of local configuration files shoul not be something we would decide.

also, what @rvenutolo said is that we could overcome this in other ways https://github.com/ivan-hc/AM/discussions/705

ivan-hc:
Let me see your function
```
function _patch_bashrc_and_profile() {
	if ! echo "$PATH" | grep -q "$BINDIR"; then
		read -rp "Warming: \"${XDG_BIN_HOME:-$HOME/.local/bin}\" is not in PATH, would you like appman to patch your .profile or .bash_profile? (Y/n): " yn
		if ! echo "$yn" | grep -i '^n' >/dev/null 2>&1; then
			if test -f "$HOME/.profile"; then
				printf '\n%s\n' 'export PATH="$PATH:${XDG_BIN_HOME:-$HOME/.local/bin}"' >> "$HOME/.profile"
			elif test -f "$HOME/.bash_profile"; then
				printf '\n%s\n' 'export PATH="$PATH:${XDG_BIN_HOME:-$HOME/.local/bin}"' >> "$HOME/.bash_profile"
			fi
			echo "You might need to resource your .profile / .bash_profile or reboot for these changes to take effect"
		else
			exit 1
		fi
	fi
}
```
web view sucks, I still have not understand how to split the views.

Your function looks good.

Samueru-sama:
> @Samueru-sama we have a guide for AppMan https://github.com/ivan-hc/AM#how-to-install-appman
> 
> the behaviour of local configuration files shoul not be something we would decide.
> 
> also, what @rvenutolo said is that we could overcome this in other ways #705

Alright if you don't want it, simply remove the if statements that check for `"$HOME/.profile"` and `"$HOME/.bash_profile"`

I don't agree with not having the option to do the patching, you have to be aware that there are people that have no idea what PATH is and won't bother trying to fix it if appman tells them that their `$HOME/.local/bin` is missing from it.

ivan-hc:
Something seems to be too strange, if an user using AppMan uses AppMan from ~/.local/bin... why should not have this in $PATH?

Samueru-sama:
> Something seems to be too strange, if an user using AppMan uses AppMan from ~/.local/bin... why should not have this in $PATH?

Sorry what?

ivan-hc:
> > Something seems to be too strange, if an user using AppMan uses AppMan from ~/.local/bin... why should not have this in $PATH?
> 
> Sorry what?

if an user is able to use something like `appman` from ~/.local/bin, why should need to add ~/.local/bin in $PATH?

should it be already in $PATH?

ivan-hc:
I think its better to write this in the installation guide and completelly remove that function from APP-MANAGER

ivan-hc:
> > should it be already in $PATH?
> 
> No! that's the big issue.
> 
> It actually happened to me when I started using linux, I couldn't get some scripts to work and it was because arch doesn't bother to add `$HOME/.local/bin` to PATH.

I agree to add it if not in place, OK.

ivan-hc:
Discord

PR #715, opened 2024-06-24
👀

Samueru-sama:
image

Let me know if you don't like something so I can roll it back.

Notice what I did to add `version` to libreoffice, on the libreoffice AM-updater I ended up adding this: 

```
VERSION=$("$APP" --version | awk '{print $2}' | grep ".")
echo "$VERSION" > ./ver.
```

which is the only way we can get the version from appimages with a fixed url.

### DIFF
--- a/modules/files.am
+++ b/modules/files.am
@@ -62,7 +62,7 @@ _files_if_script() {
 }
 
 _files_if_appimage() {
-	if ! strings -d "./$arg/$arg" >/dev/null 2>&1 | grep 'AppImages require FUSE to run' >/dev/null 2>&1; then
+	if ! strings -d "./$arg/$arg" | grep 'AppImages require FUSE to run' >/dev/null 2>&1; then
 		if grep "aisap-am" "$(command -v "$arg")" >/dev/null 2>&1; then
 			echo " ◆ $arg	|	appimage3🔒" >> "$AMCACHEDIR"/files-type
 		else
--- a/modules/files.am
+++ b/modules/files.am
@@ -154,7 +154,6 @@ _files() {
 	for arg in $INSTALLED_APPS; do
 		if test -f ./$arg/remove 2>/dev/null; then
 			APPVERSION=$(grep -w " ◆ $arg	|" "$AMCACHEDIR"/version-args | tr '	' '\n' | tail -1)
-			if [ -z "$APPVERSION" ]; then APPVERSION=$(cat ./"$arg/$arg/ver."); fi
 			APPTYPE=$(grep -w " ◆ $arg	|" "$AMCACHEDIR"/files-type | tr '	' '\n' | tail -1)
 			APPSYZE=$(grep -w " ◆ $arg	|" "$AMCACHEDIR"/files-sizes | tr '	' '\n' | tail -1)
 			echo " ◆ $arg	|	$APPVERSION	|	$APPTYPE	|	$APPSYZE" >> "$AMCACHEDIR"/files-args
--- a/modules/files.am
+++ b/modules/files.am
@@ -169,10 +169,10 @@ _files_show_only_number() {
 
 _files_total_size() {
 	if [ "$(_files_show_only_number)" = "0" ]; then
-		printf "\n TOTAL SIZE: $(du -ch $(find -type d -name 'modules' -printf "%h\n") 2>/dev/null | awk 'END {print $1"iB"}' | sed 's/...$/ &/') of disk space in use\n"
+		printf "\n TOTAL SIZE: $(du -ch $(find -type d -name 'modules' -printf "%h\n") 2>/dev/null | awk 'END {print $1"iB"}' | sed 's/...$/ &/') of disk space in use\n\n"
 	else
 		command -v aisap >/dev/null 2>&1 && printf '\n%s\n' " AppImages with 🔒 are sandboxed with aisap"
-		printf "\n TOTAL SIZE: $(du -ch $(find -type f -name 'remove' -printf "%h\n") 2>/dev/null | awk 'END {print $1"iB"}' | sed 's/...$/ &/') of disk space in use\n"
+		printf "\n TOTAL SIZE: $(du -ch $(find -type f -name 'remove' -printf "%h\n") 2>/dev/null | awk 'END {print $1"iB"}' | sed 's/...$/ &/') of disk space in use\n\n"
 	fi
 }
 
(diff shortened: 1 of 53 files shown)


ivan-hc:
> Let me know if you don't like something so I can roll it back.

Nope, I don't like it... I LOVE IT! THANKS!

> Notice what I did to add `version` to libreoffice, on the libreoffice AM-updater I ended up adding this:
> 
> ```
> VERSION=$("$APP" --version | awk '{print $2}' | grep ".")
> echo "$VERSION" > ./ver.
> ```
> 
> which is the only way we can get the version from appimages with a fixed url.

is the one normally checked an incorrect one? Also, I see that your libreoffice is type3, so your .zsync file will not work, this is why you have two different versions shown from the AppImage and the .zsync file. It is used in the official version right for this, to check the installed version.

I'd prefer to avoid searchin a version directly from the --version/-v of a program, you may launch it. I invite you to install the official `avidemux` and `avidemux-nightly` (the latter is not an AppImage) to understand.

ivan-hc:
> It would be nice the text on the type column was centered, example:
> 
> ```
> appimage3
>  binary
>  script
> ```

sure, do it as it were yours. :+1:

ivan-hc:
Just noticed that you added  a check for a "ver." file in "files.am". Consider that if you want to add this to the scripts, you must know the behaviour of all the apps you upload.

I still think that its unsafe to check version of info into any app using the common options. We need an external source.

Samueru-sama:
How do you get a version in libreoffice before? thru the zsync file?

I checked the `version` in libreoffice and it is a fixed url that only indicates if it is the standard, fresh, etc

I know that checking for `--version` doesn't always work with each application, because it may launch it as you say. that's not the case with libreoffice though. 

So I remove the check for `ver.` in `files.am`? I think you will eventually run into an app where there is no other way to check its version other than getting it from the app itself, and hopefully the app doesn't launch when that is done.

Note that `ver.` is created by the AM-updater instead of the module, this way prevent an issue that it might take too long to get it when `am -f` is ran. 

So yeah this means that each troublesome applications where version is unknown has to have its AM-updater edited carefully. hopefully of the 2000+ apps those are less than 50 or so though lol.

Samueru-sama:
Also why does `lf` show as unknown and other? lf is a binary and the version is `r32`

ivan-hc:
I noticed that you're using the Libreoffice Still, I have the Fresh... and yes, the .zsync file contains the version
```
ivan@debian:/opt/libreoffice$ strings -d ./*.zsync | grep -i "$(echo "libreoffice" | sed 's/-appimage//g')" | grep -Eo "([0-9]{1,}\.)+[0-9]{1,}" | head -1
24.2.4.2
ivan@debian:/opt/libreoffice$ libreoffice --version
LibreOffice 24.2.4.2 51a6219feb6075d9a4c46691dcfe0cd9c4fff3c2
```
but you need not to convert your AppImage to Type3  to made the .zsync file work.

ivan-hc:
> Also why does `lf` show as unknown and other? `lf` is a binary and the version is `r32`

I don't know, I've not checked yet the apps with "other" or stange version numbers.

Just today I added a new filter https://github.com/ivan-hc/AM/commit/84afb88eb764357a1dba1ffba7a4e7cc8788f914

Samueru-sama:
Oh I noticed that zsync file gets removed when I use `nolibfuse`. 

Why don't we instead rename the file to `$APP.zsync.disabled` and that way the files can still get the version? renaming the zsync file that way should also stop zsync from running.

ivan-hc:
> Oh I noticed that zsync file gets removed when I use `nolibfuse`.
> 
> Why don't we instead rename the file to `$APP.zsync.disabled` and that way the files.am can still get the version? renaming the zsync file that way should also stop zsync from running.

Uhm... you say that also if the app updates itself with version comparison, the version will still appear into the .zsync file?

I think its useless. The .zsync file will no more work. You need to use --rollback or reinstall the app to reuse the original one.

Samueru-sama:
> > Oh I noticed that zsync file gets removed when I use `nolibfuse`.
> > Why don't we instead rename the file to `$APP.zsync.disabled` and that way the files.am can still get the version? renaming the zsync file that way should also stop zsync from running.
> 
> Uhm... you say that also if the app updates itself with version comparison, the version will still appear into the .zsync file?
> 
> I think its useless. The .zsync file will no more work. You need to use --rollback or reinstall the app to reuse the original one.

I mean if the version is in the `version` file then we get it from there, that also means that the file updates by `if version = 0version` and it means everytime that `am -f` is used we also get the correct version.

If the version is instead is gotten from `$APP.zsync` and we disable the file by renaming, the app wont update itself, and `files.am` can check the `$APP.zsync.disabled` file and still report its version, which won't change because in this case the app has a fixed url and can't update by `if version = 0version`


It should be as simple as changing line 287 in management to `mv -f ./*.zsync ./*.zsync.disabled`

ivan-hc:
its OK for me. Do it.

ivan-hc:
@Samueru-sama maybe, since the final step asks you to remove the old file, why not to backup it?

We have a -b option. Ask the user if him want to keep a backup copy.

Samueru-sama:
> We have a -b option. Ask the user if him want to keep a backup copy.

By renaming the file we also back it up. If the user wants to restore it they have to rename it back. Which I don't see a point since it is very likely that even by reconvering the appimage back to type2 that zsync will still be broken. 

However I don't understand why we have to ask the user if they would prefer to delete it?, if it gets deleted then we will have `unkown` versions again?

We are preserving the file so that am -f doesn't result in unknown version.

ivan-hc:
> > We have a -b option. Ask the user if him want to keep a backup copy.
> 
> By renaming the file we also back it up. If the user wants to restore it they have to rename it back. Which I don't see a point since it is very likely that even by reconverting the appimage back to type2 that zsync will still be broken.
> 
> However I don't understand why we have to ask the user if they would prefer to delete it?, if it gets deleted then we will have `unkown` versions again?
> 
> We are preserving the file so that am -f doesn't result in unknown version, nothing else. I know it cannot be updated afterwards.

I've already covered a cache cleaning to have no data on versions, so they will be restored using (for example) the version file or something.

About ask to keeping the old version at the end, just remove it. We will add the -b option at begin.

Samueru-sama:
Alright time for discord because I don't understand

Samueru-sama:
Mmmmm I fucked up lol, let me rebase this and I'll pr again.

Issue #719, opened 2024-06-25
Apps


GeoBoi254:
archimede
https://github.com/gennaro-tedesco/archimede/releases

bunnyfetch
https://github.com/Rosettea/bunnyfetch/releases

freshfetch
https://github.com/K4rakara/freshfetch/releases

macchina
https://github.com/Macchina-CLI/macchina/releases

nitch
https://github.com/ssleert/nitch/releases

uwufetch
https://github.com/ad-oliviero/uwufetch/releases

DeepLX
https://github.com/OwO-Network/DeepLX/releases

mp4grep
https://github.com/o-oconnell/mp4grep/releases

ttyper
https://github.com/max-niederman/ttyper/releases

7-Zip
https://www.7-zip.org/download.html

FreeFileSync
https://github.com/hkneptune/FreeFileSync/releases

TestDisk & PhotoRec
https://www.cgsecurity.org/wiki/TestDisk_Download

kmon
https://github.com/orhun/kmon

qrscan
https://github.com/sayanarijit/qrscan/releases

tabby
https://github.com/Eugeny/tabby/releases

lan-mouse
https://github.com/feschber/lan-mouse/releases

ivan-hc:
I've created the last 11 of the list, @Samueru-sama the first 4.

I have some errors with "testdisk" when extracting the archive:
```
◆ "TESTDISK": starting installation script

-                       [ <=>                ]  42,78K  --.-KB/s    in 0,03s   
testdisk-7.2.linux2 100%[===================>]  19,32K  --.-KB/s    in 0s      
bzip2: (stdin) is not a bzip2 file.
tar: Child returned status 2
tar: Error is not recoverable: exiting now
chmod: impossibile accedere a '/home/ivan/.local/share/appman/testdisk/*_static': File o directory non esistente

 "TESTDISK" INSTALLED (1 MB OF DISK SPACE) 
```

ivan-hc:
We have reached 2100 installation scripts... but I need help with "testdisk", I got errors while extracting the archive.

ivan-hc:
I'm stuck

Samueru-sama:
The issue with testdisk is that the file that gets downloaded is corrupted, it can't extract it with ark either.

ivan-hc:
May this be an useragent issue?

Samueru-sama:
> May this be an useragent issue?

[...]

PR #721, opened 2024-06-25
Fix appimages and templates to allow nolibfuse to work correctly, changed files.am to indicate when an appimage is sandboxed

Samueru-sama:
There are like 5 or 4 appimages that I need to fix manually, will do so later.

### DIFF
--- a/modules/files.am
+++ b/modules/files.am
@@ -28,7 +28,7 @@ _files_sizes() {
 }
 
 _files_if_binary_executable() {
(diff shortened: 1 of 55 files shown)


Samueru-sama:
Alright I think that's all of them.

Samueru-sama:
I did the first 5 apps from #719 I will go to sleep, you can add the rest in the meantime if you want.

PR #723, opened 2024-06-25
Fix testdisk

Samueru-sama:
Btw Ivan check how I changed the remover, that is how the remover is meant to be used when new files have to be removed. 

Adding the extra files instead of the existing line is something one is likely to make a mistake when doing, that is why I have a commented out line for the .desktop as that is what is more likely to be the extra file that needs to be removed, but it also applies to all other extra files. 

Also you had the `chmod a+x "/opt/$APP/*static"` inside quotes which would never expand.


Samueru-sama:
@ivan-hc discord btw that I have some questions.

PR #725, opened 2024-06-26
Optimizations to files.am

Samueru-sama:


Samueru-sama:
Btw @ivan-hc so we don't forget. 

We need to fix the issue that `am -f` doesn't update after an application is sandboxed because the cache isn't cleared right after that is done.

ivan-hc:
> Btw @ivan-hc so we don't forget.
> 
> We need to fix the issue that `am -f` doesn't update after an application is sandboxed because the cache isn't cleared right after that is done.

Last tip before I go, so you can solve this.

After the process, add this command.
```
_remove_info_files
```
the function is in APP-MANAGER, this is what it does
```
function _remove_info_files() {
	rm -f "$AMCACHEDIR"/files*
	rm -f "$AMCACHEDIR"/version-args
}
```
without these files, we have no versions or info about type or anything. All info will be rebuilt when needed (-l, -f, -i... and so on).

Solved!

ivan-hc:
@Samueru-sama the above function is used in --rollback, remove, nolibfuse... now I don't remember other options.

So add `_remove_info_files` at the end of the "sandbox" process and you have solved

programs/x86_64/deadbeef
ivan-hc, line 1: can this script recognize the $AMCLI variable?

Samueru-sama: What do you mean? The program has been on main and working fine since like 2 days ago

Samueru-sama:
> So add `_remove_info_files` at the end of the "sandbox" process and you have solved

Done and it works! thanks.

programs/x86_64/deadbeef
ivan-hc: I mean, if you run the script using AM or AppMan, you may write the script like this
```
t can be sandboxed with $AMCLI --sandbox deadbeef'
```
instead of
```
it can be sandboxed with 'am --sandbox deadbeef' or 'appman --sandbox deadbeef'
```
I don't know if this will work

ivan-hc: I've just done the test by myself and does not worked

Istantanea_2024-06-26_14-40-17 png

pretend I've said nothin :shushing_face:

PR #728, opened 2024-06-27
Add support for BINDIR

Samueru-sama:
Do not rush this to main!


Samueru-sama:
image

Works perfect!

I defined `XDG_BIN_HOME=$HOME/Applications` I know that's a very weird location, but I did just to test that it works like it should and it did. 

The warning also worked, as when I defined `$XDG_BIN_HOME` I didn't add it to `$PATH`: 

image

PR #734, opened 2024-06-27
Small changes to the appimage template.

Samueru-sama:
Tested with librewolf and everything works fine.

Btw I changed the `.desktop` gets named. It is now `"$APP"-AM.desktop`

This is because I've noticed that if the application is named `AM-"$APP".desktop` and I want to look for a specific .desktop in my file manager, say the one for librewolf, I can't just type L on thunar, because the .desktop all start with A.

Also by default the `.desktop` will now get installed to `/usr/local/share/applications`


ivan-hc:
> Also by default the `.desktop` will now get installed to `/usr/local/share/applications`

I think its righ this, right now, also because my Debian system have changed something (maybe with a systemd update) and all my .desktop files are now in /usr/local/share instead of /usr/share

If Debian itself forces me to this, then I agree with Debian. And I agree with you for this change.

Samueru-sama:
> > Also by default the `.desktop` will now get installed to `/usr/local/share/applications`
> 
> I think its righ this, right now, also because my Debian system have changed something (maybe with a systemd update) and all my .desktop files are now in /usr/local/share instead of /usr/share
> 
> If Debian itself forces me to this, then I agree with Debian. And I agree with you for this change.

Yeah, `/usr/share` and `/usr/bin` are for the distro. While `/usr/local` and `/opt` are for add-on packages like AM.

Samueru-sama:
Btw I don't think it is needed to do a `rm -f /usr/local/share/applications/"$APP"-AM.desktop` and then `>> /usr/local/share/applications/"$APP"-AM.desktop` when you can just use a single `>` that will replace any existing file, but I didn't do the change because I don't know if you have a function that tells it to use sudo before removing the file or something.

ivan-hc:
the removal of a preexisting .desktop file was to prevent writing existing ones of failed installations... before I wrote the new install.am module. So yes, you can remove it, since install.am detects a previous installation exists.

ivan-hc:
@Samueru-sama why you added two dots? See my review.

Samueru-sama:
> @Samueru-sama why you added two dots? See my review.

I can't see the review. 

But you mean this? `printf "#!/bin/sh\nset -e\nrm -f /usr/local/bin/$APP\nrm -R -f /opt/$APP" > ../remove`

That's because we are in the `tmp` directory when the remover gets made, and that places the remover in  `/opt/$APP` instead of `/opt/$APP/tmp`.

And we are 100% that we are in tmp, because the step before is `[ -n "$APP" ] && mkdir -p "/opt/$APP/tmp" "/opt/$APP/icons" && cd "/opt/$APP/tmp" || exit 1` is that fails the script will not continue. 

I already tested the template by generating the librewolf script again and it works fine, **in fact it even appman placed the .desktop in my `$XDG_DATA_HOME/Applications`** which I didn't think it would do since I changed the path from `/usr/share/applications` to `/usr/local/share/applications`

ivan-hc:
So now I can merge, right? Or should you add something else?

Samueru-sama:
> So now I can merge, right? Or should you add something else?

I don't know if there is another mention of `AM-$APP.desktop` that needs to be changed to `$APP-AM.desktop`

And also I'm assuming that you have a function that patches `/usr/local/share/applications` for appman like I saw before, there has to be one because otherwise I don't know how I got the .desktop installed, but I'm just asking for confirmation lol.

ivan-hc:
> I don't know if there is another mention of `AM-$APP.desktop` that needs to be changed to `$APP-AM.desktop`

no, there isn't, I was checking it and changing the template.am for you... but you were faster

> And also I'm assuming that you have a function that patches `/usr/local/share/applications` for appman like I saw before, there has to be one because otherwise I don't know how I got the .desktop installed, but I'm just asking for confirmation lol.

this is the whole function responsible of the conversion, in install.am
```
function _install_appman_local_patch() {
	sed -i "s#/opt/#$APPSPATH/#g" ./"$arg"
	sed -i "s# /opt# $APPSPATH#g" ./"$arg"
	sed -i "s# /usr/local/bin# $BINDIR#g" ./"$arg"
	sed -i 's# "/usr/local/bin# "'"$BINDIR"'#g' ./"$arg"
	sed -i "s# /usr/bin# $BINDIR#g" ./"$arg"
	sed -i "s# /usr/local/games# $BINDIR#g" ./"$arg"
	sed -i "s# /usr/games# $BINDIR#g" ./"$arg"
	sed -i "s# /usr/share/applications# $DATADIR/applications#g" ./"$arg"
	sed -i "s# /usr/local/share/applications# $DATADIR/applications#g" ./"$arg"
	sed -i "s# /usr/share/pixmaps# $DATADIR/icons#g" ./"$arg"
	sed -i "s# /usr/local/share/pixmaps# $DATADIR/icons#g" ./"$arg"
}
```
if it is compatible with your script, I can merge

PR #747, opened 2024-07-02
Implement "torsocks" and add filters for non x64 apps  in template.am / update more installation scripts

ivan-hc:


ivan-hc:
@Samueru-sama since we have many "`REPO="$user/$project"`", how to quich determine if a fariable looks like this
```
$user/$project
```
?

I'd like to add an extra condition to determine if a github URL is in use, at
```
						if echo "$RESPONSE" | grep -q "https://github.com"; then
							_template_then_github_source
						elif echo "$RESPONSE" | grep -q "sourceforge"; then
							_template_if_sourceforge
						else
							_template_if_hosted_elsewhere
						fi
```
to convert scripts quickly

Samueru-sama:
you might also want to add `windows/|apple/|darwin` to the filters. 

I wouldn't really add `i386\|i686` to the filter as it is highly unlikely that we will see such appimages in the future, but that is up to you. I'm not saying that it will never happen again, but I'm saying that at most we might see such case once or twice a year 😆

It would also be great if the script for `aarch64` gets made automatically even if the app doesn't exist yet, as one can just quickly check if version returns a valid `aarch64` url and add it as well to the list of apps. 

In that case the filter has to remove `x86_64` `x64` `amd64` and `x86-64` you can likely use extended regexes in grep, something like `grep -viE "(amd|86-|86_|x)64"` for example iirc. 

>@Samueru-sama since we have many "REPO="$user/$project"", how to quick determine if a variable looks like this

I'm not sure what you want to do with that. When I'm making an script I just paste the full github url, in that case templates can grep for `github` and use the template for github. Same check can be done for other websites.

Samueru-sama:
Just in case if the filter is just for appimages, you don't need to add `windows/|apple/|darwin` as there are no such appimages lol. However that is needed when it is an archive.

ivan-hc:
> I'm not sure what you want to do with that. When I'm making an script I just paste the full github url, in that case templates can grep for `github` and use the template for github. Same check can be done for other websites.

I open the old scripts in first, and then I copy/paste `user/project`, because many old scripts are wrote like this

at moment I use the `-t` option, it prompts me to add the URL, and I have to write/rewrite/paste `https://github.com/` always and always.

as a workaround I've disabled all called functions eccept "`_template_then_github_source`" so I'm sure that I'm creating scripts for github with all dedicated functions of the module.

---------------------------------------

about ecceptions, I only added `grep -vi "i386\|i686\|aarch64\|arm64\|armv7l"` to the command, because there are a lot of projects that are prompting also i386 and aarch64.

I'm working for the x86_64 only because the workload is a lot. Those interested in adding also apps for their architecture may work only for that.

ivan-hc:
I've not added `windows/|apple/|darwin` anywhere

Samueru-sama:
> I open the old scripts in first, and then I copy/paste `user/project`, because many old scripts are wrote like this

Most of those are github, something like +1900 apps are from gtihub. If you are converting old scripts then I don't see much point in trying to automate that, because the scripts that aren't from github are very few and you can just group them together in your file manager to make sure by simply searching for the ones tha contain github. 

>about ecceptions, I only added grep -vi "i386\|i686\|aarch64\|arm64\|armv7l" to the command, because there are a lot of projects that are prompting also i386 and aarch64

Okay.

>I'm working for the x86_64 only because the workload is a lot.

Is it? what I said is basically have something that replaces the `grep -vi "i386\|i686\|aarch64\|arm64\|armv7l"` for something that does the opposite for aarch64 and that way the script is already generated, one just has to check that the `aarch64` actually exists (check the url) and that is it.

ivan-hc:
I have solved
```
						if echo "$RESPONSE" | grep -q "https://github.com"; then
							_template_then_github_source
						elif echo "$RESPONSE" | grep -q "sourceforge"; then
							_template_if_sourceforge
						elif ! echo "$RESPONSE" | grep -q "://"; then
							_template_then_github_source
						else
							_template_if_hosted_elsewhere
						fi
```

PR #748, opened 2024-07-03
Update installation scripts: part 1

ivan-hc:
The first commit is about all portable apps from github, that were the ones we need to test the more.

NOTE, only portable apps, next commits are all about AppImages, that are the easier to update like this.

PR #752, opened 2024-07-03
Update installation scripts: part 2

ivan-hc:
Continuation of https://github.com/ivan-hc/AM/pull/748

Here I'm updating installation scripts for AppImages hosted on github.com

I'm checking the URLs one by one, to make sure I don't change anything keyword-wise and if the scripts point to "latest" instead of a generic "releases"

I'm proceeding in alphabetical order, and I've currently reached the letter F

If any of you want to take charge of a certain group of subsequent installation scripts, for example, the letter S or R, or T or whatever you prefer, but so as to avoid doing the same job twice, it would be of big help for me.

I'm working within github API connection limits, so I can upload between 20 and 40 scripts per hour on average.


ivan-hc:
1141 left

Samueru-sama:
> I'm checking the URLs one by one, to make sure I don't change anything keyword-wise and if the scripts point to "latest" instead of a generic "releases"

I think this can be automated something, basically the older link gets stored and compared to the newer link, and if they both match then it is good, if they don't match an error is given.

ivan-hc:
> I think this can be automated something, basically the older link gets stored and compared to the newer link, and if they both match then it is good, if they don't match an error is given.

and I want to prevent errors. This is what I call "revision" ;)

ivan-hc:
1085

ivan-hc:
@Samueru-sama  a screenshot on how I'm working, now I've to wait 30 minutes/1 hour before I start again

Istantanea_2024-07-03_19-13-55 png

PR #759, opened 2024-07-04
Unify braves

Samueru-sama:


ivan-hc:
@Samueru-sama those that want a specific version of brave would install only that, normally èpeople want the stable release. Why should them choose from a script to select the version? This is why they are all listed like this.

ivan-hc:
Also, Brave, Firefox, Thunderbird... are portable apps, and they are very mainstream, choosing an Appimage of them is optional.

Leave them as they are.

I think that we have a lot of "wine" and "gimp" or "vlc (stable/oldstable)" instead, why not to remove them?

Samueru-sama:
> @Samueru-sama those that want a specific version of brave would install only that, normally èpeople want the stable release. Why should them choose from a script to select the version? This is why they are all listed like this.

You still have the stable release, only that the scripts asks you if you want it as an appimage or the portable build. 

Same is done for the nightly and beta versions.

ivan-hc:
its OK for brave, but please don't do the same with Firefox

PR #769, opened 2024-07-07
Update some random apps. ES-DE IS BROKEN!

Samueru-sama:
es-de broke again, I mean the original way the website was being parsed, I think I will try to contact the devs because this thing has broken 3 times in less than 6 months already.


ivan-hc:
what you mean with broken? The app itself?

Samueru-sama:
> what you mean with broken? The app itself?

The original way the website was being parsed doesn't work.

PR #774, opened 2024-07-08
Road to 7: part 4, THE END!

ivan-hc:
- 99% of applications are updateable
- all applications will install the launcher with the following path
```
/usr/local/share/applications/$PROGRAM-AM.desktop
```
- removed many apps again, after making sure the download URLs were still good or not


ivan-hc:
Now we can go back to innovating "AM"

PR #775, opened 2024-07-08
Renamed 78 apps

ivan-hc:
Removed "-appimage" suffix from 78 installation scripts for AppImage packages.


ivan-hc:
@Samueru-sama these are the apps that have not been renamed

Istantanea_2024-07-08_06-37-16

ivan-hc:
- nordvpn-appimage require nordvpn installed
- qemu-appimage is not an official solution
- supertuxkart-appimage is mine, and the original one is standalone and available as supertuxkart
- podman-appimage because the real podman already exists
- about thunderbird and waterfox, I don't care, if you feel for it, rename them
- firefox, I prefer to leave the two things separated, they are not the same thing

claro amgigo?

ivan-hc:
Now I go sleep, tomorrow I'll workon the README and other things to be perfected before the new release

PR #779, opened 2024-07-09
Create AppImages on the fly!

ivan-hc:
The script to install the AppImages is patched to download another script from this repository which is intended to create the AppImage on the fly.

In this example, "calibre"

https://github.com/ivan-hc/AM/assets/88724353/05d7146b-853a-44d4-ac39-35c26a06895c

NOTE, the scripts points on main, I had to adit it to dev because this is still in the dev branch

@Samueru-sama

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="7"
+AMVERSION="7.1"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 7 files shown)


ivan-hc:
For now I can't add script testing, I first need to have at least one app of this type in "main" to run the test.

I will release this feature in version 7.1.1

ivan-hc:
The latest commit includes checking for such dependencies for on-the-fly AppImage package creation:
- `ar`
- `convert`
- `glib-compile-schemas`
- `tar`
- `zip`

The limit will be for all scripts of this type only in version 7.1, and currently "calibre" is the only one.

In the version even later, the check will be directly on the available scripts, checking if the URL for download exists and if the single keywords are contained in that script, directly.

So that, for example, if we install "calibre", the only requirement will be "`tar`", since other commands listed above are not necessary.

ivan-hc:
Or at least, I would like this to be applicable only for `convert` and `glib-compile-schemas`, contained in `imagemagick` and `libglib2.0-bin` or `glib2` respectively (depends on the distribution).

ivan-hc:
This is how actually looks

https://github.com/ivan-hc/AM/assets/88724353/caba0ebd-d976-4bdb-888a-caa56c57d3d6

PR #792, opened 2024-07-17
Update install.am: divide installation cases into functions

ivan-hc:
The installation cases have been divided into functions, so as to be able to analyze the individual cases one by one.

some commands have been simplified, in order to remove duplicates or unnecessary "waltzes" in the directories where the installation process is carried out.

### DIFF
--- a/modules/install.am
+++ b/modules/install.am
@@ -49,20 +49,20 @@ function _post_installation_processes() {
 	# Add a command to line 2 of the "remove" script to prevent use without root privileges
 	# and take control of the directory to be able to manage and update the app without root privileges
 	if [ "$AMCLI" == am ] 2>/dev/null; then
-		$SUDOCOMMAND sed -i '1 a if [ "$(id -u)" -ne 0 ]; then echo "Permission denied"; exit 1; fi' "$APPSPATH"/"$(cd "$APPSPATH" && ls -td * | head -1)"/remove 2> /dev/null
-		$SUDOCOMMAND chown -R $currentuser "$APPSPATH"/"$(cd "$APPSPATH" && ls -td * | head -1)" 2> /dev/null
+		$SUDOCOMMAND sed -i '1 a if [ "$(id -u)" -ne 0 ]; then echo "Permission denied"; exit 1; fi' "$APPSPATH"/"$(ls -td "$APPSPATH"/* | head -1 | sed 's:.*/::')"/remove 2> /dev/null
+		$SUDOCOMMAND chown -R $currentuser "$APPSPATH"/"$(ls -td "$APPSPATH"/* | head -1 | sed 's:.*/::')" 2> /dev/null
 	fi
 	# Check if an AM-updater script exists, so the CLI can manage updates for the installed app
 	# by comparing the hosted script with a downloaded one
-	if test -f "$APPSPATH"/"$(cd "$APPSPATH" && ls -td * | head -1)"/AM-updater; then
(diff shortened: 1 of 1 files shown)


ivan-hc:
@Samueru-sama in your specific case, the function related to your issue is "_install_local_script"

Here are listed the two functions responsible of local scripts:
```
# This function is for installation scripts run in place, using dot
function _install_local_script() {
	cp ./"$arg" "$AMCACHEDIR"/tmp/"$arg"
	cd "$AMCACHEDIR" && mv ./tmp/"$arg" ./"$arg" && rmdir ./tmp || return
	_install_arg
}

# This function is always for local install scripts, but with the entire path as argument
function _install_local_script_by_following_a_path() {
	path2arg=$(echo "$arg")
	arg=$(echo "$path2arg" | sed 's:.*/::')
	cp "$path2arg" "$AMCACHEDIR"/tmp/"$arg"
	cd "$AMCACHEDIR" && mv ./tmp/"$arg" ./"$arg" && rmdir ./tmp || return
	_install_arg
}
```

and now I'm finally able to reproduce the issue, but in a different situation (after installing it for the first time).

The first "ERROR" message is the wrong one, the second is the exact one.

Istantanea_2024-07-17_17-52-31

ivan-hc:
This is really strange. Why the first time I retry to install the same script I get the wrong error message?

ivan-hc:
By adding this condition
```
elif echo "$arg" | grep "^./" 2> /dev/null; then
	_install_local_script
```
I get this
```

![Istantanea_2024-07-17_18-07-34](https://github.com/user-attachments/assets/f9b20664-f47b-4502-8e3f-d93baa9d0ffe)

**Comment by ivan-hc** [comment]
@Samueru-sama still another stress-test

Istantanea_2024-07-17_18-19-38

**Comment by ivan-hc** [comment]
For some reasons, the time I retry to install the same local script followed by a dot, it appears as "././adb.sh" instead of "./adb" but the argument is the same listed

**Comment by ivan-hc** [comment]
to use one function for both kind of local scripts is enough
```
# This function is for local installation scripts
function _install_local_script() {
	path2arg=$(echo "$arg")
	arg=$(echo "$path2arg" | sed 's:.*/::')
	cp "$path2arg" "$AMCACHEDIR"/tmp/"$arg"
	cd "$AMCACHEDIR" && mv ./tmp/"$arg" ./"$arg" && rmdir ./tmp || return
	_install_arg
}
```
but it still does not solve the issue

ivan-hc:
@Samueru-sama I added a `pwd` and this is the result

Istantanea_2024-07-17_21-21-51

why this happens

Samueru-sama:
> This is really strange. Why the first time I retry to install the same script I get the wrong error message?

you tell me 🤣

ivan-hc:
@Samueru-sama just noticed that each time an installation of multiple arguments starts, only the first installation shows the correct path

Istantanea_2024-07-17_21-38-03

this is why the last one is not installed

ivan-hc:
@Samueru-sama I've tried by adding an environment variable "STARTING_DIR=$(pwd)" to use at start of the installation and to "cd" in at the end of each process... and the last argument ./adb.sh has been installed without problem... and also `pwd` always shows my ~/Scrivania directory (the desktop).

But again, when I triad with `am -i ./adb.sh`, I got the same error the first time, and not all the other times. Still `pwd` have shown /opt the first time, and no the others.

Its a really strange behaviour. I can't explain why this happens.

Issue #653, opened 2024-06-11
Sometimes running am -i on a local script fails the first time


Samueru-sama:
Hey, this is an issue that I've noticed has been happening ever since I made the mpv appimage: 

![image](https://github.com/ivan-hc/AM/assets/36420837/b27c1f8e-8ba0-4484-b278-450731f925e9)

Sometimes am tires to pull from the database instead of using the script I'm telling it use, if I didn't rename the script to have the `.sh` at the end of the name, it would have used the existing `cpu-x` script on the database instead of the local one that I'm telling it to use. 

(am is appman btw).

Samueru-sama:
@ivan-hc I'm testing cpu-x again because the dev said to have fixed the issue, but it doesn't look like it works 😭

ivan-hc:
I'm testing this ar appman in normal mode, no issue here

![Istantanea_2024-06-11_17-05-15 png](https://github.com/ivan-hc/AM/assets/88724353/58ef4e9a-f601-4be4-9151-2707fde394da)

Samueru-sama:
> I'm testing this ar appman in normal mode, no issue here
> 
[...]

PR #793, opened 2024-07-18
Add "$REALDIR" variable to "APP-MANAGER" and "install.am" to keep the shell in the current directory

ivan-hc:
Fix https://github.com/ivan-hc/AM/issues/653

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -7,7 +7,7 @@ AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
 AMBRANCH=$(basename "$AMREPO")
 
 # Determine the name of this script and its working directory
-export REALDIR=$(pwd)
+export REALDIR="$PWD"
 DIR="$( cd "$( dirname "$0" )" && pwd )"
 CLI=$(basename "$0")
 
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -7,6 +7,7 @@ AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
 AMBRANCH=$(basename "$AMREPO")
 
 # Determine the name of this script and its working directory
+export REALDIR=$(pwd)
 DIR="$( cd "$( dirname "$0" )" && pwd )"
 CLI=$(basename "$0")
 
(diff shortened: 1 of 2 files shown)


Samueru-sama:
I think `REALDIR=$(pwd)` can just be `REALDIR=$PWD` 

Because as far as I know the shell already sets the `$PWD` env variable, so maybe it is not needed to be calling pwd to know that?

I haven't tested it though.

ivan-hc:
> I think `REALDIR=$(pwd)` can just be `REALDIR=$PWD`
> 
> Because as far as I know the shell already sets the `$PWD` env variable, so maybe it is not needed to be calling pwd to know that?
> 
> I haven't tested it though.

tested, and worked

ivan-hc:
we have another issue now: priority!

see what happens if in the PWD directory exists a file with the name of an argument

Istantanea_2024-07-18_16-10-05

ivan-hc:
by inverting the priority order, the module will check hosted arguments before the local ones
```
					if test -f "$AMPATH/neodb"; then
						_install_from_third_party_repo
					elif curl --output /dev/null --silent --head --fail "$APPSDB"/"$arg"  1>/dev/null; then
						_install_normally
					elif test -f "$arg" 2> /dev/null; then
						_install_local_script
					else
						echo " 💀 ERROR: \"$arg\" does NOT exist in the database, see \"$AMCLI -l\""
					fi
```

Istantanea_2024-07-18_16-14-00

ivan-hc:
this will conflict with our style of tests: I should rename my scripts to .sh as you do

its not the way I wanted to made this to work for

ivan-hc:
the last commit solves all issues

her I added an `echo "$arg"` to show the name of the argument

Istantanea_2024-07-18_17-16-12

ivan-hc:
and to be sure that "chromium" is not installed using other cases, I disabled tadabases in this test and this is the result

Istantanea_2024-07-18_17-19-15

ivan-hc:
the last commit adds a message in case of missing file

Istantanea_2024-07-18_17-56-07 png

ivan-hc:
Just tested on Fedora and works.

PR #827, opened 2024-08-09
Add initial doas support

Samueru-sama:
Btw Ivan I noticed this and I think there is an error in this function: 

```
function _no_sudo() {
	if [ -n "$SUDO_COMMAND" ]; then
		printf '\n Please do not use "sudo" to execute "'"$CLI"'", try again.\n\n'
		exit 1
	fi
}
```

What is `$SUDO_COMMAND`? I don't see it being defined anywhere, instead there is `$SUDOCOMMAND`

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -59,7 +59,7 @@ function _am() {
 	elif command -v doas >/dev/null 2>&1; then
 		SUDOCOMMAND="doas"
 	else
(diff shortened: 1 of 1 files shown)


ivan-hc:
@Samueru-sama https://www.baeldung.com/linux/sudo-manage-environment-variables#1-sudo-variables

PR #839, opened 2024-08-12
Optimizations to `database.am`

Samueru-sama:
I wasn't able to get the colors on the second column to work, ran into the same `ESC` issue as you.

I made `less` optional and also fixed an issue that if you exited less you would still get the full list of files printed on the terminal by running a `clear` after `less`.

There is likely some more work to be done here, specially with the very long lines that are hard to read, I spent a while trying to understand what the long `printf` for list were for only to realize that there was a lot of duplicated code in them after I broke them down into pieces.

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -10,6 +10,11 @@ _pretty_list() {
 	sed -E 's#(http|https|ftp)://[^ ]*##g; s#(SITE|SOURCE):##g; s/$/\n/g' | fold -sw 75 | sed 's/^/   /g; s/  ◆ /◆ /g'
 }
 
+_pretty_list_compat() {
+	# Removes references to URLs, "-a" elements in "-l" and "-q"
+	sed -E 's#(http|https|ftp)://[^ ]*##g; s#(SITE|SOURCE):##g' | fold -sw 75 | sed 's/^/   /g; s/  ◆ /◆ /g' && printf '\n'
+}
+
 _download_markdown() {
 	disk_usage=""
 	app_version=""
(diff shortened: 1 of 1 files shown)


ivan-hc:
why you removed less??? Not all terminals are configured to show the whole list or more than 200 lines. Also DistroTube faced this issue in the review, and you removed `less`

Now -l have no more an header or a title

Also, why you use clear? Not everybody want to remove previous logs.

Nope, this PR must reset some defaults.

PS: for me its ok not to show installed apps. It ir a task of -f after all.

ivan-hc:
No title on the lists to made people aware of the number of available items

https://github.com/user-attachments/assets/02ec7710-e474-4a8b-abcb-6a5b813388d6

and the terminal is alco cleared, so I can't see previous logs.

I don't like this behaviour.

ivan-hc:
if I remove the "clear" this is what I get

Istantanea_2024-08-12_16-43-07

ivan-hc:
I have copied and edited the first function to made it have a different behaviour fow "queries", I noticed that a list of keywords may bee too much big using "_pretty_list"

Now I'd like to fix -l in the meantime you are not here.

ivan-hc:
just found that the way you set up the message on -l in the case statement is not included in "less"

Samueru-sama:
> why you removed less???

Didn't say that.

>Now -l have no more an header or a title

Then that would be a bug in my changes because that shouldn't happen, I compared the script before and after and didn't see any difference. The only issue was that after exiting `less` my terminal still showed the full list of applications, that is why I added the clear after the less.

ivan-hc:
> Then that would be a bug in my changes because that shouldn't happen, I compared the script before and after and didn't see any difference. The only issue was that after exiting `less` my terminal still showed the full list of applications, that is why I added the clear after the less.

yes, it was an issue of "less" in the end

to let the header be in "less", the whole message must be there, and you added an "if" function out of the target of "less"

I have solved this way
```
	'-l'|'list')
		_list
		if [ "$2" = --appimages ]; then
			_list_appimages
			LIST=$(sort "$AMCACHEDIR/$arch-appimages" | _pretty_list)
			printf "\n$MESSAGE\n$MESSAGE2\n LIST OF THE $AVAILABLE_APPIMAGES_NUMBER APPIMAGES AVAILABLE IN THE 'AM' REPOSITORY:\n$LIST\n" | less -I
		else
			LIST=$(sort "$AMPATH/$arch-apps" | _pretty_list)
			printf "\n$MESSAGE\n$MESSAGE2\n LIST OF THE $AVAILABLE_APPS_NUMBER APPLICATIONS AVAILABLE IN THE 'AM' REPOSITORY:\n$LIST\n" | less -I
		fi
		#printf "\n$MESSAGE\n$MESSAGE2\n"
		;;
```

plus other variables dedicated

Issue #826, opened 2024-08-09
sudo: please make it optional


vitaly-zdanevich:
Also because I want to package it for Gentoo Linux - so it already will be installed with root.

ivan-hc:
@Samueru-sama I need your help here, I know you use "doas"

ivan-hc:
While using "AM", this is the function that made it work with "sudo"

```
function _am() {
	AMCLI="am"
	AMCLIPATH="$AMCLI"
	SUDOCOMMAND="sudo"
	COMPLETIONPATH="/etc/bash_completion.d"
	COMPLETIONFILE="am-completion.sh"
	APPSPATH="/opt"
	AMPATH="$APPSPATH/$AMCLI"
	_create_cache_dir
	mkdir -p "$AMPATH"/modules
}
```
https://github.com/ivan-hc/AM/blob/4d30b36ff2a17adde3f8f593f1236b195ff1a3ab/APP-MANAGER#L54C1-L64C2

Samueru-sama:
I'm a doas user, however I use AppMan instead of AM which doesn't need elevated privileges. A while ago I did some changes that makes AM depend less on sudo, but if I'm not mistaken it still isn't ready.

@vitaly-zdanevich https://github.com/ivan-hc/AM/pull/827

Once that is merged to dev, please enable the developer mode and update AM and let me know if you still have issues without sudo.

ivan-hc:
@vitaly-zdanevich do `am --devmode-enable` and then, update with `am -s`

ivan-hc:
@Samueru-sama discord

Samueru-sama:
Btw @ivan-hc the same if command -v sudo else command -v doas has to be done in the INSTALL script.

ivan-hc:
> Btw @ivan-hc the same if command -v sudo else command -v doas has to be done in the INSTALL script, I will do it tomorrow, but if you want to do it now do it.

done https://github.com/ivan-hc/AM/commit/fa2dc8b9cbae46adc13c535f947e25aad384905b and https://github.com/ivan-hc/AM/commit/249a3c8957b80c269a8870ad68d33dd8b6704fd2

also updated the README about one hour ago https://github.com/ivan-hc/AM/commit/839ad9fd32706d25b2178c8269da16a805d00f61

in my use case also the installer worked like a charm on the Fedora VM.

vitaly-zdanevich:
Oh my, I do not know what `doas` is...

[...]

PR #847, opened 2024-08-15
"AM": made "sudo" and "doas" optional (default is "root")

ivan-hc:
fix https://github.com/ivan-hc/AM/issues/826

This is the common code in APP-MANAGER and AM-INSTALLER added in this pull request
```
	if [ "$(id -u)" = 0 ]; then
		SUDOCOMMAND=""
	elif command -v sudo >/dev/null 2>&1; then
		SUDOCOMMAND="sudo"
	elif command -v doas >/dev/null 2>&1; then
		SUDOCOMMAND="doas"
	else
		echo "ERROR: you need to be \"root\" to use \"AM\""
		exit 1
	fi
```

1. all the function above does is to check if the user that is running AM is "root"
2. if not, check if the "sudo" command exists, so you can use it
3. alternativelly, use "doas"
4. if none of the options above is satified, you cannot use AM at all

so the new command added here is
```
if [ "$(id -u)" = 0 ]; then
	SUDOCOMMAND=""
```

all others were already been added by @Samueru-sama and available from the "AM" 7.6 [release](https://github.com/ivan-hc/AM/releases/tag/7.6)

----------------------------

## Test 1: install and use "AM" as the "root" user
Tests to be done (at least in a Virtual Machine):
- login as root and install AM using the AM-INSTALLER script, option 1
- once AM is installed, do not run it!
- in first you must change the content of /opt/am/APP-MANAGER (this is the one from "main") with the one from this PR, in "dev"
- use "AM" normally, still using the "root" session

tests to do after:
- try to use AM also with other admins, "AM" should not work at all
- using other admins, open a root shell and use "AM"

NOTE: in this test, the owner of /opt/am is "root"

----------------------------

## Test 2: install and use "AM" as regular user
Tests to be done (at least in a Virtual Machine):
- install AM using AM-INSTALLER, option 1, as a regular user, and enter the password normally
- once AM is installed, do not run it!
- in first you must change the content of /opt/am/APP-MANAGER (this is the one from "main") with the one from this PR, in "dev"
- use "AM" normally

tests to be done after:
- try to use AM as "root", also by login the "root" account
- try the same with other users

### DIFF
--- a/AM-INSTALLER
+++ b/AM-INSTALLER
@@ -27,12 +27,14 @@ _install_am() {
 	mkdir -p "$CACHEDIR" || true
 	rm -f "$CACHEDIR"/INSTALL-AM.sh || true
 	wget -q https://raw.githubusercontent.com/ivan-hc/AM/main/INSTALL -O "$CACHEDIR"/INSTALL-AM.sh && chmod a+x "$CACHEDIR"/INSTALL-AM.sh
-	if command -v sudo >/dev/null 2>&1; then
+	if [ "$(id -u)" = 0 ]; then
+		SUDOCOMMAND=""
+	elif command -v sudo >/dev/null 2>&1; then
 		SUDOCOMMAND="sudo"
 	elif command -v doas >/dev/null 2>&1; then
 		SUDOCOMMAND="doas"
 	else
-		echo 'ERROR: No sudo or doas found'
+		echo "ERROR: you need to be \"root\" to use \"AM\""
 		exit 1
 	fi
 	$SUDOCOMMAND "$CACHEDIR"/INSTALL-AM.sh && rm -f "$CACHEDIR"/INSTALL-AM.sh
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -54,12 +54,14 @@ _no_sudo
 function _am() {
 	AMCLI="am"
 	AMCLIPATH="$AMCLI"
-	if command -v sudo >/dev/null 2>&1; then
+	if [ "$(id -u)" = 0 ]; then
+		SUDOCOMMAND=""
+	elif command -v sudo >/dev/null 2>&1; then
 		SUDOCOMMAND="sudo"
 	elif command -v doas >/dev/null 2>&1; then
 		SUDOCOMMAND="doas"
 	else
-		echo 'ERROR: No sudo or doas found'
+		echo "ERROR: you need to be \"root\" to use \"AM\""
 		exit 1
 	fi
 	COMPLETIONPATH="/etc/bash_completion.d"


ivan-hc:
installing "AM" is gone well, but using "AM" to install applications, don't

at the end of the video below I must remove the "if" condition added in this pull request to install applications

https://github.com/user-attachments/assets/c8d25def-5e1e-4fdf-b02d-d5b6b7fec8c9

this is strange, $SUDOCOMMAND should act as an empty and dummi variable, so the installation should proceede.

Also, it does not show any message that redirects me on the right way.

However, I'm using "AM" with this change on my host machine and I have no issues, because I'm not the "root" user.

ivan-hc:
its trange that I'm instead able to remove the apps, not to install them, as "root" user

ivan-hc:
@vitaly-zdanevich any idea?

as "root" user I can't install applications
```
	if [ "$(id -u)" = 0 ]; then
		SUDOCOMMAND=""
	elif command -v sudo >/dev/null 2>&1; then
		SUDOCOMMAND="sudo"
	elif command -v doas >/dev/null 2>&1; then
		SUDOCOMMAND="doas"
	else
		echo "ERROR: you need to be \"root\" to use \"AM\""
		exit 1
	fi
```
but if I change the above like this
```
	if command -v sudo >/dev/null 2>&1; then
		SUDOCOMMAND="sudo"
	elif command -v doas >/dev/null 2>&1; then
		SUDOCOMMAND="doas"
	else
		echo "ERROR: you need to be \"root\" to use \"AM\""
		exit 1
	fi
```
the $SUDOCOMMAND variable is "sudo", so I can install programs because the variable is no more empty

ivan-hc:
also if I place "root" as third choice, like this
```
	if command -v sudo >/dev/null 2>&1; then
		SUDOCOMMAND="sudo"
	elif command -v doas >/dev/null 2>&1; then
		SUDOCOMMAND="doas"
	elif [ "$(id -u)" = 0 ]; then
		SUDOCOMMAND=""
	else
		echo "ERROR: you need to be \"root\" to use \"AM\""
		exit 1
	fi
```
this will not solve the fact that as "root" I cannot install anything withour "sudo" or "doas" installed

ivan-hc:
command "`sh`", or "`bash`"

found the solution

ivan-hc:
change of program:

since the reason because the issue was opened was the packaging of AM for Gentoo, I think that "sudo" or "doas" will keep be dependences for "AM" itself.

A package for repositories being executed as ROOT, needs to use the INSTALL script from the repository.

That said, usage of "AM" and its installation via AM-INSTALLER have non sense while the program is installed via system package manager.

I will close this pull request and remove the changes.

PR #849, opened 2024-08-15
Introduce check for spooky scripts, and some refactoring

### LINKED ISSUE #820: About non-official apps.
Following @probonopd comments we have to add a way for "mere mortals" to be able to know which third-party provided AppImages are high-quality and trustworthy.

Thankfully, I think over 90% of the appimages in AM are all hosted on github and have a transparent build process behind them.

So I think what needs to be done is: 

- We the appimage is not official but has a transparent build process behind, like this [example](https://github.com/srevinsaju/Firefox-Appimage) we don't need to issue any warning, just point out that it is not official in the description.
- If the appimage is just a random uploaded binary, [like in this example](https://github.com/AlexanderP/tesseract-appimage), we need to add a warning before installation.

I think the simple solution is adding a comment like `# spooky` to the installation script, and AM should check for that word before giving the warning.

Samueru-sama:
I tested installing some applications and everything seems to work.

**If you find a problem let me know what the problem is, don't try to debug it yourself.**

closes #820

### DIFF
--- a/modules/install.am
+++ b/modules/install.am
@@ -62,12 +62,10 @@ _post_installation_processes() {
 	# Check for AM-updater script sothat CLI can manage updates
 	[ -f "${LASTDIRPATH}"/AM-updater ] && mkdir "${LASTDIRPATH}"/.am-installer 2>/dev/null \
 		&& wget -q "$APPSDB"/"$arg" -O "${LASTDIRPATH}"/.am-installer/"$arg"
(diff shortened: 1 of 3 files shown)


ivan-hc:
really good, @Samueru-sama , thanks

https://github.com/user-attachments/assets/cac9429a-3505-4e3e-9ce8-44712d2a2ac2

PR #851, opened 2024-08-16
Remove zsync from template

Samueru-sama:
The next step is adding a check to use appimageupdatetool with a fallback to the AM-updater.

I did not remove the `rm -R -f ./*zs-old ./*.part` because those files can remain if the zsync update fails half way and then the AM-updater would have to remove them.


ivan-hc:
remember to keep the spaces at same level ar before, since the function for github apps in -t is the same for Appimages and other apps

PR #862, opened 2024-08-21
refactor sandboxes.am, allow passing multiple args

Samueru-sama:
I did basic testing and it works.

I only need to make sure that the options that allow access to certain directories work, which I will do later.


Samueru-sama:
All that's left is making sure that it works with AM as I only tested it with AppMan

Issue #903, opened 2024-08-30
move bash_completion file to $XDG_CONFIG_HOME"/bash-completion


prodigalshadow:
Hi, first of all, thank you for your hard work! As the title says, would it be possible to move the bash_completion file to $XDG_CONFIG_HOME"/bash-completion/bash_completion to make it compliant with XDG Base Directory Specification? Thanks!

ivan-hc:
Hi, my cooperator had issues while trying to use it via ZSH, this is why we left it in $HOME. Anyway I also want to find a way not to add dotfiles in the $HOME like this.

ivan-hc:
I just tried to replace
```
# BASH AND ZSH COMPLETION
completion_file="${ZDOTDIR:-$HOME}"/.bash_completion
```
with
```
# BASH AND ZSH COMPLETION
mkdir -p "${ZDOTDIR:-$XDG_CONFIG_HOME}/bash-completion"
completion_file="${ZDOTDIR:-$XDG_CONFIG_HOME}/bash-completion/bash_completion"
```
but completion does not work for me (I use BASH)

ivan-hc:
Also using ~/.local/share does not work
```
mkdir -p "$DATADIR/bash-completion"
completion_file="${ZDOTDIR:-$DATADIR}/bash-completion/completions"
```

ivan-hc:
I'm testing with both file names bash_completion and completions (as specified ad https://github.com/scop/bash-completion )

ivan-hc:
@prodigalshadow it seems that the right location is ~/.local/share, not ~/.config

and I also was wrong, "completions" is a directory, not a file

I just tested this in both ~/.config and ~/.local/share, this is the working one
```
mkdir -p "$DATADIR/bash-completion/completions"
completion_file="${ZDOTDIR:-$DATADIR}/bash-completion/completions/am"
```

the file may be named "am", for both AM and AppMan

ivan-hc:
I moved the change to the "dev" branch, its enough to run `am --devmode-enable && am -s` to test it, same for "AppMan".

I need to test this in ZSH

ivan-hc:
I've done a test right now with both ~/.bash_completion and the new file, for the first one i changed "list" with a random text file and then closed/reopened the terminal: it has priority over the new file.

Since it still redirects to "list" its OK, but what if someone has still a previous version of AppMan?

ivan-hc:
[...]

PR #905, opened 2024-08-30
Move completions file to ~/.local/share/bash-completion/completions

ivan-hc:
Fix https://github.com/ivan-hc/AM/issues/903

Change completion file name and path, from `~/.bash_completion` to `~/.local/share/bash-completion/completions/am`

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -303,7 +303,8 @@ function _completion_lists() {
 }
 
 # BASH AND ZSH COMPLETION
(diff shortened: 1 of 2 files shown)


Samueru-sama:
every time I run appman I get this error: 

```
/home/samuel/Local/bin/appman: line 310: /home/samuel/Local/config/zsh/bash-completion/completions/am: No such file or directory
Shell completion has been enabled!
```

And my zshrc gets patched every time as well.

PR #906, opened 2024-08-31
Added install script for pcsx-redux

marco-calautti:
This pull request attempts to add an install script for the pcsx-redux emulator.

It looked simple enough that I decided to give it a shot :).

This pull request is associated with the pull request here: https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/pull/17

Let me know in case I missed something.

~~P.S. I tried testing to uninstall the app, and appman says that the app has not been removed. I am not sure whether this has to do with the fact that I used my script not included in the db, or whether there is an error in how I prepared the script. However, running the "remove" script manually works as expected.~~
Nevermind, I was not indicating Y as an answer to the request of deleting the app. However, usually when the answer in parentheses is specified in uppercase, like Y, it means it will be the default. It does not seem to be the case here.

### DIFF
--- a/programs/x86_64-apps
+++ b/programs/x86_64-apps
@@ -1415,6 +1415,7 @@
 ◆ pb-for-desktop : Pushbullet Desktop app.
 ◆ pboy : A small .pdf management tool with a command-line UI.
 ◆ pcloud : Europe's most secure cloud storage trusted by 20 million people.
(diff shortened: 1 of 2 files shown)


ivan-hc:
I did a little "redux" (reduction) on your script, to save space and quickly review the affected line in case of problems

Issue #922, opened 2024-09-04
Change the way icons are handled to automatically follow icon theme.


LostWarriorr:
in the .desktop file of applications downloaded the Icon path points directly to the downloaded Icon as follows:
`Icon=/home/$USER/Applications/libreoffice/icons/libreoffice-base`

However if  the icons downloaded were instead put into `/home/$USER/.local/share/icons` we would be able to point to the icon directly:
`Icon=libreoffice-base`

Then the system will automatically theme it according to the selected icon theme if applicable. If it is unable to find a theme it will automatically default to the one downloaded by AppMan that are placed into `/home/$USER/.local/share/icons` 

However in XFCE I believe a `/home/$USER/.xsessionrc` file with the following contents is needed for this to work:

`case "$PATH" in`
`   *$HOME/.local/share/icons/) ;;`
`   *) PATH=$HOME/.local/share/icons/:"$PATH"`
`esac`
`export PATH`

I'm not sure what other DEs need that file but I know KDE doesn't.

ivan-hc:
what if I write an option to remove the path from the launcher instead?

ivan-hc:
for example
```
appman icon-theme $APP
```
to remove the path for one or more apps, or
```
appman icon-theme --all
```
to remove the path for all installed apps?

ivan-hc:
if you want to test it for a preview, enable support to developer mode and update
```
appman --devmode-enable
appman -s
```
to exit developer mode, use
```
appman --devmode-disable
```

@Samueru-sama @LostWarriorr  this is what I changed https://github.com/ivan-hc/AM/commit/50c52e6b735b0f963f176d0fb0458800103f1cde

if it is OK, I'll add also an help message to the option -h and on the README.

Of course, `icon-theme` is a temporary name, I can change it if you want.

ivan-hc:
I have tested this with
```
appman -i lxtask anydesk chromium gimp
```
and "gimp" is the only one I have installed also as system package via APT.

I performed tests on one app first, then on two, and finally using the --all flag, and since GIMP is already installed system wide, the AppImage uses the same icon because listed among themes.

ivan-hc:
PS: run `appman -s` two times, I just fixed a bug that prevented me to download modules from the "dev" branch https://github.com/ivan-hc/AM/commit/e65659cb4f8370efed27184856c8c2197fcfd3df

LostWarriorr:
[...]

PR #924, opened 2024-09-04
"AM" 8.2: added an option to remove icon path and allow the use of custom icon themes

ivan-hc:
All it does is use "`sed`" to remove the "$APPSPATH"/"$arg"/icons path from the launcher, at the "Icon=" entry in the .desktop file.

It is intended to be used both for single/multiple applications (adding the name/names) and for all of them at once.

Syntax (temporary):
```
am icon-theme $APP
am icon-theme --all
```
or
```
appman icon-theme $APP
appman icon-theme --all
```

"AM" requires a password, while "AppMan" always works without special permissions (as always).

To reverse this option, you need to remove/reinstall the app, or add the path manually to the .desktop file.

Fix https://github.com/ivan-hc/AM/issues/922

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="8.1.1-2"
+AMVERSION="8.2-beta1"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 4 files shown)


ivan-hc:
Now option will be `icons` or `--icons`

rauldipeas:
This option is not working as expected on Debian stable.
image

ivan-hc:
@rauldipeas just opened an issue for this comment

PR #980, opened 2024-10-02
add puddletag appimage

Samueru-sama:
[puddletag.md](https://github.com/user-attachments/files/17220364/puddletag.md)

Btw Ivan Ryujinx got taken down, so the package likely has to be removed from AM (not the appimage though)


ivan-hc:
I have renamed the script for the appimage and now is the main one, also if unofficial... and also the page on the catalogue. Thanks.

PR #994, opened 2024-10-10
Update zen-browser to latest stable release

Samueru-sama:
The Twilight builds are new and very alpha (the stable releases themselves are alpha).


ivan-hc:
so we need to add also the other builds, as we do with brave

Issue #988, opened 2024-10-06
AppImageLauncher and "AM" (or CLI management of AppImages in general) cannot coexist!


ivan-hc:
More and more users are reporting problems like this:

- from https://github.com/ivan-hc/AM/issues/984#issuecomment-2395198607

![373908070-c72a976b-108a-414f-bf74-a951168a3f28 png](https://github.com/user-attachments/assets/47d1d405-4aeb-497a-972b-1ce751ee0b4b)

- https://github.com/ivan-hc/AM/issues/955
- issues where happened also here https://github.com/ivan-hc/AM/issues/293
- https://github.com/ivan-hc/AM/issues/107
- https://github.com/ivan-hc/Database-of-pkg2appimaged-packages/issues/6
- here also happened something related https://github.com/ivan-hc/GIMP-appimage/issues/26

Even on Reddit they contact me privately or via message to report the problem.

Anyway, it seems that installing "appimagelauncher" causes problems with AppImages running via the command line, both with "AM" and normally.

I don't know if this is a problem with the AUR installation, on Arch Linux... or with the AppImage of AppImageLauncher itself. I don't have a complete picture of why these problems happen.

But one thing is for sure:

### to fix the problem, simply remove "appimagelauncher", reboot and try again.

And magically the problem disappears!

This is what I always have to suggest to users, and it seems unfair and competitive for me to say it. I don't want it to be that way.

Help me find the exact problem, I need to add something in "AM" (and maybe also in my AppImages and Archimage) to prevent these errors from happening again.

I can't keep suggesting people to remove "appimagelauncher" and reboot. Those who don't know, will always blame "AM" for being a bad program!

Who can help me investigate?

ivan-hc:
Steps to perform:
- compare the behavior of AppImageLauncher if installed via "AM", via AUR and via AppImage download
- reboot
- install programs in AppImage format from "AM" or "AppMan", or even just launch them from the command line
- debug to find the exact problem (system daemons, libraries loaded in background...)
- report the problems encountered here, in this issue

It is the only way to analyze the problem and propose essential points to work on together, in order to solve this problem once and for all.

Thanks for the collaboration.

ivan-hc:
I can't count on a direct intervention from the AppImageLauncher developers, since they haven't updated the stable version since 2020 and the development version since 2022 (the latter is the one available in AUR)... and yet, their program is still the most suggested on the web: **The problem is now intrinsic**! It's too late to update and fix it quickly and in place!

[...]

PR #995, opened 2024-10-11
Add a blacklist for files that may cause issues to "AM"/"AppMan"

ivan-hc:
fix https://github.com/ivan-hc/AM/issues/988

The cataloging of programs, daemons, libraries and services that could compromise the functioning of "AM" begins.

It often happens that some AppImage package manager in the form of a service or system daemon, takes control of such packages, preventing other managers, such as "AM", from being able to manage them as well.

For example, it could prevent the extraction of the package (to obtain the launcher and the icon), the cataloging of installed apps based on the format, even the updating of them. All this can appear in the form of error messages like this:
```
❯ appman install evince baobab-gtk3
============================================================================

                  START OF ALL INSTALLATION PROCESSES

============================================================================

 ◆ "EVINCE": starting installation script

Document-Viewer_46.3.1-2- 100%[===================================>]  55.69M  6.23MB/s    in 9.1s    
This doesn't look like a squashfs image.
Failed to open squashfs image
This doesn't look like a squashfs image.
Failed to open squashfs image
sed: can't read ./evince.desktop: No such file or directory
mv: cannot stat './evince.desktop': No such file or directory

 "EVINCE" INSTALLED (56 MB OF DISK SPACE)
____________________________________________________________________________

 ◆ "BAOBAB-GTK3": starting installation script

Disk_Usage_Analyzer-3.38. 100%[===================================>]   3.74M  3.83MB/s    in 1.0s    
This doesn't look like a squashfs image.
Failed to open squashfs image
This doesn't look like a squashfs image.
Failed to open squashfs image
sed: can't read ./baobab-gtk3.desktop: No such file or directory
mv: cannot stat './baobab-gtk3.desktop': No such file or directory

 "BAOBAB-GTK3" INSTALLED (4 MB OF DISK SPACE)
____________________________________________________________________________
============================================================================
```
and the user who doesn't know where the problem is, will blame the new AppImage manager for this failure.

We don't want this to happen again!

------------------------------

# 1. appimaged
The first blacklisted file is the program and system daemon "**appimaged**", guilty of recognizing all AppImage files in the system to integrate them, causing the addition of more entries in the application menu (sometimes broken and not well patched), and managing updates differently from the one supported by "AM", compromising its scheme.

From now on, its presence will be reported, preventing the use of "AM" and explaining the reason via message.

Istantanea_2024-10-11_02-13-03 png

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -221,6 +221,22 @@ function _online_check() {
 	fi
 }
 
+# BLACKLIST FILES
+function _blacklisted_file() {
+	echo ""
+	echo -e " ${RED}💀WARNING! Detected \"$blacklisted_file\"\033[0m"
+	echo ""
+	echo "It will prevent \"$AMCLIUPPER\" from working correctly with AppImages, especially when extracting, integrating and updating them." | fold -sw 77 | sed 's/^/ /g'
+	echo ""
+	echo " Please remove \"$blacklisted_file\", reboot and retry!"
+	echo ""
+	exit 0
+}
+
+if command -v appimaged &>/dev/null; then
+	blacklisted_file=$(command -v appimaged) _blacklisted_file
+fi
+
 ################################################################################
 #				3RD PARTY
 ################################################################################
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -222,19 +222,31 @@ function _online_check() {
 }
 
 # BLACKLIST FILES
+appimagelauncher_msg="Your installation of AppImageLauncher may have been done via DEB, RPM, or AUR, which interrupts the \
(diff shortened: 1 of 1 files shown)


ivan-hc:
## 2. appimagelauncherd

The reference is the system daemon "appimagelauncherd", common in all known packages (DEB, RPM and PKGBUILD) and a sign of a system-wide installation of AppImageLauncher. The damage caused by such an installation is bad for all alternative AppImage package helpers, not just for "AM"/"AppMan".

It is still suggested to use AppImageLauncher as a standalone AppImage package, also available in the database of available applications.

It is the only way to make AppImageLaucher coexist with other AppImage package managers without them conflicting, and without necessarily having to use root privileges to solve compatibility problems.

ivan-hc:
Istantanea_2024-10-11_04-27-56 png

PR #1006, opened 2024-10-16
Option "`-b`" or "`backup`", give the name you want to the snapshots

ivan-hc:
Now it is possible to give the name you want for the snapshots:
- press enter to to give date and time as aname, like always
- press "1" to use the version of the app
- write anything you want (eccept "1", its busy, see above), it will be used as a snapshot

Istantanea_2024-10-16_15-40-29 png

The option shows also the exact directory name.

Istantanea_2024-10-16_15-44-55 png

Now to restore backups using `-o` is more user friendly!

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="8.3.2"
+AMVERSION="8.4"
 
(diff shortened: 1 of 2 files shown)


Samueru-sama:
lgtm

PR #1039, opened 2024-10-27
Allow user to install both AM and AppMan at once.

Twig6943:


ivan-hc:
Hi, thanks. This isn't necessary, since I'm working on a way to handle apps installed locally from AM without switching to AppMan Mode.

Anyway, I'll keep this into a wait list.

PS: a suggestion, to ease installation of both, it would be enough to add a option 3 instead of writing both 1 and 2. Also mathematics says that: `1+2=3`

Twig6943:
Edited the AM-INSTALLER :+1:

ivan-hc:
I close this PR because, as you already know, I'm working on a way to allow "AM" to manage applications also in the way AppMan does:

- AppMan users are not very inclined to act with root privileges;
- "AM" users on the contrary would like to be able to manage more things at the same time, having full control over them, with or without root privileges.

I hope to provide you with the best version of "AM" possible.

Issue #1055, opened 2024-10-30
Optimize Portable Linux Apps Website


hotlittlewhitedog:
Hi all 😄 👍 

I would suggest you to optimize the load of Portable Linux Apps website, because it's too heavy (on mobile data...), not user friendly. I don't remember the exact max allowed size of a page but it should be less than 1MB. 

Check following screenshot:   

<img width="1521" alt="Screenshot 2024-10-30 at 18 36 28" src="https://github.com/user-attachments/assets/698b01be-88f0-48e4-a593-6bfb30ae0e45">

- Total download of 24 MB
- Load of 1.71 min 
- 2100 pictures of 128x128 not useful

Some solutions:
- Personally, I would remove the pictures of the main page,  
let's say: 2100 apps x ~7000 bytes (there are some pictures > 11000 bytes) => 14 MB. (24 - 14 => 10MB)
- I would also suggest you to create subcategories for AppImage section.
- Put the logo in the app description page 
- It's maybe also possible to compress the download page

Could be a good start and doesn't ask too much effort.

ivan-hc:
I need an expert of GitHub Pages for this.

I don't know how to handle them, I'm a total newbie here.

If you know how to configure the site in a way that the pictures will not be loaded on mobile view, let me know. Maybe something to add here https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/blob/main/_config.yml or here https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/tree/main/assets/css ...but again, I'm not an expert.

Please open an issue or a PR at https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io next time, it would be helpful for those that want to help me mprove the site.

Thanks for reporting it.

I'll close but pin this in the meantime.

hotlittlewhitedog:
OK 👍 I installed Ruby & compiled the site but it's not the last version of Ruby (on AntiX), got some errors and the server doesn't start. I install another machine and will try again... 😄

ivan-hc:
thank you very much

ivan-hc:
I move this to the dedicated repo, as I said, I think we need help on the JS/CSS/HTML level rather than a SHELL intervention.

ivan-hc:
OK, I don't know how to transfer this issue to https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io

@hotlittlewhitedog it is better that you re-open the issue there, there is someone that can help much more in this.

PR #1066, opened 2024-10-31
commit-am2plasite-remove-icons-from-table

hotlittlewhitedog:
modified:   tools/am2pla-site

Hi all 😄👍  

In this commit, I removed completely the icon field from the tables.   
I think it's the way to go because we can't just hide the icons in client side!  
It will be more efficient to do like this.  

Next step would be to create sub files from A to Z for AppImages category but first we have to check if it's quick enough.   

Here is a screenshot of current commit: 
  
Screenshot 2024-10-31 at 22 42 42

### DIFF
--- a/tools/am2pla-site
+++ b/tools/am2pla-site
@@ -83,8 +83,8 @@ function _table_head() {
 	echo ""
 	echo "-----------------"
 	echo ""
-	echo "| ICON | PACKAGE NAME | DESCRIPTION | INSTALLER |"
-	echo "| --- | --- | --- | --- |"
+	echo "| PACKAGE NAME | DESCRIPTION | INSTALLER |"
+	echo "| --- | --- | --- |"
 }
 
 function _categories_buttons() {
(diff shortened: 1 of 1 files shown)


hotlittlewhitedog:
Ref: https://github.com/ivan-hc/AM/issues/1055

ivan-hc:
hi, I already know that that is how the site appears without icons... it was like this before I added the icons.

So nope, to remove the icons like this is not the way to go... alternativelly can be created more pages for a mobile view... or a JS to hide the icons.

I added the icons because someone asked me, time ago, I can't step back. Sorry.

ivan-hc:
I'd better reopen the issue and move it to the dedicated repository.

I think we need help on the JS/CSS/HTML level rather than a SHELL intervention.

hotlittlewhitedog:
This kind of thing as to be done server side because on client side, it's too late. If you don't want to remove the icons, the only way is to divide the big categories for example in sections A-Z (by name) and the section by pages of maybe 20 items. But again, the subdivisions have to be done server side. // On client side, you have poor controls, it's just esthetic.

ivan-hc:
hi @hotlittlewhitedog , this is how I solved https://github.com/ivan-hc/AM/commit/183f0e31193cfe09391e85b8ecaf919dfb047398

PR #1084, opened 2024-11-06
Fixed rbw-agent

00sapo:
Not tested (I have just created the symlink on my machine where I had installed it with `appman`)

### DIFF
--- a/programs/x86_64/rbw
+++ b/programs/x86_64/rbw
@@ -3,6 +3,7 @@
 # AM INSTALL SCRIPT VERSION 3.5
 set -u
 APP=rbw
+APPAGENT=rbw-agent
 SITE="doy/rbw"
 
 # CREATE DIRECTORIES AND ADD REMOVER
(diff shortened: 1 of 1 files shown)


Samueru-sama:
line 12 has to be changed to add the extra symlink you added to the remover script, otherwise a broken link will be left in `/usr/local/bin` when the app is uninstalled. 

normally that extra file is a .desktop but in this case it would be that symlink.

programs/x86_64/rbw
```diff
@@ -45,6 +49,7 @@ if [ "$version" != "$version0" ]; then
 	cd ..
 	if [ -d ./tmp/* 2>/dev/null ]; then mv --backup=t ./tmp/*/* ./; else mv --backup=t ./tmp/* ./"$APP" 2>/dev/null || mv --backup=t ./tmp/* ./; fi
 	chmod a+x ./"$APP" || exit 1
+	chmod a+x ./"$APPAGENT" || exit 1
```
Samueru-sama, line 52: this isn't needed and will likely cause the scirpt to fail since iirc AM won't ask you for elevated rights when updating.

PR #1088, opened 2024-11-07
Update reaper

rauldipeas:
I've updated the `.desktop` with common icon name for most themes and include the window class, so app can be pinned in GNOME panel.

Besides cleaned an empty action that's causing an _Unamed_ entry on launcher quicklist.


ivan-hc:
I'm just working on the other thing... it would be the second release in about one hour

PR #1113, opened 2024-11-13
Show total number of programs available in `-l`...

ivan-hc:
...and keep all the lists locally so you can consult them offline, in case of no internet connection.

Istantanea_2024-11-13_04-33-40 png

@Samueru-sama @xplshn @Azathothas

The Toolpacks and AppImage lists will only be updated if called with the respective flag in `-l` or if synchronization is performed, which will be notified during the execution of `-s` or at the end of `-u`

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -722,7 +722,7 @@ function _sync_toolpacks_list() {
 }
 
 function _sync_databases() {
-	echo -e "$DIVIDING_LINE\n Sync databases..."
+	echo -e "$DIVIDING_LINE\n Check and update offline lists of additional databases..."
 	_sync_appimages_list
 	_sync_toolpacks_list
 	_completion_lists
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="9.1-4"
+AMVERSION="9.1-5"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
@@ -380,7 +380,7 @@ function _completion_lists() {
 	# Remove existing lists and download new ones
 	curl -Ls "$APPSLISTDB" > "$AMDATADIR/$arch-apps"
 	awk -v FS="(◆ | : )" '{print $2}' <"$AMDATADIR"/"$arch"-apps > "$AMDATADIR"/list
-	curl -Ls "$toolpack_readme" | grep -v "\.appimage " 2>/dev/null | cut -d' ' -f2- | awk '{print $1}' | uniq >> "$AMDATADIR"/list
+	curl -Ls "$toolpack_readme" | grep -v "\.appimage " 2>/dev/null | cut -d' ' -f2- | awk '{print $1}' | uniq | sed -e 's/$/.toolpack/' >> "$AMDATADIR"/list
 	# Append options to the list
 	for o in $available_options; do
 		echo "$o" >> "$AMDATADIR"/list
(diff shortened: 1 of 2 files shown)


ivan-hc:
Only, as you can see from the video, regenerating the list takes a long time

https://github.com/user-attachments/assets/a8700df6-34ae-4224-906c-103ef60191d6

...and this happens because the responsible process is this one, in APP-MANAGER
```
# 3RD PARTY SOURCES -- TOOLPACK
toolpack_repo="https://github.com/Azathothas/Toolpacks"
toolpack_readme="https://bin.pkgforge.dev/${ARCH}/AM.txt"
export toolpack_name="1" toolpack_description="2" toolpack_site="3" toolpack_dl="4" toolpack_ver="5"

...

function _sync_toolpacks_list() {
	rm -f "$AMDATADIR/$ARCH-toolpack"
	toolpack_md=$(curl -Ls "$toolpack_readme" | grep -v "\.appimage ")
	TOOLPACK_NAMES=$(echo "$toolpack_md" | cut -d' ' -f2- | awk '{print $1}')
	for toolpack in $TOOLPACK_NAMES; do
		toolpack_arg=$(echo "$toolpack_md" | grep -i "^| $toolpack " | tr '|' '\n' | cut -c 2- | grep .) \
		&& toolpack_about_description=$(echo "$toolpack_arg" | awk -F: "NR==$toolpack_description" | sed 's/ $/./g') \
		&& echo "◆ $toolpack : $toolpack_about_description" >> "$AMDATADIR/$ARCH-toolpack" &
	done
	wait
	sort "$AMDATADIR/$ARCH-toolpack" | uniq > "$AMDATADIR/$ARCH-toolpack-new" && mv "$AMDATADIR/$ARCH-toolpack-new" "$AMDATADIR/$ARCH-toolpack" || exit 1
}
```
...and I'd like to find a way to make it faster.

ivan-hc:
The purpose of this pull request is as follows

https://github.com/user-attachments/assets/bb3baacc-9a22-47f3-9391-60746ca738d3

xplshn:
I guess you could get ahold of `jq` and all of the deps that `AM` currently has, this includes, `bash`, etc, during the first run of `AM`, via using the Toolpacks repo or a binary manager like `dbin`.

Parsing Json is much faster than parsing a Markdown file manually via regexp.
There are various alternatives to JQ which support various file formats, Miller is one of them, there's also YQ, XQ, and a lot more.

There exists a JSON parser written in PURE POSIX shell too. So you may give that a try too.

The slowest thing here is the network call and the parsing.

ivan-hc:
> I guess you could get ahold of `jq` and all of the deps that `AM` currently has, this includes, `bash`, etc, during the first run of `AM`, via using the Toolpacks repo or a binary manager like `dbin`.
> 
> Parsing Json is much faster than parsing a Markdown file manually via regexp. There are various alternatives to JQ which support various file formats, Miller is one of them, there's also YQ, XQ, and a lot more.
> 
> There exists a JSON parser written in PURE POSIX shell too. So you may give that a try too.
> 
> The slowest thing here is the network call and the parsing.

Just found that `awk` does miracles https://github.com/ivan-hc/AM/pull/1113/commits/960354e7cb44d47a58730078cdaab2406474600b

https://github.com/user-attachments/assets/f07267d4-62c1-46b1-9c40-d6a9313cac5c

SOLVED!

ivan-hc:
PS: @xplshn have you added new programs? I see that the list is increased of about 500 programs.

ivan-hc:
Now creating Toolpacks list is faster than the AppImages list

Istantanea_2024-11-13_17-55-47 png

xplshn:
I added to `dbin`'s metadata the AppBundles generated/built by AppBundleHUB, the rest is Ajam's doing

PR #1131, opened 2024-11-16
Update genymotion

Twig6943:
Fixed the script

### DIFF
--- a/programs/x86_64/genymotion
+++ b/programs/x86_64/genymotion
@@ -7,24 +7,53 @@ SITE="https://www.genymotion.com"
 
 # CREATE DIRECTORIES AND ADD REMOVER
 [ -n "$APP" ] && mkdir -p "/opt/$APP/tmp" "/opt/$APP/icons" && cd "/opt/$APP/tmp" || exit 1
-printf "#!/bin/sh\nset -e\nrm -f /usr/local/bin/$APP\nrm -R -f /opt/$APP" > ../remove
+printf "#!/bin/sh\nset -e\n/opt/$APP/genymotion.bin -u -d /opt/$APP -y\nrm -f /usr/local/bin/$APP\nrm -R -f /opt/$APP" > ../remove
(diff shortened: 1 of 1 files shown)


Twig6943:
The app creates a .desktop entry itself

ivan-hc:
What the command `/opt/$APP/genymotion.bin -u -d /opt/$APP -y` does in the "remove" script?

ivan-hc:
also, if you already do `cd "/opt/$APP/tmp"` at line 9, why you do `wget "$version" -O /tmp/genymotion.bin` at line 19?  Also, /tmp is a system directory, when/opt/$APP/tmp is a working directory we create for safety pourpouses.

ivan-hc:
that said, `mv /tmp/genymotion.bin "/opt/$APP/genymotion.bin"` as no sense, you move the same file from a directory to another

ivan-hc:
```
# Execute the downloaded .bin file with sudo
if [ -x "/opt/$APP/genymotion.bin" ]; then
    sudo /opt/$APP/genymotion.bin -d /opt/$APP -y
else
    echo "Error: The downloaded file is not executable!"
    exit 1
fi
```
this means that it is a complete installer? Maybe?

If so, I must remove it from the database.

ivan-hc:
Have you tested this in both AM and AppMan, before submitting? I'm a bit scared, to be honest.

ivan-hc:
> I have, it kinda works isn't perfect tho https://www.youtube.com/watch?v=QdZZ7gSP1jE

headache :rofl: 

well, it should be installed with root permissions... it's not a program that AppMan can handle.

Another program that works with root privileges on both sides is Ventoy, but at least the install/remove always goes well.

That said, **I don't think it's a program worth adding to the database**. **It would be better if we removed it**. Installers who do what they want are not welcome here.

Sorry.

PS: @Twig6943  congratulations on your progress. Keep it up.

PR #1138, opened 2024-11-17
add `/run/media`

candrapersada:


ivan-hc:
@candrapersada what you are adding is in no way related to https://github.com/ivan-hc/MPV-appimage/issues/15

- AM is a package manager that distributes AppImages, if the AppImages are not working is not an issue related to AM, but to the app that has been packeged wrong
- Archimage (at https://github.com/ivan-hc/ArchImage) is a way to create AppImage packages by exportin a portable Arch Linux container ino an AppImage... I have already added these changes from https://github.com/ivan-hc/MPV-appimage/issues/15 to https://github.com/ivan-hc/ArchImage/commit/caa46a27b2eaf68b2cdeb4c163f6a0120982a2cf

What you are changing in this PR is a piece of documentation related to a total different thing.

Samueru-sama:
What the PR changes is actually kinda of correct though, `/run/media` is the default mount point of removable drivers set by gvfs iirc. 

However the wording should just mention any current not mounted location instead, because the user can mount another drive anywhere really.

ivan-hc:
what changes is a piece of documentation related to the option `--launcher`, and in particular to the use of the `-c` cleanup option that detects if the AppImage available on an external drive is in a mounted drive. If it is not, icons and launchers will not be touched, while if it is, they will be removed but only if the AppImage is no longer present in that path indicated in the launcher.

So no, it has absolutely nothing to do with Bubblewrap, Archimage and the functioning of containers.

PR #1185, opened 2024-12-03
Fix total size, by @Azathothas

ivan-hc:


modules/database.am
```diff
@@ -292,10 +292,9 @@ _files_appimage_type_notes() {
 }
 
 _files_total_size() {
-	_determine_args
 	printf "\n"
 	command -v aisap >/dev/null 2>&1 && printf '%s\n\n' " AppImages with 🔒 are sandboxed with aisap"
-	TOTAL_SIZE=$(echo "$ARGPATHS" | grep "^$APPSPATH" | du -ch 2>/dev/null | grep "total" | awk 'END {print $1"iB"}' | sed 's/...$/ &/')
+	TOTAL_SIZE=$(du -shc $(find "$APPSPATH" -maxdepth 2 -type f -name 'remove' -printf "%h\n" 2>/dev/null) | tail -n1 | awk 'END {print $1"iB"}' | sed 's/...$/ &/')
```
Samueru-sama, line 297: all you people ended up doing was add a `2>/dev/null` to the original find lol.

@Azathothas was telling that this was unsafe (which I don't see how) but put it back anyway?

Samueru-sama, line 297: Also I don't think the `tail -n1` is needed since the `END` in awk does the same.

PR #1186, opened 2024-12-03
`-ias` take two

Samueru-sama:
I think this is much simpler than what you are currently trying to do Ivan: 

Test: 

image

Of course sandboxes still has to be modified to not allow to install aisap when `--user` is passed.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1078,6 +1078,7 @@ case "$1" in
 		fi
 		if echo "${@}" | grep -q -- '--user'; then
 			user_true=1
+			unset SUDOCMD
 		fi
 		if [ "$CLI" = am ] && [ -f "$APPMANCONFIG"/appman-mode ]; then
 			echo -e "$APPMAN_MSG"
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1072,6 +1072,10 @@ case "$1" in
 		;;
 	'-ias')
 		shift
+		if ! command -v aisap >/dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


Samueru-sama:
>Of course sandboxes still has to be modified to not allow to install aisap when --user is passed.

Or better yet, don't allow `-ias` [without aisap](https://github.com/ivan-hc/AM/pull/1186/commits/083eb276970e39d5eb880c57bda6ece1a4c66a1a) first.

ivan-hc:
I want to avoid adding a separate entry... this seems more like an attempt to circumvent what I'm trying to solve.

I've been working on it for 3 days, and for some absurd reason those errors that I showed you in private occur.

I want to understand why the first argument is repeated 2 times, and above all what the alphabetical order has to do with it.

If you want to help me, let me understand why the behaviors that I showed you in all those screenshots in private. Don't add more useless code. We already have a base.

ivan-hc:
Also, this PR is towards "main".

Samueru-sama:
>If you want to help me, let me understand why the behaviors that I showed you in all those screenshots in private. Don't add more useless code. We already have a base.

What you are trying to do can't be done without doing more changes to the sandboxes module. 

The modules are meant to do what they were made to do, adding something like having the install module call the sandboxes module will just add a lot of unnecessary complexity, when it took me 10 minutes to get this done by doing the logic here as you can see. 

> Also, this PR is towards "main".

You already have dev busy

ivan-hc:
@Samueru-sama discord

PR #1190, opened 2024-12-06
improve func ` _check_appimage`

Samueru-sama:
This way I'm not finding the apps directory by checking `ARGSPATH` for something that matches the name of the binary in PATH.

Shouldn't give issues with appman mode, but let me know.


ivan-hc:
I've done my usual test with unprivileged/privileged/unprivileged app in the option `--sandbox`, no issue. Same for new `-ias` option. It looks good for me

PR #1201, opened 2024-12-10
use posix `LOGNAME` and shell built in test

Samueru-sama:
This way the script depends less on external binaries.

`$LOGNAME` is posix and has to be set in any posix compliant system (which includes freebsd) I also added a fallback to `$USER just in case LOGNAME gets manually unset lol`

### DIFF
--- a/INSTALL
+++ b/INSTALL
@@ -10,7 +10,6 @@ _away_all() {
 
 # DETERMINE SYSTEM ARCHITECTURE AND CURRENT USER
 arch=$(uname -m)
-currentuser=$(who | awk '{print $1}')
 
 _check_dependency() {
     program="$1"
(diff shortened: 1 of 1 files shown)


ivan-hc:
@Samueru-sama if you run the INSTALL script with `sudo`, both LOGNAME and USER are `root`

I have tested this for 2 hours. I already have another solution
```
[ -n "$SUDO_USER" ] && chown -R "$SUDO_USER" /opt/am 2> /dev/null \
|| [ -n "$DOAS_USER" ] && chown -R "$DOAS_USER" /opt/am 2> /dev/null
```
no $currentuser anymore

Samueru-sama:
> @Samueru-sama if you run the INSTALL script with `sudo`, both LOGNAME and USER are `root`
> 
> I have tested this for 2 hours. I already have another solution
> 
> ```
> [ -n "$SUDO_USER" ] && chown -R "$SUDO_USER" /opt/am 2> /dev/null \
> || [ -n "$DOAS_USER" ] && chown -R "$DOAS_USER" /opt/am 2> /dev/null
> ```
> 
> no $currentuser anymore

In that case all that had to be changed was `"${LOGNAME:-$USER}"` for `"${SUDO_USER:-$DOAS_USER}"` and another var can be added just in case both are unset.

ivan-hc:
ths way `chown` is added only if one of the two values exists, else only `root` can own /opt/am

PR #1206, opened 2024-12-11
replace `grep -P` for posix alternative

Samueru-sama:
test carefully. 

Haven't found issues: 

image

image

@xplshn @leo-arch If you have a better idea for the `_find_no_order` func let me know.

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -505,8 +505,14 @@ _query_lists() {
 }
 
 _find_no_order() {
-	grep -i "$1" | grep -i "$2" | grep -i "$3" | grep -i "$4" | grep -i "$5" \
-		| grep -i "$6" | grep -i "$6" | grep -i "$7" | grep -i "$8" | grep -i "$9"
+	while read -r line; do
+		match=1
+		for term in "$@"; do
+			[ -n "$term" ] || continue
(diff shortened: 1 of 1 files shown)


Samueru-sama:
@xplshn Had a better idea that works, I get the same word count as before and all of that.

ivan-hc:
by mistake I used the flag `--appimage` instead of the real one, `--appimages` (with filan "s"), this is what's happened

Istantanea_2024-12-11_13-32-28

I had to CTRL+C to close this

Samueru-sama:
fixed, it was the grep interpreting the flag. 

This shows that `grep -q` is not always the same as `grep 2>/dev/null` lol

xplshn:
> fixed, it was the grep interpreting the flag.
> 
> This shows that `grep -q` is not always the same as `grep 2>/dev/null` lol

You can use `--` to tell programs to stop interpreting anything that comes after `--` as a flag.

I'm not sure if this behavior is truly POSIX, but it is respected in Busybox, GNU coreutils and Toybox, I've never had to use `--` in BSD-land before so I'm not sure about it

Samueru-sama:
Yeah we have had to use `--` before, we do that already in other places with AM.

Issue #1199, opened 2024-12-10
Issues on FreeBSD


leo-arch:
A few issues when attempting to install and use AM on FreeBSD:

1. When installing `am` (git method), **wget**(1) isn't found (and thereby nothing was actually installed), but I still get this message: `"AM" successfully installed`.

2. I installed **wget**(1). After `am` installation (git method), I get this message:

```sh
Installing "AM" on /opt/am
WARNING: Couldn't link am to "/usr/local/bin/am!"
```
The thing is that `/usr/local/bin/am` was actually generated and properly points to `/opt/am/APP-MANAGER`.

3. I tried to install **clifm** using the `am -i clifm` command. Here's the output:

```sh
? WARNING: ACCESS TO USER NAMESPACES IS RESTRICTED!

Some apps may to run, you need to enable access to user namespaces,
please visit https://github.com/ivan-hc/AM#ubuntu-mess to know more

-----------------------------------------------------------------------------------------------------

START OF INSTALLATION PROCESS

==========================================================

? ERROR: "clifm" does NOT exist in the "AM" database, please check the list,
run the "am -l" command

==========================================================

END OF ALL INSTALLATION PROCESS

```

https://github.com/ivan-hc/AM#ubuntu-mess redirects to the main AM github page: no `ubuntu-mess` section.

The namespaces messages is triggered by APP-MANAGER:258: There's no **unshare**(1) command in FreeBSD. I've modified APP-MANAGER to only run **unshare**(1) if it exists:

```sh
if type unshare >/dev/null 2>&1 && ! unset --user -p /bin/true >/dev/null 2>&1
```
Still get the above error screen, but without the namespaces warning

Ok. Let's see what `am -l` has for us:

```sh
grep: /home/user/.cache/am/version-args: No such file or directory

YOU HAVE INSTALLED -1 PROGRAM OUT OF 622 AVAILABLE, AND     0 LIBRARIES

LIST OF THE 1 PROGRAMS AVAILABLE IN THE "AM" DATABASE:

To list all installable programs, use the "--all" flag

404: Not found.
```
Oops!

Let's run `am -l --all`. I only see a bunch of numbers, like these:

```
>Public Access</a

d="M103.074 142.831C103.048 ...
```
And the thing goes on. It seems to be parsing an HTML file, and failing.

### Summarizing

1. What about namespaces?
[...]

PR #1207, opened 2024-12-12
remove `-d` flag from `strings`

Samueru-sama:
Related to #1199

Using `grep -m 1` reduced the time it takes for `am -f` to complete over what was done before with `strings -d`:

Test (2nd result being with the change): 

image

@xplshn if you know something that could speed this up even more, this feature of am has always been slow in first try without the cached files (+4 seconds).

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -250,6 +250,7 @@ _files_files() {
 		 		_files_type &
 			fi
 		done
+		wait
 	fi
 	rm -f "$AMCACHEDIR"/files-sizes "$AMCACHEDIR"/files-db
 	for arg in $INSTALLED_APPS; do
--- a/modules/database.am
+++ b/modules/database.am
@@ -241,13 +241,13 @@ _files_db() {
 _files_files() {
 	cd "$APPSPATH" || exit 1
 	INSTALLED_APPS=$(find . -name 'remove' -printf "%h\n" 2>/dev/null | du -sh -- * 2> /dev/null | sort -rh | sed 's@.*	@@')
-	if ! test -f "$AMCACHEDIR"/version-args; then
+	if [ ! -f "$AMCACHEDIR"/version-args ]; then
 		_check_version
 	fi
-	if ! test -f "$AMCACHEDIR"/files-type; then
+	if [ ! -f "$AMCACHEDIR"/files-type ]; then
 		for arg in $INSTALLED_APPS; do
-			if test -f ./"$arg"/remove 2>/dev/null; then
-		 		_files_type
+			if [ -f ./"$arg"/remove 2>/dev/null ]; then
+		 		_files_type &
 			fi
 		done
 	fi
--- a/modules/database.am
+++ b/modules/database.am
@@ -175,7 +175,7 @@ _files_if_script() {
 }
 
 _files_if_appimage() {
-	if ! echo "$string" | grep 'AppImages require FUSE to run' >/dev/null 2>&1; then
+	if file ./"$arg/$arg" | grep -qi "static"; then
 		if grep "SANDBOXDIR" "$FILE" >/dev/null 2>&1; then
 			echo " ◆ $arg	|	appimage🔒" >> "$AMCACHEDIR"/files-type
 		else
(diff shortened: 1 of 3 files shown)


ivan-hc:
also, you have 2 references in management.am and 1 in sandboxes.am

Samueru-sama:
With the last commit now it takes less than 2 seconds to complete lol

image

ivan-hc:
> With the last commit now it takes less than 2 seconds to complete lol

@Samueru-sama I don't think its a good idea of using that final `&`

Istantanea_2024-12-12_03-23-40

each program deserve its time for colletting info

Samueru-sama:
> each program deserve its time for colletting info

mmmmm maybe I need to put some `wait` somewhere to still have the 2 second time save, but in the time being I will remove that from here then.

I can't replicate that issue btw, please make sure to clear the cache and try several times.

ivan-hc:
@Samueru-sama put a `wait` in that function
```
_files_files() {
	cd "$APPSPATH" || exit 1
	INSTALLED_APPS=$(find . -name 'remove' -printf "%h\n" 2>/dev/null | du -sh -- * 2> /dev/null | sort -rh | sed 's@.*	@@')
	if [ ! -f "$AMCACHEDIR"/version-args ]; then
		_check_version
	fi
	if [ ! -f "$AMCACHEDIR"/files-type ]; then
		for arg in $INSTALLED_APPS; do
			if [ -f ./"$arg"/remove 2>/dev/null ]; then
		 		_files_type &
			fi
		done
		wait
	fi
	rm -f "$AMCACHEDIR"/files-sizes "$AMCACHEDIR"/files-db
	for arg in $INSTALLED_APPS; do
		if test -f ./"$arg"/remove 2>/dev/null; then
			_files_sizes
			_files_db
		fi
	done
}
```

ivan-hc:
> > each program deserve its time for colletting info
> 
> mmmmm maybe I need to put some `wait` somewhere to still have the 2 second time save, but in the time being I will remove that from here then.
> 
> I can't replicate that issue btw, please make sure to clear the cache and try several times.

you scare me, you read my mind

Samueru-sama:
@ivan-hc can you try this and see if it fixes the issue?

image

put a wait after the for loop where `_files_type` is called.

ivan-hc:
> put a wait after the for loop where `_files_type` is called.

better after the `done`, since that part is acting on single arguments if not listed

ivan-hc:
I have seen that in `--launcher ` you check for `github.com/AppImage/AppImageKit/wiki/FUSE`, what if the AppImage is Type Thr.... ehm, a next generation one?

Samueru-sama:
> I have seen that in `--launcher ` you check for `github.com/AppImage/AppImageKit/wiki/FUSE`, what if the AppImage is Type Thr.... ehm, a next generation one?

It should still work, `github.com/AppImage/AppImageKit/wiki/FUSE` is a common string on all appimage runtimes as far as I know. 

In fact the addition of `github.com/AppImage/AppImageKit/wiki/FUSE` was recent because it also recognizes the `uruntime` appimages while the older method did not.

--------------------

To determine if an AppImage is ~~type3~~ next-generation you just need to run `file` on it and `grep "static"`, which I also added since before we used strings for that as well.

xplshn:
> Related to #1199
> 
> Using `grep -m 1` reduced the time it takes for `am -f` to complete over what was done before with `strings -d`:
> 
> Test (2nd result being with the change):
> 
> image
> 
> @xplshn if you know something that could speed this up even more, this feature of am has always been slow in first try without the cached files (+4 seconds).

Sorry, today I swapped distro, I had to compile 68 packages and then open my notebook  and swap the HDD drive I had prepared.

image

xplshn:
`grep -Eao '[\t\r\n -~]{4,}'`, behaves the same way as `string`, works on busybox, and should work everywhere else too

xplshn:
image
You can do both the `strings` and `grep` with a single `grep` line, but I used two grep commands for clarity

xplshn:
in `pelf-sqfs`, an old edition of `pelf`, I also used a similar trick, but to calculate the offset until we find a certain marker of string, perhaps that's useful here too? For now, I think this is the best solution.

`grep` uses the [Boyer Moore String Search Algorithm](https://en.wikipedia.org/wiki/Boyer%E2%80%93Moore_string-search_algorithm), which gets increasingly faster as the file's size increases

PR #1208, opened 2024-12-12
replace `-printf` in find with posix alternative

Samueru-sama:
Related #1199 

If you know another find that's still having issues let me know, but you can usually guess what the problem is by looking at the flags being used, this is the first time I see `-printf` being used with find so it was easy to catch.


ivan-hc:
sure, there are some references in database.am, and I have just changed them

PR #1209, opened 2024-12-12
"AM" 9.4: support for FreeBSD

ivan-hc:
https://github.com/user-attachments/assets/da9f1e72-6671-434a-bf5b-507a777136c5

fix https://github.com/ivan-hc/AM/issues/1199

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="9.3-2"
+AMVERSION="9.3-3"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
@@ -17,8 +17,10 @@ DIR="$( cd "$( dirname "$0" )" && pwd )"
 CLI=$(basename "$0")
 
 # Determine system architecture and current user
-arch="$HOSTTYPE"
-export ARCH="$arch"
+ARCH="$(uname -m)"
+[ "$ARCH" = "amd64" ] && ARCH=x86_64
+arch="$ARCH"
+export ARCH arch
 
 # XDG Variables
 export BINDIR="${XDG_BIN_HOME:-$HOME/.local/bin}"
@@ -39,7 +41,10 @@ LightBlue='\033[1;34m'
 DIVIDING_LINE="-----------------------------------------------------------------------------"
 
 # Prevent the use of "sudo" ("AM")
-[ -n "$SUDO_COMMAND" ] && echo -e "\n Please do not use \"sudo\" to execute \"$CLI\", try again.\n" && exit 1
+if [ -n "${SUDO_USER:-$DOAS_USER}" ]; then
+	printf "\n Please do not use \"sudo\" to execute \"$CLI\", try again.\n\n"
+	exit 1
+fi
 
 function _create_cache_dir() {
 	AMCACHEDIR="$CACHEDIR/$AMCLI"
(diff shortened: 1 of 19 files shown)


APP-MANAGER
```diff
@@ -213,6 +222,11 @@ AMCLIUPPER=$(echo "$AMCLI" | tr '[:lower:]' '[:upper:]')
 # Create new data directory and move important files there
 AMDATADIR="$DATADIR/AM"
 mkdir -p "$AMDATADIR"
+if [ "$sed_cmd" = gsed ]; then
+	mkdir -p "$AMDATADIR/bin"
+	export PATH="${AMDATADIR}"/bin/:"${PATH}"
+	[ ! -f "${AMDATADIR}"/bin/sed ] && ln -s /usr/local/bin/gsed  "${AMDATADIR}"/bin/sed
```
Samueru-sama, line 228: `ln -s "$(command -v gsed)" "${AMDATADIR}"/bin/sed`

And likely with this you don't need to patch `gsed` into the install script. 

Also some check to make sure that gsed exists on the system.

ivan-hc: I need it, since apps are installed with root privileges and scripts runs out of AM itself. AM only executes them, and they rely on variables and commands available system wide.

I have shown a lot of screenshots and videos where installing an app returns a `sed` error after the download. This patch solves this.

I use AM as root, so commands available locally are only for local apps, not for system apps.

Samueru-sama: > I need it, since apps are installed with root privileges and scripts runs out of AM itself. AM only executes them, and they rely on variables and commands available system wide.
> 
> I have shown a lot of screenshots and videos where installing an app returns a `sed` error after the download. This patch solves this.
> 
> I use AM as root, so commands available locally are only for local apps, not for system apps.

That doesn't mean that you can't `ln -s "$(command -v gsed)" "${AMDATADIR}"/bin/sed` instead of the hardcoded `ln -s /usr/local/bin/gsed  "${AMDATADIR}"/bin/sed`

Samueru-sama: In other words `ln -s /usr/local/bin/gsed  "${AMDATADIR}"/bin/sed` will only work if you have gsed in `/usr/local/bin`, which I don't know how freebsd works but if you install the coreutils package `gsed` should be in `/bin` or `/usr/bin`

ivan-hc: sorry, I misunderstood. It's a language barrier issue, as you pointed out on Discord.

fixed in https://github.com/ivan-hc/AM/pull/1209/commits/67037cd0716d1c512521f2376cce6bf30ec2b98f

ivan-hc:
Next step is to add FISH completion, since AM only supports BASH and ZSH

ivan-hc:
Anyone that wants to test and contribute to this next release can enable the Developer Mode like this
```
am --devmode-enable
am -s
```
to exit the Developer Mode
```
am --devmode-disable
```

NOTE, once the version will be releases as 9.4, you can safely exit the Developer Mode.

This branch is for testing pourposes, not for daily use.

APP-MANAGER
```diff
@@ -545,10 +545,11 @@ _check_version_for_auto_updatable_apps() {
 		if [ -f "$argpath"/updater ]; then
 			_check_version_if_any_version_reference_is_somewhere
 			if [ "$AMCLI" = am ] && [ -d "$CACHEDIR"/am ] && grep -q " ◆ $arg	|" "$CACHEDIR"/am/version-args; then
-				OLDAPPVERSION=$(grep " ◆ $arg	|" "$CACHEDIR"/am/version-args | awk '{print $NF}')
+				OLDAPPVERSION=$(grep " ◆ $arg	|" "$CACHEDIR"/am/version-args | awk '{print $NF}' | tail -1)
```
Samueru-sama, line 548: `awk 'END {print $NF}'` and that way you avoid an extra pipe and` tail -1`

ivan-hc: I added `tail -1` because I had the issue on GhostBSD

ivan-hc: thank you https://github.com/ivan-hc/AM/pull/1209/commits/3f0da5a1ee8286fb97c802ec39301e893e681765

ivan-hc:
I  followed this guide to install `debootstrap` and Debian Stable on freeBSD, and not just portable programs, but also AppImages are running (quite) well n freeBSD/GhostBSD https://unixdigest.com/tutorials/how-to-install-signal-desktop-on-freebsd-using-the-linux-binary-compatibility.html

NOTE, chromium-based applications require to be executed with the `--no-sandbox` flag

Here my AppImage of Google Chrome

Istantanea_2024-12-16_18-37-37
Istantanea_2024-12-16_18-36-52

...here instead Brave AppImage...

Istantanea_2024-12-16_18-44-06

...dulcis in fundo, my LXtask AppImage

Istantanea_2024-12-16_18-38-22

Not everything works 100%, but it's a start.

PR #1210, opened 2024-12-12
Remove strings dependency

Samueru-sama:
Test carefully, works on my end and it works with busybox grep, but I'm not 100% sure with bsd grep. 

Also made the `_files_sizes` func run in the background and added the wait after it, this cut the time down to 1.4 seconds on my end.

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -297,7 +297,7 @@ _files_total_size() {
 	if command -v aisap >/dev/null 2>&1 && grep -qe "appimage🔒" "$AMCACHEDIR"/files*; then
 		printf '%s\n\n' " AppImages with 🔒 are sandboxed with aisap"
 	fi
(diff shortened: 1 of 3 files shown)


ivan-hc:
@Samueru-sama take a look at TOTAL SIZE in both the tables

Istantanea_2024-12-12_20-01-06

ivan-hc:
but this is a new bug introduced in yesterday's changes

ivan-hc:
the fix would be to replace `-maxdepth 2` with `-maxdepth 1`... but this must be tested also in case other directories not managed by AM are in place

ivan-hc:
I need to work on the way the apps are detected while creating the tables

ivan-hc:
> Test carefully, works on my end and it works with busybox grep, but I'm not 100% sure with bsd grep.
> 
> Also made the `_files_sizes` func run in the background and added the wait after it, this cut the time down to 1.4 seconds on my end.

PS: it works on GhostBSD

PR #1214, opened 2024-12-13
make sed compatible with bsd

Samueru-sama:
Related to #1199 

One quick google search of the sed error you had lead me to this: https://stackoverflow.com/questions/45712479/sed-does-not-work-as-expected-with-variables-in-script

And searching for `sed -i'' -e` on github shows thousands of results, so this should be all that's needed to at least make apps install with bsd sed, there might be some other sed errors in the main script or the modules though. 

I also tried to run ghostbsd to test but couldn't, gnome-boxes tells me no operating system found on the iso. 

------------------------------------------------------------

If you still have an error when installing an app, just tell me what the error was and I will see what I can do.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -223,8 +223,8 @@ AMDATADIR="$DATADIR/AM"
 mkdir -p "$AMDATADIR"
 if [ "$sed_cmd" = gsed ]; then
 	mkdir -p "$AMDATADIR/bin"
-	export PATH="${AMDATADIR}"/bin/:"${PATH}"
-	[ ! -f "${AMDATADIR}"/bin/sed ] && ln -s "$(command -v gsed)"  "${AMDATADIR}"/bin/sed
+#	export PATH="${AMDATADIR}"/bin/:"${PATH}"
+#	[ ! -f "${AMDATADIR}"/bin/sed ] && ln -s "$(command -v gsed)"  "${AMDATADIR}"/bin/sed
 fi
 
 # DEVELOPER MODE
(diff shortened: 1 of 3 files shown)


ivan-hc:
@Samueru-sama I have done a test in install.am, I used this command instead
```
sed -i -- "s/^sed -i/sed -ie/g" ./"$arg"
```
and the installation worked well

Samueru-sama:
Alright in that case all the bsd checks can go. 

about `sed -i'' -e` vs `sed -ie` if both methods work I will of course go with the shorter one, but I want to be 100% sure about that, in other words I want to hear what @xplshn thinks lol 

What I'm not sure is why the original suggestion had `-i''` with those single quotes.

xplshn:
> Alright in that case all the bsd checks can go.
> 
> about `sed -i'' -e` vs `sed -ie` if both methods work I will of course go with the shorter one, but I want to be 100% sure about that, in other words I want to hear what @xplshn thinks lol
> 
> What I'm not sure is why the original suggestion had `-i''` with those single quotes, my understanding is to tell `sed` that `-i` has no extension, which bsd `sed` expects.

`-i` is not the same across different sed implementations, for example, `sbase` which is 100% POSIX does not have it.

I think it is best to simply wrap around `sed`

Samueru-sama:
> I think it is best to simply wrap around `sed`

This seems to be a common issue that also affects macos and the "universal" solution is using `sed -i'' -e`, testing needs to be done.

Samueru-sama:
And btw about `sed -ie` vs `sed -i'' -e` it is not the same as seen and I tested [here](https://stackoverflow.com/questions/43582494/what-is-the-difference-between-sed-i-e-and-sed-ie).

So all that's left is make sure this does not cause issues with BSD or GNU sed, as well as test with busybox sed.

Samueru-sama:
@xplshn Go to discord because a bigger issue has been found.

PR #1227, opened 2024-12-15
drop lowercase `$arch`

Samueru-sama:
Seems to have only been used in the main script.


ivan-hc:
I didn't even realize we had already removed them all from the modules. Thanks

PR #1247, opened 2024-12-22
FDM

Twig6943:
.


ivan-hc:
Seriously, you can do better than this... none needs ChatGPT to write a script like this

ivan-hc:
It is enough to see the APP and SITE/URL reference in the installation script and the AM-updater provided with this script. Also, you had not the needing to specify an absolute path and a binary name for `chmod`, and also because you alread `cd` into a directory. Also, `cat`/`EOF`, the way you create the file.... all this sucks! Really you need to use ChatGPT to create a script like this?

If you were using `am -t` like all other app submitters, you had not these issues.

ivan-hc:
I can't even say I appreciate the effort, because the only effort was adding the line to the apps list.

PR #1255, opened 2024-12-25
Option -f, add "🔒" to versions locked with option "lock"

ivan-hc:
Applications that have version locked not to be updated will be shown with 🔒

In the example below I locked Virtualbox using the command `am lock virtualbox`

Istantanea_2024-12-25_22-20-49

Fun fact, I just updated [my VirtualBox KVM AppImage](https://github.com/ivan-hc/VirtualBox-appimage) yesterday, version 7.1.4 is based on QT6 and is way too slow for my taste... plus Kvantum 6 hasn't been ported to Debian Testing/Unstable yet, so I downgraded with `--rollback` and used the 7.0.20 release, which works better.


ivan-hc:
This is what the last commit does, if we have both sandboxed appimages and locked versions this will be the message

Istantanea_2024-12-26_01-56-15

...this is with only locked versions

Istantanea_2024-12-26_01-56-57

...and finally the classic one with only sandboxed appimages

Istantanea_2024-12-26_01-58-03

...not to waste space for the table.

PR #1279, opened 2025-01-13
Update README.md

daUnknownCoder:
move to /tmp to avoid clogging

### DIFF
--- a/README.md
+++ b/README.md
@@ -91,7 +91,7 @@ The "AM-INSTALLER" script allows you to choose "[AppMan](#what-is-appman)" inste
 
 Copy/paste the following one line command to download and run the "[AM-INSTALLER](https://github.com/ivan-hc/AM/blob/main/AM-INSTALLER)" script
 ```
-wget -q https://raw.githubusercontent.com/ivan-hc/AM/main/AM-INSTALLER && chmod a+x ./AM-INSTALLER && ./AM-INSTALLER
+wget https://raw.githubusercontent.com/ivan-hc/AM/main/AM-INSTALLER -O /tmp/AM-INSTALLER && chmod a+x /tmp/AM-INSTALLER && /tmp/AM-INSTALLER
 ```
 ...below, the screenshot of what will appear.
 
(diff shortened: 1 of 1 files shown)


zen0bit (approved):

ivan-hc:
I appreciate the idea, but the installation method is fine as is. I think it is up to the user to consider what they download and install. Just read the guide and see what a series of commands do. It is also a way to learn.

Aside from that, there is also a typo in your PR.

Regardless, the scripts to install AM are already extensively tested and proven, there is no need to change, even if it were downloaded twice by mistake, the installation would still be successful.

That said, I prefer not to change. It was better to tell you than to close without explanation.

Thanks for your interest and contribution.

daUnknownCoder:
yes thats fine, it may be beneficial for users like me who are in a random config directory and then and run this and get a file and yay AM is installed, now the program they run tell theres an error and we start debugging to find the script was here...

nvm, cachedir for install script is better anyway and yeah one thing i would suggest is `rm` the file after installation specially `AM_INSTALL` because if not `/tmp` removing it would be for the better

Thanks for your reply tho, btw i didnt understand what `typo` i had? is it in the code? or the PR itself?

ivan-hc:
The typo is that there is no AM_INSTALL file.

Btw if someone uses AM-INSTALLER to install AM, no risks  of duplicates at all... and if you read it, it uses the cache directory to install AM system wide.

The quick installation is as third choice or "other methods" because the first is the recommended one.

daUnknownCoder:
> The typo is that there is no AM_INSTALL file.

`wget https://raw.githubusercontent.com/ivan-hc/AM/main/INSTALL -O /tmp/AM_INSTALL && chmod a+x /tmp/AM_INSTALL && sudo /tmp/AM_INSTALL` -> im telling `wget` to download the file to `AM_INSTALL`

> Btw if someone uses AM-INSTALLER to install AM, no risks of duplicates at all... and if you read it, it uses the cache directory to install AM system wide.

not talking abt dupes, the first AM-INSTALLER _will_ remain in the `cwd` even if `INSTALL` is in `~/.cache`

> The quick installation is as third choice or "other methods" because the first is the recommended one.

Yep, no issues abt that, but if someone wants to use 3rd option, again specify download to `~/.cache/INSTALL` which is missing iirc

PR #1312, opened 2025-01-26
Success if all already tested

zen0bit:
If totest.list will be empty (All apps already tested)
[run](https://github.com/oSoWoSo/AM/actions/runs/12974850343)


ivan-hc:
@zen0bit why the workflow tries (and fails) when aps are no longer available in the database?

https://github.com/ivan-hc/AM/actions/runs/12979209934

Should unavailable apps just be ignored?

PR #1344, opened 2025-02-11
Enhancement of the Program List Functionality

jbwfu:
Introducing two new comment fields for each program entry:

```
# Type: webapp | appbundle | pkg2appimage | appimage | rollback
# Description: A one-line description for this program.
```

The `Type` field stores the installation type of the script. There may be naming or categorization inconsistencies in the current pull request.

`run_once_add_description_to_programs_scripts.sh`: This script adds the new comments to all existing program scripts.

`generate-programs-list.sh`: This script iterates through all folders in the current directory, reads the comments from all programs within the subfolders, and updates the list. It can accept a specific folder to process.

The main purpose of this pull request is to improve the maintainability of the program list. The list generation method remains the same: `$arch-apps` and `$arch-appimages`.  The `appimage` list includes AppImage programs built from the `ivan-hc/(KDE[^ ]*|Data[^ ]*pkg2)appimage.*` repositories (Type: pkg2appimage). Therefore, `$arch-appimage` is a larger list than the original. These scripts can be modified to change this.

Because all existing program scripts need to be updated, the number of changes is significant. Only the scripts are uploaded.

image

Note that several programs are currently missing descriptions, as indicated in the script output.

### DIFF
--- a/programs/generate-programs-list.sh
+++ b/programs/generate-programs-list.sh
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+set -eu
+
+generate_list() {
+    file="$1"
+
+    [ ! -f "$file" ] && { echo "Error: not found '$file'"; return; }
+
+    name=$(basename "$file")
+    type=$(sed -n "/^# Type:[\ ]*/{s///p;q}" "$file")
+    description=$(sed -n "/^# Description:[\ ]*/{s///p;q}" "$file")
+
+    if [ -z "$description" ]; then
+        echo "Warn: not description: '$file' " >&2
+    fi
+
+    line="◆ $name : $description"
+
+    if echo "$type" | grep -iqE "^(pkg2)?appimage$"; then
+        echo "$line" >> "$arch-appimages"
+    fi
+    echo "$line" >> "$arch-apps"
+}
+
+DIRS=$(find . -type d | grep "/" | sed 's:.*/::')
+[ $# -gt 0 ] && DIRS="$*"
+
+for arch in $DIRS; do
+    [ ! -d "$arch" ] && { echo "Error: directory not found: $arch"; continue; }
+    echo "Info: Generate program list for '$arch'"
+
+    [ -f "$arch-apps" ] && rm "$arch-apps"
+    [ -f "$arch-appimages" ] && rm "$arch-appimages"
+
+    find "$arch" -type f -print | sort | while IFS= read -r file; do
+        generate_list "$file"
+    done
+done
+
--- a/programs/run_once_add_description_to_programs_scripts.sh
+++ b/programs/run_once_add_description_to_programs_scripts.sh
@@ -1,0 +1,59 @@
+#!/bin/sh
+
+set -eu
+
+get_program_type() {
+    [ $# -ne 2 ] && { echo "need 2 args" >&2; exit 1; }
+    arch="$1"
+    program_name="$2"
+    file="$arch/$program_name"
+
+    if grep -qe 'APP=ffwa-*' "$file"; then
+        echo "webapp"
+        return
+    elif grep -qe '--pbundle_desktop' "$file"; then
+        echo "appbundle"
+        return
+    elif grep -qe '\./appimagetool \./pkg2appimage' "$file" \
+    || grep -qE 'SITE="ivan-hc/(KDE[^ ]*|Data[^ ]*pkg2)appimage' "$file"; then
+        echo "pkg2appimage"
+        return
+    #elif grep -qe '--appimage-extract' "$file" \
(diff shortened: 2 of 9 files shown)


jbwfu:
> The main purpose of this pull request is to improve the maintainability of the program list.

This is the purpose of my PR, to add descriptions within the program installation scripts and then generate the program list by reading the script information. This ensures that the information in the list always remains consistent with the program library (this is the key point). The implementation might be a bit complex, but once modified, future updates only require modifying the program scripts themselves and then running `sh generate-programs-list.sh` to update the list. Furthermore, adding the "Type" field is intended not only to assist with list generation but also to consider the potential for future list format changes and/or program management convenience (I'm not sure if it will actually achieve this purpose...).

> To add a description in the script is additional work not needed (and not required... if you want info about what you download, follow the related url or command... just read the code).

Adding the descriptions in these fields is a necessary measure to achieve the goal.

> You are doing a difference in templates types that is not required, pkg2appimage and appimage a re one thing... also "archives" can be used for appimages. Each script can have dedicated functions.

I think pkg2appimage programs need additional potential information: These are AppImage files converted from official non-AppImage files. If it is not needed, you can modify the script to determine and remove it.

> for example, "rollback" is just a sort of AM-updater that can be used in option downgrade for apps not hosted on github... see blender and kdenlive, they have their own in the script.

I haven't learned about this type of file in detail, but I just added a classification for it according to the existing template. It seems more appropriate to remove the rollback type.

> my kdegames and kdeutils are two appimages containing several programs... the appimages list only lists the unique items, helpers are not included
>
> I've not understand what you want to do with the pkg2appimage one.... also you said that the list is even bigger, what do you mean?

Regarding "bigger," as you said, I seem to have included the subprograms of kdegames and kdeutils in `$arch-appimages`, which seems unnecessary. This can be avoided by adding additional judgment processing for kdegames and kdeutils programs, but it may require using an additional "Type," such as appimage-attach, for the included subprograms.

> * `libfuse2` is hidden on purpose, it is not a program, but a system-wide library
> * `vivaldi` should be a script I forgot to remove, since I maintain the x86_64 appimage, and the aarch64 one has been removed time ago since it was just a prtable version extracted from the .deb package
> * `zramen`, I don't  know if I left it out of the list by mistake, I must check
>
> Anyway, I try to remove scripts not in the list to prevent their installation... but if they are in place and not listed, you can't get the markdown via `-a` at least.

I can add an additional check to exclude specific programs.

---

I'm not sure if it's necessary to continue improving this PR. As you said, it will add a certain amount of complexity, and the return seems not very obvious? (Although I don't think so)

If you think it's necessary to continue doing this, please give corresponding modification suggestions (I'm not familiar with the complete process of AM management programs. After this PR is improved, it will hardly have any impact on the current program in terms of results, except for maintainability...)

ivan-hc:
@jbwfu but what maintenance are you talking about?

The only maintenance are needed on scripts are:
- check if the urls have been changed upstream
- check if the file type has been changed upstream
- check if the source still exists

...I have already a workflow that mades me aware of this https://github.com/ivan-hc/AM/actions/workflows/test-apps.yml

Even updating the lists is something that normally takes a few seconds, and it reconnects to the catalog https://portable-linux-apps.github.io/ thanks to this script https://github.com/ivan-hc/AM/blob/main/tools/am2pla-site

To rewrite a script, I can also use `-t` without even having to specify a description, but only the command (which is often already complex in itself).

Finally, the "rollback" template is for illustration purposes, even for me, if I have to add it to a script. It has its reason to exist.

As I see it, this PR does not solve a problem, rather, it adds complexity to something that can normally be solved in few minutes or even seconds.

I don't know how from the three scripts related to Flatimage (all belonging to a single developer, who I know and who did not inform me by the way) and `rimage` we ended up with this need of yours to add these complexities.

**I appreciate your effort**, but I know how hard it is to maintain this project, and I can't see any real benefits in this change. With my collaborator I tried to make the scripts as similar as possible, just to simplify maintenance... and somehow we succeeded, after almost 4 years and templates that were 12 instead of 4, with only 3 lines of difference in each script and the possibility to add .desktop file, icon and "rollback" scripts optionally, if needed (for archives).

jbwfu:
> But what maintenance are you talking about?

The purpose of this PR was to simplify the maintenance of the program list based on my understanding. Everything else was secondary.

However, it seems I don't have a sufficient understanding of this project. It has a complete detection and release process, and even categorization is already implemented.

I apologize, this PR seems to be redundant... Thank you for your patient response.

ivan-hc:
For any doubts and if you need anything, I'm here

Samueru-sama:
Looks like my comment at a certain go project made someone instantly rush a PR here 👀

I had a similar experience when I found out about AM, I made a PR for a feature that was already added lol.

PR #1371, opened 2025-02-25
Add kdeconnect

Samueru-sama:
[kdeconnect.md](https://github.com/user-attachments/files/18972340/kdeconnect.md)

### DIFF
--- a/programs/x86_64-appimages
+++ b/programs/x86_64-appimages
@@ -872,6 +872,7 @@
 ◆ kardi : Application for taking notes with capabilities to view all content from any device while keeping all data accessible by you only.
 ◆ katvan : A bare-bones editor for Typst files, with a bias for Right-to-Left editing.
 ◆ kcov : Unofficial, An AppImage of Kcov code coverage tool.
+◆ kdeconnect : Unofficial, A project that enables all your devices to communicate with each other.
 ◆ kdegames : Unofficial, Full & updatable KDE Games suite, metapackage.
 ◆ kdenlive : A powerful Video Editor provided by KDE.
 ◆ kdenlive-daily : A powerful Video Editor provided by KDE (daily builds).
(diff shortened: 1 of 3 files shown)


zen0bit:
Thought that will work as expected without qt/kde libs...
😢 @Samueru-sama

Tried accept file or open plugin settings
```
QQmlComponent: Component is not ready
file:///tmp/.mount_kdecoremp941301078772902596/shared/lib/qt6/qml/org/kde/kirigami/PageRow.qml:1078: Error: Error while loading page: qrc:/qt/qml/org/kde/kdeconnect/app/PluginSettings.qml:11 Cannot load library /tmp/.mount_kdecoremp941301078772902596/shared/lib/qt6/qml/org/kde/kitemmodels/libitemmodelsplugin.so: libKF6ItemModels.so.6: (can't open file common error message here)
```
If I try to send file (from AppImage), kdeconnect crashes instantly

Samueru-sama:
Try again.

kdeconnect depends on qml, however most of the qml libs don't get added because I can only open the app in the CI, I cannot simulate the app transfering a file or similar which is when kdeconnect tries to use such libs, I originally did copy the entire qml dir, which is likely already overdeploying a lot of stuff, but that of course doesn't cover the dependencies of the qml libs themselves which was the problem here. 

Note I don't have kdelibs on my system either, I just never used that feature lol, instead I use the filesystem share feature where I can mount my phone directly on my PC.

zen0bit:
I spotted (yet) only 1 missed lib libKF6ItemModels.so.6 part of `qt6-declarative` on void...

Samueru-sama:
Try again 👀

zen0bit:
qrc:/qt-project.org/imports/QtQuick/Dialogs/quickimpl/qml/FileDialog.qml:8 Cannot load library /tmp/.mount_kdecoremp10859558893681425582/shared/lib/qt6/qml/QtQuick/Controls/Basic/impl/libqtquickcontrols2basicstyleimplplugin.so: libQt6QuickControls2BasicStyleImpl.so.6
`libQt6QuickControls2BasicStyleImpl.so.6`

Samueru-sama:
Try again 👀

zen0bit:
Better...

No complains about .so file, but xml complains stays..  
  
Needs more investigating

later...

PR #1400, opened 2025-03-12
CI: trigger - WIP

zen0bit:
Can be merged after `TOKEN` added to repo for workflows in amchceck repo

preview


ivan-hc:
Please can you explain point-by-point what your workflows do?

I also seen that you removed the lister for AppImages... an as simple as quick and useful workflow that had the only task of quickly update the lists of AppImages when normal applications lists are updated... does your implementation made the process even long?

Also, what "Clean main" does?

Also... if we already have Test-100, why also adding the workflow to expand variables?

zen0bit:
### ~Lister~ Sync AppImages listing
Lister is moved to amcheck repo and bit tweaked.
 Now can:
- work from different repository
- be triggered from anywhere

### Test-apps.yml
Also moved to amcheck repo and tweaked
 Now can:
- work from different repo
- also in forks (checking forks not upstream repo)
- be triggered from anywhere

### Check variables in 250 apps
- just trigger for workflow already in amcheck repo

NEW:

### trigger
- Will trigger our workflows in amcheck repo from actions in AM
(screenshot above)

### autotrigger WIP
- Will trigger specific workflow for file changed
(Will be split into parts)

### clean
- create "cleaner" branch named 'main-clean' from 'main'
- rebase without empty changes (PRs) useless commits just cluttering history
- All changes/commits stays same, except empty ones
(You already expressed yourself that you don't want cleanup history, but...)

ivan-hc:
@zen0bit thanks, but again... actually what you propose is a bit too much compared to what I need

if you saw the latest commits, I'm still trying to optimize Test100 according to my needs

adding a new workflow that does unnecessary things doesn't help me

i need to improve Test100, because i need to check exactly which scripts are broken and which are not, i also want to make sure that the errors are not false alarms, and i also want to extend the workflows to the currently filtered/blacklisted apps, so that their scripts are also tested

to do this, i need to focus on a single workflow

i repeat, the AppImage lister must be a SEPARATE workflow! It must last a few seconds. Triggering a single big workflow for such a small and fast operation... seems too pretentious to me

--------------------------------------------

## Question

why do you need the other workflows?

how does cleaning history and expanding variables (where scripts are already tested in Test100) help core functionality of AM?

ps: core functionality is that scripts must work, that's what I need Test100 for... I don't need anything else

ivan-hc:
I think I'm done with the improvements in Test100

https://github.com/oSoWoSo/AM/compare/trigger...ivan-hc%3AAM%3Amain

maybe, on the existing workflow, I need a step after installing the dependencies in Ubuntu, to verify that Torsocks works... to avoid it failing

the problem is related to the runner in use, which sometimes is "drunk" and does not run Torsocks

that said, we already have a great base to work on... there is no point in adding things that I will not use for the maintenance of the scripts. What we have is already too good.

zen0bit:
i can live without triggers now

PR #1437, opened 2025-03-30
fix(APP-MANAGER): Use _clean_all_tmp_directories_from_appspath

albertonoys:
Use existing function to remove tmp directories to avoid removing tmp directories from apps that are not handled by am.

I has having errors when running `am -u`, from directories unrelated to the apps managed by AM:

`rm: cannot remove '/opt/sophos-spl/tmp': Permission denied`


ivan-hc:
thanks for reporting this

PR #1456, opened 2025-04-13
fix broken non-root check

Samueru-sama:
right now this is what is being added to the remover instead due to the usage of double quotes: 

image

This change using single quotes fixes it. 

image


ivan-hc:
Thanks for this fix, just reinstalled all apps with `am reinstall --all`

PR #1459, opened 2025-04-14
WIP: Localization

### LINKED ISSUE #1449: AM in native language
Would be nice for people not yet learned english language

@ivan-hc what about AM in `lingua italiana`

PS: Already did that in my other project, so if will be allowed. I will make initial PR to dev branch soon.

zen0bit:
### This PR will add localization support to AM
(for use in native language)

Tasks:
- [x] Add $ to strings for extraction
- [x] Created simply extraction script
- [ ] Translate text hidden in variables
- [ ] Rewrite and translate `read -p` commands
- [ ] Rewrite and translate `yes`, `no`
- [ ] Add anything missed
- [ ] Create translated language files
- [ ] Add localization support to AM
- [ ] Translate to all supported language
- [ ] Finalization

Optional: (will follow later?)
- [ ] Use (AI, DeepL or GitLocalize ...) for raw translation
- [ ] Use web translation service (weblate or other)

Closes #1449


ivan-hc:
Since I delete the "dev" branch to keep it in sync with "main", I diverted the PR to "main", so as not to close it by mistake.

You can continue working on the PR, then when the time comes, I will redirect it back to "dev".

Sorry for the inconvenience.

ivan-hc:
The "trans.sh" script could very well become a "locale.am" module that can be used as a source from the start.

For this occasion, some messages before the "MODULES_PATH" definition will probably need to be moved further down. This is just my consideration.

However, I see that you are using `gum` in the script, which is not a common command. I would prefer not to use it.

zen0bit:
> The "trans.sh" script could very well become a "locale.am" module that can be used as a source from the start.

👍

> For this occasion, some messages before the "MODULES_PATH" definition will probably need to be moved further down. This is just my consideration.

I see
> However, I see that you are using `gum` in the script, which is not a common command. I would prefer not to use it.

For sure (now is optional) 😉

PS: I have AM working in Czech language (help message not yet translated)
Some message strings are not exported correctly, so manual fix have to be applied for now.

Some strings are taken from variables so not yet translated

Prepared empty translation file for others
(Created empty `it` translation file)

modules/translate.am
```diff
@@ -0,0 +1,112 @@
...
+# To add new language, add language code to LINGUAS file
+echo 'cs' > po/LINGUAS
+echo 'it' >> po/LINGUAS
+mapfile -t linguas <po/LINGUAS && \
+ for i in "${linguas[@]}"; do
+   mkdir -p usr/share/locale/"${i}"/LC_MESSAGES/ && \
+   msgfmt -o usr/share/locale/"${i}"/LC_MESSAGES/am.mo po/"${i}".po
+ done
+
+gumQ "Copy files to /usr/share/? (sudo needed)" && sudo cp -r usr/share/locale /usr/share/
```
ivan-hc, line 112: we don't use `sudo` but $SUDOCMD that may vary depending if the AM user uses `sudo`, `doas` or the root account... and AppMan users don't use root privileges at all

Please see if you can install/use these file locally, in XDG_DATA_DIR (please @Samueru-sama join this too... XDG specifications are your specialty)

APP-MANAGER
```diff
@@ -23,6 +23,21 @@ if uname | grep -q "BSD\|DragonFly\|DynFi\|FreeNAS\|helloSystem\|OPNsense\|pfSen
 	export TERM
 fi
 
+# Use translations
+export TEXTDOMAIN=am
+# AM
+if [[ -f "/usr/share/locale/${LANGUAGE}/LC_MESSAGES/am.mo" ]] || [[ -f "/usr/share/locale/${LANG:0:5}/LC_MESSAGES/am.mo" ]] || [[ -f "/usr/share/locale/${LANG:0:2}/LC_MESSAGES/am.mo" ]]; then
+	export TEXTDOMAINDIR=/usr/share/locale
+# AppMan
```
ivan-hc, line 31: consider to use only this path for both AM and AppMan, no system/root paths

APP-MANAGER
```diff
@@ -1324,14 +1340,19 @@ case "$1" in
 		_determine_args
 		for n in $ARGPATHS; do sed -e '/notify-send/ s/^#*//' -i "$n/AM-updater" 2>/dev/null; done
 		;;
+    # Translate
```
ivan-hc, line 1343: instead of a new option... made translations installable via `am -i`, using installation scripts

zen0bit: Not thought about that..
you want include them in some listings
`am -f` or...?

zen0bit:
> languages in INSTALL and AM-INSTALLER should be integrated, I see no sense to add translations there

Removed

APP-MANAGER
ivan-hc: sure, it would be better, so new locales can be added with `am -u`

APP-MANAGER
ivan-hc: well, talking about the package that can be installed, we can put everything in one common place, for both AM and AppMan users, in a rootless way, so we have not to force root

APP-MANAGER
zen0bit: 🤔

zen0bit:
Can we merge this to dev?

PR #1489, opened 2025-04-28
Fish shell now relases official static binaries

00sapo:
I have just added the script for the static binaries in x86_64. They also offer them for aarch64, though...

I have moved the old fish script to a fish-appimage file (not sure if we should remove it since it's unofficial)

### DIFF
--- a/programs/x86_64/fish
+++ b/programs/x86_64/fish
@@ -3,25 +3,25 @@
 # AM INSTALL SCRIPT VERSION 3.5
 set -u
 APP=fish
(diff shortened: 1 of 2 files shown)


ivan-hc:
Uhm... in my opinion you can completely replace the existing script with the new one of the official static release.

I don't see a sense to have an AppImage of s SHELL. Please remove the fish-appimage script.

ivan-hc:
Done, thanks for this contribution... ciao!

PR #1528, opened 2025-05-12
remove `--no-sandbox` nonsense from apps

Samueru-sama:
I noticed that electron apps (freetube) are starting to ship with the `--no-sandbox` flag in the `.desktop` due to ubuntu's apparmor nonsense.

This patch removes this since [we already warn and tell users what do to for ubuntu.](https://github.com/ivan-hc/AM/blob/main/docs/troubleshooting.md#ubuntu-mess)

NOTE: This only fixes it for `appman`, not sure how this can be fixed with `am` btw 👀

### DIFF
--- a/modules/install.am
+++ b/modules/install.am
@@ -114,7 +114,7 @@ _check_if_script_installs_a_metapackage() {
 			return 1
 		fi
 	fi
-	
+
 }
 
 _check_if_spooky_flag_exists() {
@@ -210,6 +210,7 @@ _post_installation_processes() {
 			sed -i "s#Exec=$pure_arg#Exec=$BINDIR/$pure_arg#g" "$a" 2>/dev/null
 			sed -i "s#Exec=/usr/bin/#Exec=$BINDIR/#g" "$a" 2>/dev/null
 			sed -i "s#Exec=/opt/#Exec=$BINDIR/#g" "$a" 2>/dev/null
+			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
 		done
 	fi
 	# Export all icons for hicolor theme usage


ivan-hc:
> NOTE: This only fixes it for `appman`, not sure how this can be fixed with `am` btw 👀

Yep, the above patch is to create the launchers for AppMan, a better fix can be done in the scripts one-by-one instead, when the issue appears.

Samueru-sama:
> Yep, the above patch is to create the launchers for AppMan, a better fix can be done in the scripts one-by-one instead, when the issue appears.

I won't do that, there is like a zillion electron apps in am.

fiftydinar:
> NOTE: This only fixes it for `appman`, not sure how this can be fixed with `am` btw 👀

Should this work?

```sh
	# Patch .desktop to change paths if the app is installed locally
	if [ "$AMCLI" = "appman" ]; then
		for a in $DATADIR/applications/*-AM.desktop; do
			sed -i "s#Exec=$pure_arg#Exec=$BINDIR/$pure_arg#g" "$a" 2>/dev/null
			sed -i "s#Exec=/usr/bin/#Exec=$BINDIR/#g" "$a" 2>/dev/null
			sed -i "s#Exec=/opt/#Exec=$BINDIR/#g" "$a" 2>/dev/null
		done
	fi
        if [ "$AMCLI" = "appman" ]; then
		for a in $DATADIR/applications/*-AM.desktop; do
			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
                done
        elif [ "$AMCLI" = "am" ]; then
		for a in /usr/local/share/applications/*-AM.desktop; do
			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
                done
        fi       
```

Samueru-sama:
> > NOTE: This only fixes it for `appman`, not sure how this can be fixed with `am` btw 👀
> 
> Should this work?
> 
> ```shell
> 	# Patch .desktop to change paths if the app is installed locally
> 	if [ "$AMCLI" = "appman" ]; then
> 		for a in $DATADIR/applications/*-AM.desktop; do
> 			sed -i "s#Exec=$pure_arg#Exec=$BINDIR/$pure_arg#g" "$a" 2>/dev/null
> 			sed -i "s#Exec=/usr/bin/#Exec=$BINDIR/#g" "$a" 2>/dev/null
> 			sed -i "s#Exec=/opt/#Exec=$BINDIR/#g" "$a" 2>/dev/null
> 		done
> 	fi
> 	# Patch .desktop files to NOT contain insecure '--no-sandbox' flag for Electron apps	
>         if [ "$AMCLI" = "appman" ]; then
> 		for a in $DATADIR/applications/*-AM.desktop; do
> 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
>                 done
>         elif [ "$AMCLI" = "am" ]; then
> 		for a in /usr/local/share/applications/*-AM.desktop; do
> 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
>                 done
>         fi       
> ```

Yes but a more efficient would be this way: 

 ```shell
 	# Patch .desktop to change paths if the app is installed locally
 	if [ "$AMCLI" = "appman" ]; then
 		for a in $DATADIR/applications/*-AM.desktop; do
 			sed -i "s#Exec=$pure_arg#Exec=$BINDIR/$pure_arg#g" "$a" 2>/dev/null
 			sed -i "s#Exec=/usr/bin/#Exec=$BINDIR/#g" "$a" 2>/dev/null
 			sed -i "s#Exec=/opt/#Exec=$BINDIR/#g" "$a" 2>/dev/null
 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
 		done
 	else
 		for a in /usr/local/share/applications/*-AM.desktop; do
 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
                 done
         fi       
```

however the problem here is that editing the `.desktop` entries in `/usr/local` needs elevated privileges, so I don't know what's the best way to solve it for am.

**Comment by ivan-hc** [comment]
Guys, instead of changing the parameter "globally" using this hack, make the change to the single script that has the problem... otherwise force the electron-build community to update their workflows to be more consistent with modern times

...actually, I want to invite someone random to join the conversation, maybe you can solve it at the base. 

Someone from https://github.com/electron-userland/electron-builder/issues/8686 and https://github.com/issues/created?issue=AppImage%7CAppImageKit%7C1395

### For your attention

@mmaietta @charlag @arekzaluski @jadhavmanoj the AppImage format is evolving, and with it the challenges that new systems bring. Some are removing FUSE2 from the base distro, and some are changing kernel permissions to prevent some apps from working.

Aside from the FUSE issue that has been resolved recently (and that lacks adequate media coverage, as well as widespread adoption, like in your project), there are other challenges that my collaborators are bringing to your attention. It seems that using the `--no-sandbox` flag is something nonsensical.

I am not a user of any electron apps, to be honest... but I know that a multitude of AppImages are based on electron, and maintaining old and obsolete methods for their creation does not help either the individual apps or the reputation of AppImage itself as a packaging format.

I invite you and your community to update electron-builder to adapt to the changes implemented by individual distributions.

I, with my package manager "AM", can do nothing but inform users that their system has restrictions on Namespaces or that individual apps still depend on the obsolete and insecure libfuse2 library.

You, however, who provide the basis for so many projects, are invited to eradicate the problem, if you want a much wider adoption of AppImage.

**Comment by Samueru-sama** [comment]
>Guys, instead of changing the parameter "globally" using this hack, make the change to the single script that has the problem... otherwise force the electron-build community to update their workflows to be more consistent with modern times

Ivan the problem is that electron builder is doing this nonsense to please ubuntu's mess with disabling unpriv namespaces. ubuntu is not going to disable that and nobody wants to deal with users complaining that their appimage doesn't work because they don't understand that ubuntu disabled a feature of the kernel. 

**This is not safe to be running electron apps with --no-sandbox**

Are you ok with having am users having their applications with `--no-sandbox` in the desktop entry? Why did we even bother to check and warn about that before if you are ok with this...

**Comment by Samueru-sama** [comment]
> I don't know if `--no-sandbox` is useful or not for some systems

The only systems where `--no-sandbox` is useful are the [systems where am will give this warning](https://github.com/ivan-hc/AM/blob/main/APP-MANAGER#L311-L321).

And there is no need to grep for `--no-sandbox`, just always remove it. 

>...it is a check of the same type as the one we adopt for StartupWMClass, where required.

You originally wanted to automate this for all apps and couldn't because it is not possible to know the window class without opening the app.

**Comment by ivan-hc** [comment]
> Yes but a more efficient would be this way:
> 
> ```shell
> 	# Patch .desktop to change paths if the app is installed locally
> 	if [ "$AMCLI" = "appman" ]; then
> 		for a in $DATADIR/applications/*-AM.desktop; do
> 			sed -i "s#Exec=$pure_arg#Exec=$BINDIR/$pure_arg#g" "$a" 2>/dev/null
> 			sed -i "s#Exec=/usr/bin/#Exec=$BINDIR/#g" "$a" 2>/dev/null
> 			sed -i "s#Exec=/opt/#Exec=$BINDIR/#g" "$a" 2>/dev/null
> 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
> 		done
> 	else
> 		for a in /usr/local/share/applications/*-AM.desktop; do
> 			sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
>                 done
>         fi       
> ```
> 
> however the problem here is that editing the `.desktop` entries in `/usr/local` needs elevated privileges, so I don't know what's the best way to solve it for am.

```
$SUDOCMD sed -i "s#--no-sandbox##g" "$a" 2>/dev/null
```

ivan-hc:
> > I don't know if `--no-sandbox` is useful or not for some systems
> 
> The only systems where `--no-sandbox` is useful are the [systems where am will give this warning](https://github.com/ivan-hc/AM/blob/main/APP-MANAGER#L311-L321).
> 
> And there is no need to grep for `--no-sandbox`, just always remove it.
> 
> > ...it is a check of the same type as the one we adopt for StartupWMClass, where required.
> 
> You originally wanted to automate this for all apps and couldn't because it is not possible to know the window class without opening the app.

in this case, it is possible to identify the systems where the problem is, and then apply the patch.

And anyway, since AM warns and the user who uses AM should apply the patch... even the installation scripts of the individual apps are usable via AM, so a line should be added to the script itself, where necessary.

The hacks applied to all launchers... well, that doesn't seem like the right way to do things. There are many electron apps, but they are not all. The fix should be applied to individual apps, and therefore to individual scripts.

---------------------------------

If, as you say, the developers of electron-build do it only for Ubuntu, I think it is right that they work on a different method... since Ubuntu is just one distro among many, oriented to servers more than anything else (and to its own packaging system, Snap, to the detriment of other formats, including DEB), so the focus should be on Desktop-oriented distributions.

Ubuntu is dead (on desktops), as far as I'm concerned.

Samueru-sama:
>in this case, it is possible to identify the systems where the problem is, and then apply the patch.

**The problem applies to all systems** Because the issue is the bloody `--no-sandbox` flag that is used as a workaround to get electron apps to work on ubuntu. 

That is not safe, and everyone unknowingly ends up with that even non ubuntu users and ubuntu users that use am [because we already warn them about the issue](https://github.com/ivan-hc/AM/blob/main/APP-MANAGER#L311-L321).

**Once again it makes no sense to have this warning if you are ok with people using `--no-sandbox`.** They will just disable the apparmor restriction for nothing.

>And anyway, since AM warns and the user who uses AM should apply the patch... 

Alright you prefer a warning if we detect `--no-sandbox` in the `.desktop` entry?

>The hacks applied to all launchers... well, that doesn't seem like the right way to do things. There are many electron apps, but they are not all. The fix should be applied to individual apps, and therefore to individual scripts.

Yes, but it will only patch the launchers that contain `--no-sandbox`, which is likely more than 500 applications in AM. if you want to patch them individually you are free to do it. 

This is also a security issue but you are still stuck on "hacks"  and that "it is not the right way"...

>Ubuntu is dead (on desktops), as far as I'm concerned.

The doesn't make the problem disappear...

fiftydinar:
> This is also a security issue but you are still stuck on "hacks" and that "it is not the right way"...

Yes, I think that both ways are hacks, but the global hack being better if it can be achieved, as there would be no worry to patch new Electron apps.

ivan-hc:
> > The hacks applied to all launchers... well, that doesn't seem like the right way to do things. There are many electron apps, but they are not all. The fix should be applied to individual apps, and therefore to individual scripts.
> 
> Yes, but it will only patch the launchers that contain `--no-sandbox`, which is likely more than 500 applications in AM. if you want to patch them individually you are free to do it.

But how did we do it until now? If you had a problem with an app... it was enough to open an issue or a PR to fix the script. We have always worked this way. So we try to adopt "targeted" solutions instead of involving other programs.

I have known about this `--no-sandbox` flag since before I started building AppImage packages myself, so it's been more than 5 years. Is it possible that the problem is only now manifesting itself?

Finally, I am trying to push for a radical solution, involving upstream developers of electron-builder. I believe it is also in their interest to make their packages more effective and future-proof. So do not draw conclusions about why "they don't want to do it", whether this is true or not, until proven otherwise it remains your opinion. And I think it is more constructive to talk to them directly... and I am surprised that I was not called upon to join the dialogue.

ivan-hc:
> Yes, I think that both ways are hacks, but the global hack being better if it can be achieved, as there would be no worry to patch new Electron apps.

If people consider AppImages a crappy packaging format, it's because there are still people using old, obsolete and insecure methods to create them.

Can AM always fix upstream mistakes?

As much as I try to add features to AM to improve existing AppImages... the developers of those will continue to build them badly, and without anyone telling them.

¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯¯

### I address the electron-builder community: do you really want to not improve the situation?

ivan-hc:
> You knew about this issue and did nothing to fix it before?

I don't repeat myself a second time :point_down: 

> > Yes, I think that both ways are hacks, but the global hack being better if it can be achieved, as there would be no worry to patch new Electron apps.
> 
> If people consider AppImages a crappy packaging format, it's because there are still people using old, obsolete and insecure methods to create them.
> 
> Can AM always fix upstream mistakes?
> 
> As much as I try to add features to AM to improve existing AppImages... the developers of those will continue to build them badly, and without anyone telling them.
> ### I address the electron-builder community: do you really want to not improve the situation?

AM must be the way, not the solution. I will continue to add patches endlessly for all the mistakes that are made upstream... and I will not fix anything, because upstream will continue to make mistakes!

Honestly, I've been working on improving AppImage for years, and my project and I are still in the shadows. Do you really want to keep chasing other people's mistakes? I don't. I'm tired. I would like things to be done right! If there is a possibility to create better packages using better and newer tools, we might as well do it now! Otherwise it is better that AppImage electron and AppImage are no longer produced. Because where there are developers who create excellent packages for AppImage, on the other hand there will always be poor packages, which will cast a shadow on the work of those who really want to support AppImage!

Do you really want to end up in oblivion? I don't. It's an insult to the years of sacrifice I've put into AppImage, here on github. I would like to see some light, damn it.

Samueru-sama:
Are ok if I come and patch the +500 individual apps with `--no-sandbox` right? 

yes or no?

mmaietta:
> I address the electron-builder community: do you really want to not improve the situation?

Maybe I'm missing something here as I'm quite unfamiliar with AppImages under-the-hood. 

Appimage `type2-runtime` is on my lengthy todo list of investigations as I am attempting to build pipelines for compiling each release artifact in a reproducible manner, for both github artifact attestation and more secure distribution, from electron-builder-binaries (the upstream dependency of electron-builder for on-demand tooling that I now have write access to).

Samueru-sama:
> > I address the electron-builder community: do you really want to not improve the situation?
> 
> Maybe I'm missing something here as I'm quite unfamiliar with AppImages under-the-hood.
> 
> Appimage `type2-runtime` is on my lengthy todo list of investigations as I am attempting to build pipelines for compiling each release artifact in a reproducible manner, for both github artifact attestation and more secure distribution, from electron-builder-binaries (the upstream dependency of electron-builder for on-demand tooling that I now have write access to).

This is an unrelated issue that Ivan pinged you about. 

It seems all electron appimages have the `--no-sandbox` option in the desktop entry, regardless if unpriv namespaces are disabled on the host. 

this is not safe and it is something I'm trying to fix, since we make sure in am that the user has this enabled already.

Right now I'm scrapping the am database to get all the electron apps with `--no-sandbox` the list is over 500 rn

PR #1565, opened 2025-05-23
chore: Don't auto-generate shell completion when `am` is in `/usr/bin/`

fiftydinar:
Shell completions should be supplied by the packager/distro in this scenario.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -528,21 +528,23 @@ _completion_lists() {
 }
 
 # BASH, FISH AND ZSH COMPLETION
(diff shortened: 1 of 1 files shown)


fiftydinar:
Closing, because shell completions are dynamic based on the constantly updated application list, so it's a better experience to have shell completions in `$HOME`

fiftydinar:
Actually, it's easy to implement this in distro-level, just apply this for bash in `/usr/share/bash-completion/completions/am`:

`complete -W "$(cat $HOME/.local/share/AM/list 2>/dev/null)" am`

PR #1569, opened 2025-05-24
Add a message summarizing Fedora's mess with GNU Wget2

ivan-hc:


ivan-hc:
@Jonathing thanks for your feedback. Can you suggest a better message?

I would also prefer to summarize everything.

ivan-hc:
Istantanea_2025-05-24_21-16-21

?

ivan-hc:
@Jonathing I think this is what you prefer

Istantanea_2025-05-24_21-25-14

PR #1574, opened 2025-05-24
Reduce the message about Fedora's mess with GNU Wget2

ivan-hc:
Quick patch for https://github.com/ivan-hc/AM/pull/1569

The message was too big and invasive, so I decided to reduce it to two lines

Istantanea_2025-05-24_21-16-21

-------------------------------------

the message is to indicate this mess :point_down: 

Istantanea_2025-05-24_21-25-14

-------------------------------------

The URL in use is the following

https://bugzilla.redhat.com/show_bug.cgi?id=2358792

...and will only appear if the distro is based on Fedora.

I preferred to add the URL to give more context and better instructions than I could.

This is the function responsible of this
```
_check_fedora_mess() {
	if wget --version | head -1 | grep -qi ' 2.\|wget2 '; then
		printf "\n%bWARNING: your distro maintainer ships \"wget2\" as \"wget\" \033[0m\n\n" "${RED}" | _fit
		[ -f /etc/redhat-release ] && printf "See https://bugzilla.redhat.com/show_bug.cgi?id=2358792\n\n" | _fit
	fi
}
```

Other distros will have a different message in the future, if they decide to do the same stupid thing the Fedora team did on 2024... but I'm confident enough to believe that no one will follow their bad example. Or at least I hope.

PS: I have nothing against Fedora, it's a fantastic distro.

For more context on the situation, I refer you to my last post, on Reddit

https://www.reddit.com/r/Fedora/comments/1kuag0y/gnu_wget_has_been_reintroduced_alongside_wget2/

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="9.7.1-3"
+AMVERSION="9.7.1-4"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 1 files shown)


Jonathing:
Hey. Sorry for not responding earlier, I left the house for a bit and I don't have access to GitHub outside of my computer currently.

I think this is a pretty good solution. It would've been ideal if it only displayed on installation errors. But, this gives users the notice if something is wrong with wget on their systems, which is important. Thanks again.

ivan-hc:
@Jonathing thanks to you for the suggestion. If you need I'm here.

ivan-hc:
@Jonathing just added some instructions with color https://github.com/ivan-hc/AM/commit/42dd432155c8d34cbdb4d6d9ea3de1fb4495b3a6

this should be helpful

Istantanea_2025-05-25_08-53-17

thanks to @Shikakiben for suggestion

PR #1610, opened 2025-06-02
Improve language detection logic

fiftydinar:
All symbols after `_` or `@` are variants of one language, so I just remove all those symbols.
AM only supports translations of 1 variant per language.

As a result, detection is `sr` instead of `sr_RS`.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -59,7 +59,7 @@ _check_valid_iso_codes() {
 			language_code="$lang"
 		fi
 	done
-	[ -n "$language_code" ] && iso_code="$language_code"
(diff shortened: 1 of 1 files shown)


fiftydinar:
Do not merge yet, I also have to do this for `@` symbol.

fiftydinar:
> Do not merge yet, I also have to do this for `@` symbol.

Did it, it can be merged.

fiftydinar:
I think that `cut -c -2` is not needed here, so I'll remove it (for example, there's `nds` locale, which has 3 letters)

```sh
[ -f "$DATADIR/AM/locale" ] && iso_code="$(cat "$DATADIR/AM/locale" | head -1 | cut -c -2)"
```

PR #1613, opened 2025-06-02
fix nasty bug

Samueru-sama:
This is wrong and I mistake that didn't discover until @fiftydinar pointed it out in the discord. Because the `~` is in quotes it never expands to the value of `$HOME`.

Amazingly this problem never showed up before because aisap resolves the literal `~`.

### DIFF
--- a/modules/sandboxes.am
+++ b/modules/sandboxes.am
@@ -109,13 +109,13 @@ _generate_sandbox_script() {
 	  eval $DIR="$XDG_DIR"
 	done
 	# Use default location if var is not set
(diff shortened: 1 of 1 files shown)


fiftydinar (approved):

ivan-hc:
> This is wrong and I mistake that didn't discover until @fiftydinar pointed it out in the discord. Because the `~` is in quotes it never expands to the value of `$HOME`.
> 
> Amazingly this problem never showed up before because aisap resolves the literal `~`.

Good catch... "XDG lawyer"

kek xD

PR #1618, opened 2025-06-03
Don't install desktop files for CLI AppImages

fiftydinar:
Desktop files should be installed for GUI applications only.

We miss the template for CLI AppImages, which would omit that by default.

Let me know if there are more AppImages that are in CLI, but still install the desktop file.

@Samueru-sama I also did this for `sas`, let me know if you agree with this.

### DIFF
--- a/programs/x86_64/aisap
+++ b/programs/x86_64/aisap
@@ -55,23 +55,3 @@ else
 fi
 EOF
 chmod a+x ./AM-updater || exit 1
-
-# LAUNCHER & ICON
-./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
-./"$APP" --appimage-extract .DirIcon 1>/dev/null && mv ./squashfs-root/.DirIcon ./DirIcon
-COUNT=0
-while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
-	if [ -L ./"$APP".desktop ]; then
-		LINKPATH=$(readlink ./"$APP".desktop)
-		./"$APP" --appimage-extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./"$APP".desktop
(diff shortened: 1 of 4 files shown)


ivan-hc:
yes, this should be done for all AppImages that do not have GUI but only command line

if you know others, add them... you told me Aisap on Discord, right?

fiftydinar:
> yes, this should be done for all AppImages that do not have GUI but only command line
> 
> if you know others, add them... you told me Aisap on Discord, right?

Yes, `aisap`.

When I think about it a bit more, `android-tools` should also have desktop file removed.

I'll check for others too.

Maybe a future template for AppImage CLI would be good.

ivan-hc:
> Maybe a future template for AppImage CLI would be good.

This will mean... a new line to translate :eyes:

ivan-hc:
@fiftydinar Also note that some users, when asking to add some apps, do not say if it is CLI or Desktop, perhaps because they do not speak much English and cannot express themselves.

Therefore, a pull request like this would be enough. Anyone can do it, it does not involve adding code, but only removing, so nothing easier.

It would also make people more engaged in the project, in my opinion.

Otherwise we should document every app we add, even if we never use it.

fiftydinar:
> @fiftydinar Also note that some users, when asking to add some apps, do not say if it is CLI or Desktop, perhaps because they do not speak much English and cannot express themselves.
> 
> Therefore, a pull request like this would be enough. Anyone can do it, it does not involve adding code, but only removing, so nothing easier.
> 
> It would also make people more engaged in the project, in my opinion.
> 
> Otherwise we should document every app we add, even if we never use it.

That's also ok, we can remove desktop file if needed afterwise.

Let me know if you find some more AppImages in CLI

ivan-hc:
> That's also ok, we can remove desktop file if needed afterwise.
> 
> Let me know if you find some more AppImages in CLI

Sure, I'm starting to merge these 5, if you find others, let me know with a pull request.

Adding another template when it would be enough to modify the existing one manually, would be a waste of code.

PR #1639, opened 2025-06-09
optimize with shell builtins

Samueru-sama:
image

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -7,8 +7,9 @@ AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
 
 # Determine the name of this script and its working directory
 export REALDIR="$PWD"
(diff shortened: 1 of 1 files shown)


ivan-hc:
@Samueru-sama can you do the same for `-f`? I noticed that at first run it take 2-3 seconds

Samueru-sama:
> @Samueru-sama can you do the same for `-f`? I noticed that at first run it take 2-3 seconds

That's a lot of work. 

There is also a lot of:

```shell
entries="$(echo "$@" | cut -f2- -d ' ')"
for arg in $entries; do
...
```

This is very slow, it spins another shell instance to shellout `cut` on the array. When shell has this built in: 

```shell
shift # drops the first element in array
for arg in "$@"; do
...
```

ivan-hc:
no commitment, don't worry

PR #1701, opened 2025-07-07
Export native shell built-ins to modules

fiftydinar:
Exporting those globally is not a good practice, so let's export those to what AM uses.

Maybe those can be done too for templates & app scripts in the future.
For now, this covers the basics.

`sandboxes` module uses `POSIX` & `<( )` file redirection is not allowed there, so I'll have to see the alternative, or it can be switched to use `bash` like the other modules.


ivan-hc:
I get a lot of `sed: impossibile leggere am: File o directory non esistente` because "$AMCLIPATH" is not recognized as a file

PR #1703, opened 2025-07-07
Use shell built-ins for basename & dirname

ivan-hc:


ivan-hc:
@fiftydinar @Samueru-sama I think that to specify this in each module is not needed. We already have `sed` and `less` acting as functions, also in modules.

Is it worth to revert this change only in modules?

PR #1721, opened 2025-07-26
Option -a, code rewrite for better detection of installed apps

ivan-hc:
The status of installed and uninstalled apps is determined by cataloging scripts in the hidden directory, and if present, they will be used as a reference.

Visually, not much should change, except for a correct reading of available apps, which previously relied on reading offline list files.

The only regression is that apps not available in any database will still be treated as available in some database. I haven't been able to do anything about this yet.


ivan-hc:
> The only regression is that apps not available in any database will still be treated as available in some database. I haven't been able to do anything about this yet.

solved at https://github.com/ivan-hc/AM/pull/1721/commits/3dea63bcc850ddf7f800cdaba082e7bd58dcf47a

PR #1742, opened 2025-08-03
add `sas` integration

Samueru-sama:
Tested it working with `pavucontrol-qt`. 

Note I removed all the options to install aisap from sandboxes.am, it is just a mess and extending it to include sas will just make it take more code than the entire sandbox script, people can manually do `am -i aisap` or `am -i sas` instead.

### DIFF
--- a/modules/sandboxes.am
+++ b/modules/sandboxes.am
@@ -54,7 +54,19 @@ _check_aisap() {
 	elif ! command -v aisap 1>/dev/null && ! command -v sas 1>/dev/null; then
 		printf $'\n%s\n' " ERROR: You need aisap or sas for this script to work"
 		printf $'\n%s\n\n' " NOTE: Only sas can sandbox DWARFS AppImages"
-		return 1
+		read -r -p $" ◆ DO YOU WISH TO INSTALL SAS? Install size <5 MiB? (Y/n) " yn
+		case "$yn" in
+			n*|N*) echo $" OPERATION ABORTED!"; return 1;;
+		esac
+		if [ "$CLI" = am ] && [ -f "$APPMANCONFIG"/appman-config ]; then
+			read -r -p $" ◆ DO YOU WISH TO INSTALL SAS LOCALLY? (Y/n) " yn
+			case "$yn" in
+				n*|N*) "$AMCLIPATH_ORIGIN" -i --user sas >/dev/null 2>&1;;
+				''|*)  "$AMCLIPATH_ORIGIN" -i sas >/dev/null 2>&1;;
+			esac
+		fi
+		command -v sas 1>/dev/null || return 1
+		echo $" aisap installed successfully!"
 	fi
 	if [ -f "$BINDIR"/"$1" ]; then
 		SUDOCMD=""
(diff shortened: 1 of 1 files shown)


ivan-hc:
> Note I removed all the options to install aisap from sandboxes.am, it is just a mess and extending it to include sas will just make it take more code than the entire sandbox script

I don't agree with the removal of the automated installation of `aisap`, on the contrary... if `sas` is a rewrite of `aisap` that you think is better, have `sas` installed instead of `aisap`

ivan-hc:
@Samueru-sama please restore the code to install `aisap`, but change
```
"$AMCLIPATH_ORIGIN" -i aisap >/dev/null 2>&1
```
to
```
"$AMCLIPATH_ORIGIN" -i sas >/dev/null 2>&1
```
and any reference to AISAP with SAS

Samueru-sama:
> I don't agree with the removal of the automated installation of `aisap`, on the contrary... if `sas` is a rewrite of `aisap` that you think is better, have `sas` installed instead of `aisap`

😫

Note that right now there is a lot of nonsense in the sandboxes module because of issues with aisap, like having to check the values of the xdg variables because aisap has them hardcoded. 

Sure if aisap removed then all of that can be removed as well, but undoing all of those changes will take time so I will instead restore the install but for sas only, the rest can still use aisap as well.

Samueru-sama:
check if the install thing is working correctly in am. 

I noticed a bug that it thinks that I'm using `am` because `$CLI = am` due to me having `appman` symlinked as `am`.

ivan-hc:
> Note that right now there is a lot of nonsense in the sandboxes module because of issues with aisap, like having to check the values of the xdg variables because aisap has them hardcoded.
> 
> Sure if aisap removed then all of that can be removed as well, but undoing all of those changes will take time so I will instead restore the install but for sas only, the rest can still use aisap as well.

If you want, I'll leave this PR open so you can work comfortably. You know this module better than I do.

> check if the install thing is working correctly in am.
> 
> I noticed a bug that it thinks that I'm using `am` because `$CLI = am` due to me having `appman` symlinked as `am`.

If you did what I asked, then it will definitely work for AM too. You just had to put that function back where it was and change `aisap` to `sas`.

Regarding your specific case, I don't know what to do... perhaps a test to see if $AMPATH is read-only?

Samueru-sama:
> If you want, I'll leave this PR open so you can work comfortably. You know this module better than I do

It is fine as it is right now.

> Regarding your specific case, I don't know what to do... perhaps a test to see if $AMPATH is read-only?

Instead of:

```
CLI="${0##*/}"
CLI_PATH="$(readlink -f "$0")"
```
do

```
CLI_PATH="$(readlink -f "$0")"
CLI="${CLI_PATH##*/}"
```

NOTE: This means that `CLI` will be `APP-MANAGER` for am.

ivan-hc:
> NOTE: This means that `CLI` will be `APP-MANAGER` for am.

and what for distro packagers (such as @fiftydinar )?

Samueru-sama:
>and what for distro packagers (such as @fiftydinar )?

There is no `APP-MANAGER` in that case? 

Anyways, note this bug is unrelated to the changes in this PR.

PR #1757, opened 2025-08-08
fix broken install of sas when appman is used

Samueru-sama:
turns out this worked on my end because I had `appman` symlinked as `am`. 

I just happened to discover this wasn't doing anything when real `appman` is used 👀


ivan-hc:
$AMCLIPATH_ORIGIN is needed to record AMCLI usage, to switch between AM (system) and AppMan Mode (local).

That step in particular should be needed in case no APPMANCONFIG is set, for pure AM users

PR #1764, opened 2025-08-10
Update kitty

tsuri21:
[kitty.md](https://github.com/user-attachments/files/21706067/kitty.md)

◆ kitty : Cross-platform, fast, feature-rich, GPU based terminal

---

Now using the kitty archive, which contains both kitty and kitten. Kitten is required for kitty to function correctly. I tested it without linking kitten to `/usr/local/bin`, but it behaved a bit buggy, so now both are linked. This is also reflected in the update and remove scripts. Tested on both user and system level.

### DIFF
--- a/programs/x86_64/kitty
+++ b/programs/x86_64/kitty
@@ -3,37 +3,40 @@
 # AM INSTALL SCRIPT VERSION 3.5
 set -u
 APP=kitty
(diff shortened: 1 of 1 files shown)


ivan-hc:
thanks @tsuri21 and welcome on board!

Issue #1696, opened 2025-07-06
Substitute host dependencies for native shell built-ins or for already existing dependencies


fiftydinar:
To make AM more compatible with other distributions, this is one of the steps that needs to be done.

This is a tracker for host dependencies which can be replaced for native shell built-ins or for already existing dependencies:

Not finished:
- `file`: maybe using Azathothas `file` bash script will suffice?

Mostly finished:
- `column`: Works mostly well with GNU `awk`, has extra whitespace for symbols. Doesn't work well with BSD's `awk`.
Because of that, it's only used as a fallback if `column` is not installed.

Finished:
- `realpath`: replaced by `readlink -f`
- `basename`: replaced by native shell function
- `dirname`: replaced by native shell function

If you know more of this, let me know, list will be updated.

ivan-hc:
About `basename` and `dirname`, can be these functions complete alternatives to the commands or should we set the functions to fallback in case the command is missing?

Also, how can we do with `file`?

fiftydinar:
> About `basename` and `dirname`, can be these functions complete alternatives to the commands or should we set the functions to fallback in case the command is missing?

It's good as a fallback.

But in my testing for another script, it's a good replacement, can replace host dependencies fully.

> Also, how can we do with `file`?

I'm not sure on that one.
I'm thinking at the moment about `column`, which Ubuntu Touch didn't have installed. I think it can be done with `sed`.

ivan-hc:
First of all, lat try to collect all references

`file`
- in APP-MANAGER is used 5 times
- in database.am 4 times
- in management.am 1 time

`realpath`
- APPM-MANAGER 9
- sandboxes.am 2+2 (its used also in the created script)

`basename`
- APP-MANAGER 9
- database.am 2
- install.am and management.am 1

`dirname`
- management.am and sandboxes.am 1

`column`
- database.am 3

NOTE: I excluded to collect their usage in functions in APP-MANAGER, but they are mainly used as standalone commands

ivan-hc:
I'm already working on the dev branch

## Replace "realpath" with "readlink -f"

https://github.com/ivan-hc/AM/commit/b0de5547f9209793904c0d45fc6557cc544923c3

## What is left

> `file`
> 
>     * in APP-MANAGER is used 5 times
> 
>     * in database.am 4 times
> 
>     * in management.am 1 time
> 
> 
> `basename`
> 
>     * APP-MANAGER 9
> 
>     * database.am 2
> 
[...]

PR #1769, opened 2025-08-12
Update database.am - get rid of "file" dependency

ivan-hc:
@fiftydinar @Samueru-sama here I replace `file` with `readelf` (binutils)

if you have a better idea, let me know

partially fixes https://github.com/ivan-hc/AM/issues/1696#issuecomment-3178847484

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -102,7 +102,7 @@ _about_description_for_third_party() {
 		if [ -n "$tpnumber" ] && [ "$tpnumber" -ge "1" ]; then
 			[ "$tpnumber" != 1 ] && printf $"\nAvailable families: %b\n" "$tpnumber"
 			if echo "$app_status" | grep -q "^installed"; then
-				[ -f "$argpath/$appname" ] && printf $"\nBinary info: %b\n" "$(file "$argpath/$appname")" | fold -sw 78
+				[ -f "$argpath/$appname" ] && printf $"\nBinary info: %b\n" "\n$(readelf -h "$argpath/$appname" | sed 's/$/;/g; s/:;/:/g' | xargs)" | fold -sw 78
 			fi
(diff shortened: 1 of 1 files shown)


ivan-hc:
Options involved:
- `-f`, all dynamic/static binaries
- `-a`, third-party databases (the installed argument require an extension, for example `7z` become `7z.soarpkg`)

ivan-hc:
with this change in APP-MANAGER I removed `file` from a function that detects version in portable apps like all development branches of firefox, thunderbird and tor-browser (they update by theirself, not something managed by AM)

also, this change reduces the total  `file` references to 4 in total

we are near to get rid of `file`

the 4 left are related to PNG/SVG detection, and I hope that detection is even simpler

ivan-hc:
@Samueru-sama about PNG/SVG stuff, there are two couples of lines
```
PNG="$(file "$APPSPATH"/*/icons/* | grep -i '.png' | awk -F":" '{print $1}' | grep -vi .png)"
SVG="$(file "$APPSPATH"/*/icons/* | grep -i '.svg' | awk -F":" '{print $1}' | grep -vi .svg)"
```
the above is repeated twice, for a total of 4 `file`

You said that to identify SVG was difficult, so just use a `if`/`fi` function, where if one is detected as PSG, be it PNG, else SVG

This means, you only need to detect PNG, any other thing will be recognized as SVG

Samueru-sama:
I don't think this is a good idea at all. 

`file` is POSIX, **`readelf` is not**

ivan-hc:
@Samueru-sama , @fiftydinar says that can't use AM in Maemo Leste due to the missing `file` dependency. I also tried to bundle it as AppImage, but I should do it for all architectures. What do you suggest?

Samueru-sama:
> What do you suggest?

not introducing `readelf` as a dependency for sure. I have had to avoid readelf before, see: https://github.com/Samueru-sama/simple-appimage-sandbox/issues/1

Best you can do is just make `file` optional so that the user can at least install and run applications, which this does not need file at all.

Issue #1817, opened 2025-09-17
Print out command line to install the same set of apps on different machine


eGit:
appman --clone would print out something like:

appman -i avidemux hypnotix vlc 

Then one can use this command on a different PC to establish an identical set of apps.

Maybe this is already possible but I read the entire help and couldn't find it.

ivan-hc:
This is interesting... how a set of apps should be sent to another PC?

The quick way (in AppMan) is to clone the $HOME, maybe with `rsync` or something.

eGit:
> This is interesting... how a set of apps should be sent to another PC?

For example USB Stick or Email or Pastebin or a text file via cloud drive.  Could even be a QR code jpeg I guess.

> The quick way (in AppMan) is to clone the $HOME, maybe with `rsync` or something.

Seems complicated. Just a command "appman -i  avidemux hypnotix vlc" seems most flexible. I can edit that command and add or remove apps. A detour via 'rsync' seems moot since we have appman itself. Two commands. Install appman. Install the list of apps. Done.

ivan-hc:
If I have understand well, al you need to know is what applications are installed and have all of them on one line.

This is what I wrote now
```
_use_clone() {
	_determine_args
	for arg in $ARGS; do
		argpath=$(echo "$ARGPATHS" | grep "/$arg$")
		[ -d "$argpath"/.am-installer ] && scriptname=$(ls "$argpath/.am-installer/" | head -1)
		[ -z "$scriptname" ] && scriptname="$arg"
		if [ "$arg" != am ]; then
			[ -z "$to_clone" ] && to_clone="$scriptname" || to_clone="$to_clone $scriptname"
		fi
		scriptname=""
	done
	echo "$to_clone"
}
```

...and this is how it works

<img width="747" height="600" alt="Image" src="https://github.com/user-attachments/assets/6382dc04-61b7-435b-91f9-48b6b646bcde" />

eGit:
Looks great! Just like I imagined. Maybe there is even an "internal" use for this? What if people want to move the location of the "appimages" to a different HDD on the same PC? They probably have to do some manual work in config files? A clean way would be to uninstall am/appman and make a clean re-install on the second HDD. The downside would be that it needs to download everything again.

Or is there already a better feature where you can move the "appimages" to a new location on the same PC?

Something like appman --relocate /mnt/data/newdisk/apps ?

ivan-hc:
[...]

PR #1818, opened 2025-09-17
"AM" 9.9 - clone configurations and relocate AppMan apps

ivan-hc:
fix https://github.com/ivan-hc/AM/issues/1817

Two new options added.

## Option `clone` or `--clone`
Clone your configuration and share it to other systems. Usage:
```
am clone
```
or
```
appman clone
```
this command without arguments only looks for a file named `am-clone.source`, on the Desktop first as prior path, then in the $HOME and then everywhere in the system, also on mounted devices.

Add the flag `-i` or `install` to also install the listed apps
```
am clone -i
```
or
```
appman clone -i
```

--------------------------------------

## Option `relocate` or `--relocate`
This option is for AppMan or for AM in AppMan mode and all AppMan apps (that can be managed also via AM).

As I explain on the README, AM installs everything always in /opt, by following the Linux Standard Base. That path will not be changed.

On the contrary, AppMan installs everything wherever you want, but you were unable to change the path once you have installed it and the apps.

Now you can
```
am relocate
```
or
```
appman relocate
```

https://github.com/user-attachments/assets/4d1aee94-bcd4-4ac3-9ce5-35fde080cb03

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="9.8.2-7"
+AMVERSION="9.8.2-8"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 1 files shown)


ivan-hc:
Time to merge

Issue #1921, opened 2025-11-24
Can't install Switchcraft Application appimage


AmethystGosling169:
# Premise
By opening this issue, I certify:
- that I have updated "AM"/"AppMan" to the latest version
- that I have read the README https://github.com/ivan-hc/AM/blob/main/README.md
- that I have read the documentation https://github.com/ivan-hc/AM/tree/main/docs
- that I have searched for an option or flag in `-h` that covers my use case
- that I have searched for an issue that covers my use case https://github.com/ivan-hc/AM/issues

--------------------------------------------------

## Are you using "AM" or "AppMan"?
AM

## What GNU/Linux distribution are you using it on?
Manjaro GNOME

## Please, describe the bug
Can't install downloaded latest [switchcraft](https://github.com/kem-a/switchcraft) application appimage.

## What are the steps/commands to perform?
am -i <path_to_appimage>

After running the command the app opens and these are the logs:
```
/opt/am/modules/install.am: line 60: warning: command substitution: ignored null byte in input
 ◆ "SWITCHCRAFT-1.2.0.APPIMAGE": starting installation script


(switchcraft-1.2.0.AppImage:13870): Gtk-WARNING **: 14:34:36.496: Unable to acquire session bus: Error spawning command line “dbus-launch --autolaunch=fb911329a48c4f338743e40b49f89f7a --binary-syntax --close-stderr”: Child process exited with code 1
```

## What is the expected behavior?
It should install.

ivan-hc:
It is not so that you can install apps from this database

https://github.com/ivan-hc/AM/blob/main/docs/guides-and-tutorials/extra.md

In your case, the syntax should be
```
am -e https://github.com/kem-a/switchcraft switchcraft
```
and not
```
am -i /path/to/SWITCHCRAFT-1.2.0.APPIMAGE
```

AM installs apps using installation scripts. like these ones https://github.com/ivan-hc/AM/tree/main/programs/x86_64, the same way `yay` or `paru` does in Manjaro/Arch Linux with the PKGBUILDs from AUR.


-------------

I'll fix this by adding the script to the database.

After (and ONLY AFTER) that I have closed this issue, run
```
am -i switchcraft
```

AmethystGosling169:
It worked, thanks a lot!

But how do I install a downloaded appimage file?

ivan-hc:
> But how do I install a downloaded appimage file?

Option `--launcher`

[...]

PR #1923, opened 2025-11-24
Determine if the argument is an installation script

ivan-hc:
To prevent users from misusing the installation options, which are intended for scripts, not random files, the argument will be checked to see if it is a script, an AppImage, or another file type.

Istantanea_2025-11-24_13-30-36


ivan-hc:
This change is also related to https://github.com/ivan-hc/AM/issues/1921

PR #1955, opened 2025-12-03
change icon for Dolphine-nightly en name for Eden-nightly/dolphin-nightly 

Shikakiben:
if you don't mind for dolphin icon because natively its the same for nightly and none nightly and its confusiong

### DIFF
--- a/programs/aarch64/dolphin-emu-nightly
+++ b/programs/aarch64/dolphin-emu-nightly
@@ -59,23 +59,24 @@ chmod a+x ./AM-updater || exit 1
 # LAUNCHER & ICON
 ./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
 # Tries to get Portable-linux-Apps icon
-if wget -q --spider --retry-connrefused --tries=30 "https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/raw/main/files/icons/dolphin-emu-nightly.png"; then
-    wget -O "./DirIcon" "https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/raw/main/files/icons/dolphin-emu-nightly.png"
+if wget -q --spider --retry-connrefused --tries=30 "https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/raw/main/icons/dolphin-emu-nightly.png"; then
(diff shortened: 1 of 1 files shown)


ivan-hc:
Have you tested the scripts?

Shikakiben:
yes if you can wait a moment before  merging, it doesn't work yet  sorry i was too fast   😅

ivan-hc:
Let me know when you are ready

Shikakiben:
All good 👍 im gonna do the same whith citron and eden (if you don't mind)

ivan-hc:
> All good 👍 im gonna do the same whith citron and eden (if you don't mind)

why?

Shikakiben:
because there is no official nightly icon/desktop so the build is shiped with original icon/desktop.
so when you install both, no differnce at all in the menu... so at least the name in the desktop need to be change, but i thought also an icon more "nightly" would be good but if you don't want i understand... but a least the name?

ivan-hc:
What you are trying to add should already be covered here

https://github.com/ivan-hc/AM/blob/c8d1186fcc9b7f9a74b85ab2e76b19739dbf9c6f/modules/install.am#L231-L233

ivan-hc:
> because there is no official nightly icon/desktop so the build is ship with original icon/desktop. so when you install both, no differnce at all in the menu... so at least the name in the desktop need to be change, but i thought also an icon more "nightly" would be good but if you don't want i understand... but a least the name?

These changes should be done upstream

Shikakiben:
> What you are trying to add should already be covered here

yes but the icon is not broken but its not the one in AM

> These changes should be done upstream

i don't no about that , i think for upstream , its a user choice to choose between "nigthly" or "stable", so no need for them to make difference icon/name

but i understand your  point i can do it on my computeur 🙂

ivan-hc:
> but i understand your point i can do it on my computeur 🙂

you can do it in the script, its enough that this change is ok for all other users

Shikakiben:
> you can do it in the script, its enough that this change is ok for all other users

what do you mean ?

ivan-hc:
> what do you mean ?

nothing, keep doing what you are doing

ivan-hc:
@Shikakiben I'm on Discord, for any tip in real time, I'm there

Shikakiben:
> nothing, keep doing what you are doing

ok but if you prefer i con just change the name, change the icon is not really needed, if its not the same name its enought for not being confusing

PR #2013, opened 2026-01-06
fix massive blunder

Samueru-sama:
This was already in AM, the problem is that I had a typo in the app name. 

Sorry about this.


ivan-hc:
np

ivan-hc:
Fun fact: in "amcheck" something is wrong when in the list there is one less app than the initial count. Since we removed this app and until the check cycle is finished (7 hours ago from this comment) all workflows are failed

https://github.com/ivan-hc/amcheck/actions/workflows/AMCHECK.yml

For example this (the last one)

https://github.com/ivan-hc/amcheck/actions/runs/20751264957

To solve, I only had to reset the stats

https://github.com/ivan-hc/amcheck/actions/runs/20751745780

Not an issue at all... but it is sign that we need to add even more apps to keep amcheck working (until an installation script keep fails for some else reason).

Amcheck is the tool that gives us the guarantee to use reliable sources and working installation scripts in AM.

NOTE, some apps are blacklisted because they are not installable in github actions, but as normal user you can install them all, also slowly. This is the list of blacklisted apps (I checked manually two days ago):

https://github.com/ivan-hc/amcheck/blob/9ebb298abb4251badead9fcda8f73fd817461523/.github/workflows/AMCHECK.yml#L344-L346

PR #2067, opened 2026-02-12
Add starship : SF64 PC Port.

spinktvis:
Following the instruction video you posted in my other post I attempted my very first pull request ever, not sure if it is correct at all, so please double check. One thing I have noticed is that starship is also a prompt that is available already in AM, not sure what would be the best way to handle that name "conflict", I can already see that it attempts to overwrite the old one so there is that. A name change would probably be in order, I am open for suggestions. This script also uses zip to extract the latest release version, so please check if there aren't any mistakes and it is working correctly.

Source: https://github.com/HarbourMasters/Starship

### DIFF
--- a/programs/x86_64/starship
+++ b/programs/x86_64/starship
@@ -1,0 +1,55 @@
+#!/bin/sh
+
+# AM INSTALL SCRIPT VERSION 3.5
+set -u
+APP=starship
+SITE="starship/starship"
+
+# CREATE DIRECTORIES AND ADD REMOVER
+[ -n "$APP" ] && mkdir -p "/opt/$APP/tmp" "/opt/$APP/icons" && cd "/opt/$APP/tmp" || exit 1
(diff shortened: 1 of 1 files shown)


ivan-hc:
Hi, the existing `starship` is this https://github.com/starship/starship

Do not overwrite the existing one. Call yours `starship-emu` or `starship-emulator` instead, or another name you believe fits better.

spinktvis:
I have corrected the naming, it should not overwrite anything now, let me know if there is anything else!

ivan-hc:
Welcome on board then! ;)

ivan-hc:
I re-added the starship script that instead was removed in your fork.

If the workflow do not detects the script, the app will be removed from the list, when I merge.

All solved btw.

ivan-hc:
All ok

https://github.com/ivan-hc/amcheck/actions/runs/21985675703/job/63519157579#logs

thanks for contributing.

PR #2070, opened 2026-02-13
add 26 apps

Link4Electronics:


ivan-hc:
Jackpot! xD

Samueru-sama:
@ivan-hc All of these are unofficial but most were not marked as such.

ivan-hc:
Should I implement something in `am -t` to ask the user if the app he is adding is unofficial then?

Istantanea_2026-02-15_01-01-28

From what I see, less than 300 have "unofficial" or "unoffical" in the description.

ivan-hc:
Maybe we need to implement something better to verify if apps are official/unofficial. But how to do so?

Samueru-sama:
> Should I implement something in `am -t` to ask the user if the app he is adding is unofficial then?

Nah, just check in the PR. 

I wasn't aware link would be doing this, ran into the surprise this happened when I was preparing mine to be added.

Issue #2104, opened 2026-02-24
Appimage binary file integrity verification (security issue)


vishnu350:
# Premise
Given how am/appman is growing (2000+ appimages), do we have a way to make sure the downloaded appimages are safe? Going through the install scripts, it seems there is no security mechanism in place to validate the downloaded appimages.

For example Linux Mint had a security breach in 2016, and the ISO image hosted was hacked: [https://blog.linuxmint.com/?p=2994](https://blog.linuxmint.com/?p=2994). Luckily, some people checked the md5sums of their downloads and reported the issue.

# Possible Solution
The app maintainers usually provide an "official" md5/sha1/sha256/sha512 checksum on the download page.

Maybe we should should include the md5 checksums for appimages in our install scripts? I believe most Linux distros have md5sum by default, so we can easily add a step to do the verification.

But it will get complicated as there are so many versions and updates for each appimage. I don't exactly how we can solve this problem, but I think it is good to discuss and come up with a solution.

ivan-hc:
> I don't exactly how we can solve this problem

Same here.

> but I think it is good to discuss and come up with a solution.

I completely agree.

OK, let me explain my point.

AM is like the AUR; it has scripts that download portable applications and AppImages from the original sources. It doesn't physically store anything in the database.

The difference with PKGBUILDs is that there's no one to update them. The scripts always point to the latest version. You can use the same script for months or years and install whatever you want without any problems.

This choice is intentional to ensure the continuity of the project, even without me.

As you'll understand, the conditions for saving and tracking data that guarantee the security of installed apps are practically nonexistent.

The only two weapons we already have in place are:
- adding the word "spooky" to the installation scripts of apps that don't have a clear CI or source code available
- adding "unofficial" to the app description, in the list

two weapons that, given the amount of requests and the growing number of apps in the database, are difficult to track.

As suggested in the README, you can use `-a` to read the catalog pages and get the link from the terminal, or `-d` to get the script and track the app's origin.

After all, I manage **portable apps and AppImages**, which **are designed to be downloaded from their developer's website**. AM simply saves you the hassle of opening a browser and searching for the app. Simply type the app name after `-i` and it will install it.

Ultimately, it's the user's duty to be aware of what they're installing/using. They must trust the app they're installing to use it. AM merely acts as an intermediary between you and the developer. That's all.

If an app turns out to be malicious, simply report it and I'll remove the script.

shuvashish76:
[...]

PR #2111, opened 2026-02-28
Added support for checksum verification in install scripts.

vishnu350:
Added initial work for checksum verification support in template.am, based on our discussion in #2104.

Tested with pkgforge-dev/mednafen-appimage with checksum URL = $version.zsync. It should also work for download pages that publish the checksum directly on the page.

Maintainers can slowly enable checksum verification as they see fit, else they can just select "none". Checksum verification failure will only show a warning, prompting users to manually verify the binary. It will not break the installation flow.

### DIFF
--- a/modules/template.am
+++ b/modules/template.am
@@ -342,6 +342,22 @@ _template_then_git_repo() {
 	esac
 }
 
+_template_verify_checksum() {
+	echo "$DIVIDING_LINE"
+	read -r -p $" Verify appimage checksum integrity? (sha1sum, sha256sum, sha512sum, md5sum, none)? " chksumtool
+	case "$chksumtool" in
+	'sha1sum'|'sha256sum'|'sha512sum'|'md5sum')
+		echo "$DIVIDING_LINE"
+		read -r -p $" Provide checksum URL (eg. \$version.zsync or download page): " chksumlink
+		sed -i "s/CHKSUM_TOOL/$chksumtool/g" ./am-scripts/"$MAIN_ARCH"/"$arg"
+		sed -i "s/CHKSUM_URL/$chksumlink/g" ./am-scripts/"$MAIN_ARCH"/"$arg"
+		;;
+	*)
+		sed -i "/checksum/d" ./am-scripts/"$MAIN_ARCH"/"$arg"
+		;;
+	esac
+}
+
 # ------------------------------------------------------------------------
 # FUNCTIONS SPECIFIC PER WEBSITE
 # ------------------------------------------------------------------------
@@ -483,6 +499,7 @@ for arg in $ARGS; do
 				fi
 			esac
 			_template_test_github_alt_architectures_url
+			_template_verify_checksum
 			# END OF THIS FUNCTION
 			echo "$DIVIDING_LINE"
 			echo -e $"\n All files are saved in $SCRIPTDIR/am-scripts\n"
--- a/templates/AM-SAMPLE-AppImage
+++ b/templates/AM-SAMPLE-AppImage
@@ -22,6 +22,9 @@ mv ./tmp/*mage ./"$APP"
 rm -R -f ./tmp || exit 1
 echo "$version" > ./version
 chmod a+x ./"$APP" || exit 1
+checksum=`CHKSUM_TOOL ./"$APP" | awk '{ print $1 }'`
+results=`curl -Ls "CHKSUM_URL" | grep -a $checksum`
+[ "$results" = "" ] && echo "Warning: Appimage checksum verification failed, please manually verify file integrity"
 
 # LINK TO PATH
 ln -s "/opt/$APP/$APP" "/usr/local/bin/$APP"
@@ -48,6 +51,9 @@ if [ "$version" != "$version0" ]; then
 	mv --backup=t ./tmp/*mage ./"$APP"
 	chmod a+x ./"$APP" || exit 1
 	echo "$version" > ./version
+	checksum=`CHKSUM_TOOL ./"$APP" | awk '{ print $1 }'`
+	results=`curl -Ls "CHKSUM_URL" | grep -a $checksum`
+	[ "$results" = "" ] && echo "Warning: Appimage checksum verification failed, please manually verify file integrity"
 	rm -R -f ./*zs-old ./*.part ./tmp ./*~
 	notify-send "$APP is updated!"
 else


ivan-hc:
I have tested your changes on two scripts of appimages from my repositories. Anydesk and LXtask.

This is the result

Istantanea_2026-02-28_18-16-40

ivan-hc:
According with your changes, it seems that an checksum tool must be added at the moment of creating the script with `-t`.

It is not so that should works.

$chksumtool may be vary on other systems. You can't set one if you know that other systems may have different tools.

ivan-hc:
Also, your approach uses `curl` twice. We should try to reduce `curl` calls to limit internet calls or too many calls to api.github.com or against the proxy in use.

ivan-hc:
In brief:
1. your approach uses a checksum tool only when the script is created, so all scripts already available have not this approach, and all of them should be changed
2. you rely on the tool in use on your PC, but people with different tools may have a error about "command not found"
3. when installing you use `curl` two times, the first time to check the $version, and the second time for checksum. Same happens when updating an app, and this doubles the traffic against api.github.com (that may cause errors) or against the proxy (that risks to be overloaded)

ivan-hc:
4. while updating an AppImage, if `appimageupdatetool` is installed and delta updates are available, no `wget` part will be involved (and then also your function will not be used)

ivan-hc:
NOTE: template.am is only needed to create scripts.

Templates are what you need to change for your task.

Or at least, install.am (for `-i`) and APP-MANAGER itself (for `-u`) if you want to extend the check to all the apps without changing the hundreds of scripts one by one. This would be the best approach, in my opinion.

vishnu350:
Hi, thanks for the thorough review. Sorry for the lack of polish, I wanted to see if you like the idea first.
1. We can change the checksum tool to be a variable that that is system dependent.
2. We can also check if the tool exists, and skip the verification flow if the user does not have the checksum tools available.
3. I'm not sure how we can improve the double curl traffic. Maybe for devs, we can add a switch to disable verification during testing?
4. Ill need to study this a little bit more, not familiar with `appimageupdatetool`.

> Or at least, install.am (for -i) and APP-MANAGER itself (for -u) if you want to extend the check to all the apps without changing the hundreds of scripts one by one. This would be the best approach, in my opinion.

OK, let me look into this approach. The issue is the checksum maybe placed in different URL links, so we need a way to store the correct URL link or disable it completely for apps that dont publish the checksum. Eg. for GIMP it is in https://www.gimp.org/downloads, but for mednafen it is https://github.com/pkgforge-dev/mednafen-appimage/releases/download/1.32.1%402026-02-15_1771141826/mednafen-1.32.1-anylinux-x86_64.AppImage.zsync

ivan-hc:
@vishnu350

I remind you that checksums are not human-readable. Only special programs can verify them. And checksums exist for both correct and incorrect programs.

The correct checksum data is provided upstream, by the app developer.

The only way to make them "verifiable" is through tracking in dedicated scripts and repositories, as is done in other projects whose application databases are "physical" and, above all, "tracked."

We simply download apps from sources, without authorization or direct involvement from the authors, in most cases.

In the AUR PKGBUILDs, there are people who write the script and list all its data, including version and checksum. And they can do this late, causing updates to mismatch with reality, resulting in delays and neglect in the maintenance of these scripts. It happens in the AUR, as well as in Flatpak and even Arch Linux itself (today someone filed an issue in my VLC-appimage repository, telling me that my VLC is two versions older than upstream... which is normal, given that the package is based on the VLC in Arch Linux, which is outdated).

AM's approach is more direct; it doesn't wait for updates in the scripts. Each update is real-time, precisely because the scripts aim to download the latest available artifact.

In any case, we'll never solve the problem if we don't adapt to the standards adopted in Flatpak, AUR, or Soar, which wait to update the data of individual apps before making them available for download.

---------------------------

In summary, both real and fake apps have a checksum. This is provided by upstream. Since we only provide scripts that download them directly, any attempt would be futile, if not downright harmful to the goal of the AM project, which is to provide the latest artifact published upstream in real time.

To identify problematic apps, we need user feedback or the workflows already active in amcheck that test their installability every day. Based on this, we can decide whether to modify or delete a script, which is what I've always done.

It's not easy to control what doesn't belong to us. And portable apps don't belong to us. We just download them.

If we were APT, we'd have our own DEBs; if we were DNF, we'd have our own RPMs; if we were Flatpak, we'd have our own Flatpaks. It would take a few hours or days to be in sync with upstream. But we deal with portable apps from upstream. We're in sync with upstream. And this penalizes tracking. The security of these apps is based on the user's trust in the upstream developer.

Again, if you open the developer's page and download the app or use AM to download and install it, the situation doesn't change. But with AM, you save the time of opening the browser, searching for the source, and clicking "download."

vishnu350:
> ... any attempt would be futile, if not downright harmful to the goal of the AM project, which is to provide the latest artifact published upstream in real time.

The proposed flow will not interrupt the user experience or cause any harm. It will only show a warning message, prompting the user to manually investigate. The app will still install normally and function as expected.

I disagree that it is futile to verify checksums, otherwise app devs will not publish them on their download pages. Anyway, I will just close the pull request and issue.

ivan-hc:
@vishnu350 I'm not saying it's useless, I'm just saying that you can't apply this to AM.

We should track and perhaps even archive all the programs we manage in the package manager to make a truly effective checksum.

Pick up any PKGBUILD from the AUR; you'll notice that the checksum is there because someone put it there. PKGBUILDs are always updated manually (at the cost of delaying app downloads by hours, days, weeks, months... years). It's pure bureaucracy, certainly necessary for Arch Linux's security.

But what we do with portable apps is just downloading them. There are still a lot of people... or rather, too many people who continue to rely on appimagehub.com, despite it being a site full of fake and potentially malicious apps. They certainly have a checksum, but it's useless in that case. I recently downloaded Firefox from there to see what was inside (since it's not uncommon for people who upload files there to steal them from other repositories and earn donations, as happened with my GIMP in the past), but when I extracted it, it was a completely different app from Firefox.


https://github.com/user-attachments/assets/11971730-f2e4-48d4-a866-74e8606e2331

See? appimagehub.com is a FAKE and CRIMINAL site! Yet people still use it.

At least I have scripts with clearly visible and searchable URLs and sources to track, so people can be aware of what they download, install, and update via AM.

vishnu350:
> I'm just saying that you can't apply this to AM.

Could you please properly try the proposed solution? Just do "am -t mednafen" (repo URL: https://github.com/pkgforge-dev/mednafen-appimage). When it asks for verification choose "sha1sum", and the set the checksum URL to "$version.zsync". Then install using "am -i <install_script>". I am also on Debian, so it should work for you as well.

The principle is very simple:
1. Calculate the appimage file checksum (after the install script downloads the appimage).
2. Compare the checksum vs the published checksum (can be zsync, digest, or directly on download page).
3. If not matching, print a warning. That is all.

For verification, selecting "none" will remove all checksum code from the install script, so it is completely optional.

> Pick up any PKGBUILD from the AUR; you'll notice that the checksum is there because someone put it there.

Yes, I am aware of this. But there is absolutely no need for this at all in AM. This is not what I am proposing.

Maybe you misunderstand why binary verification is important. It is not just hackers/fake sources that pose a problem, sometimes large binary downloads can get corrupted due to system issues, poor network conditions, DNS problems, host server issues, bit-rot, etc.

> At least I have scripts with clearly visible and searchable URLs and sources to track, so people can be aware of what they download, install, and update via AM.

I think the existing AM scripts are awesome, and AM is definitely the safest source for appimages at this point (thanks to you and the others efforts).

All I am proposing is the (optional) automation of checksum verification that maintainers can slowly add into the install scripts as time goes on, which will make AM even better.

ivan-hc:
> All I am proposing is the (optional) automation of checksum verification that maintainers can slowly add into the install scripts as time goes on, which will make AM even better.

You said the right word: "gradually."

The problem is that few of us also provide the .zsync file with the AppImages. It's true, between me and Pkgforge, we've produced many and distribute many... but there aren't as many of us as there are non-compliant ones.

Many simply package a generic AppImage, and many more continue to use old, obsolete, and insecure tools to package them. The fact that many still require libfuse2 to run (despite our warnings) is proof that we're marginal as a project. I had to create the `nolibfuse` option to at least regulate them in this respect.

The fact that I often have to modify these scripts because the site hosting a program changes (the last time yesterday, with "speed-dreams") would also frustrate any attempt to standardize certain types of checks. Different hosts require different standards. Some programs have such complex hosts that they require multiple `curl` calls just to identify the version, and just as many to identify the artifact to download.

I would encourage you to clone this repository locally and run this script from its folder

https://github.com/ivan-hc/AM/blob/main/tools/sources-selector.sh

...so you can get a clear picture of the different hosting types for various programs.

But to give you an idea, I'll list the data from the latest research I posted a few weeks ago on Discord:

codeberg=4
custom-aur=17
custom-multichoose=1
custom-standard=132
custom-verbose=47
ffwa=9
github=2635
github-custom=21
gitlab=11
kde=3
mozilla=11
sourceforge=30

The above is not updated, but believe me if I say that 200 apps from random sites (that are not github) are a big problem.

vishnu350:
> You said the right word: "gradually."

Yes. This is an optional feature with no urgency. Apps with zsync/digest files can be updated in their respective install scripts, whenever the maintainers get to it. But they cant even start if there is nothing provided in the templates on how to do this.

Having an **added security feature will be good for new users to trust AM/appimages** over huge 10GB+ flatpaks.

> The problem is that few of us also provide the .zsync file with the AppImages. It's true, between me and Pkgforge, we've produced many and distribute many... but there aren't as many of us as there are non-compliant ones.

This is not a problem. The existing install scripts can remain as it is for apps that don't provide zsync/digest files. As I said, **it is optional**. Over time this will change.

> The fact that I often have to modify these scripts because the site hosting a program changes (the last time yesterday, with "speed-dreams") would also frustrate any attempt to standardize certain types of checks.

Come on, how often will the variable $version.zsync or $version.digest require maintenance? Even if it fails, it will just print a WARNING and the install script will still run as usual. And the warning will certainly help in debugging any issues, maybe the file uploaded by the devs is corrupted or whatever.

> The above is not updated, but believe me if I say that 200 apps from random sites (that are not github) are a big problem.

Again, how is this a problem for an optional feature? Just leave verification disabled, there is no need to update the existing scripts for these. Maintainers can still try, eg websites like gimp.org do publish the checksum on download.html, my proposal will still work for this as it is just "curl URL | grep CHECKSUM".

> See? appimagehub.com is a FAKE and CRIMINAL site! Yet people still use it.

My personal belief is that this website was intentionally designed this way to make people dislike appimages. Lets not talk about it, and focus on how we can be better.

I have been downloading stuff online for decades, and always manually check the checksum (if possible) to avoid problems. There are times when the checksum does not match and I always re-download the file.

**My change is literally just 17 lines in total** (git diff) and only for the template generator. Only 3 lines are added into the install script template (compute checksum, compare, and print warning - 3 lines).

It is not polished, but I believe it is a good proof-of-concept and starting point. It is certainly less lines than the amount of arguments you are making over such a simple feature.

ivan-hc:
> **My change is literally just 17 lines in total** (git diff) and only for the template generator. Only 3 lines are added into the install script template (compute checksum, compare, and print warning - 3 lines).

@vishnu350 I don't want to stop you, but the template.am module has nothing to do with what you want to do. It simply downloads a template and modifies it. It has no validation purposes other than verifying that a program exists on api.github.com, codeberg.com, and sourceforge.org.

Think carefully.

You must apply this function to one of the following:
1. the template (preferably at the end of the script, so try not to change the order of the lines; they're needed for "-t" to replace lines)
2. the "-u" option, which should somehow extract the value of "$version" (you probably need `eval`)
3. the "-i" option, in the same way I explained for "-u" in point 2.

Now, points 2 and 3 are more "global" in that they are "dynamic." They would base their behavior on any script, even apps installed with the "-e" option. The simplest and most accurate way to do this is to target the AM-updater script (as the "downgrade" or "--rollback" options do).

Point 1, on the other hand, would be less rapid and specific for newly created scripts.

Determine what's best for you. I can only offer my opinion.

ivan-hc:
> > **My change is literally just 17 lines in total** (git diff) and only for the template generator. Only 3 lines are added into the install script template (compute checksum, compare, and print warning - 3 lines).
> 
> You must apply this function to one of the following:
> 
> 1. the template (preferably at the end of the script, so try not to change the order of the lines; they're needed for "-t" to replace lines)
> 
> 2. the "-u" option, which should somehow extract the value of "$version" (you probably need `eval`)
> 
> 3. the "-i" option, in the same way I explained for "-u" in point 2.
> 
> 
> Now, points 2 and 3 are more "global" in that they are "dynamic." They would base their behavior on any script, even apps installed with the "-e" option. The simplest and most accurate way to do this is to target the AM-updater script (as the "downgrade" or "--rollback" options do).

Going back to points 2 and 3 I suggested:
- changing only the "-i" option would only add the check during installation, limiting internet traffic.
- changing only the "-u" option would add the check at each update, increasing internet traffic.

Make your own considerations.

vishnu350:
Alright. I will work on changing it and will think carefully on how to do this based your suggestions. Thank you.

PR #2124, opened 2026-03-06
Added _use_verify() function into APP-MANAGER for checksum verification

vishnu350:
Hi @ivan-hc, this is a simple pull request to implement #2104, total of 48 lines. Here is the change summary:
- I have added _use_verify() function into APP-MANAGER, invoked via "am verify app" or "am -V app". It will compute the checksum of the app and then check the following sources for hashes, in sequence of availability:
  1. curl /opt/app/version.zsync (pkgforce repos always have these)
  2. curl /opt/app/version.DIGEST (some repos have these)
  3. curl /opt/app/verif (optional file containing a download page URL or similar, to be added by maintainers)

- Successful checksum verification will create /opt/app/AM-VERIFIED status file, which will result in a tick beside the version number in the generated table.
```
 - APPNAME      | VERSION          | TYPE           | SIZE
 ◆ gimp         | 3.0.8✓           | appimage       | 215 MiB
```

- Unsuccessful automatic verification will just prompt the user to manually verify the files, but a checksum mismatch will prompt the user to reinstall the app.

- If the shasum/md5sum tools are not available, AM will prompt the user to install them.

- I have updated the GIMP install script to add the download page URL into /opt/gimp/verif, as an example for how the verification function can still work when zsync/DIGEST files are not provided.
`echo "$DOWLOAD_PAGE" > ./verif`

- Updating or rolling back an app will remove the AM-VERIFIED file, if the app version is changed.

I have not yet added the use_verify function into -i or -u, because of the additional curl calls, and since it is still an experimental feature. It must be manually called for each app for now.

Interestingly, I actually did find one app in my database that failed verification (checksum mismatch). Updating it via am -u solved the problem, am verify shows successful verification.

I have not yet updated the documentation section.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1414,15 +1414,13 @@ _use_verify() {
 			printf "Source is tarball/zip" > "$argpath"/AM-VERIFIED
 			return 1
 		# Search for hash files
-		elif [ -f "$argpath"/version ] && [ ! -z $(grep -a "^http" "$argpath"/version) ]; then
+		elif [ -f "$argpath"/version ] && [ ! -z "$(grep -a "^http" "$argpath"/version)" ]; then
 			# Search for zsync file
 			digest=$(curl -Ls "$(sort "$argpath/version").zsync" | grep -a "SHA\|MD5")
 			# Search for DIGEST file if no zsync file
 			if [ -z "$digest" ]; then
 				digest=$(curl -Ls "$(sort "$argpath/version").DIGEST")
-				results=$(echo "$digest" | grep -ia "not found")
-				# TODO: Refactor this
-				if [ ! -z "$results" ]; then
+				if [ ! -z "$(echo "$digest" | grep -ia "not found")" ]; then
 					printf $" %b⚠️\033[0m Checksum for %b cannot be automatically verified, no hashes found \n" "${Gold}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
 					return 1
 				fi
@@ -1454,7 +1452,7 @@ _use_verify() {
 		[ -z "$checksum_tools_on" ] && echo "Checksum calculation tools do not exist, please install shasum/md5sum"
 		checksum_tools_on="1" # Do not repeat the message above
 	fi
-	unset checksum_sha1 checksum_sha256 checksum_sha512 checksum_md5 digest
+	unset checksum_sha1 checksum_sha256 checksum_sha512 checksum_md5 digest results
 }
 
 ################################################################################################################################################################


ivan-hc:
## 1. About `gimp` and other installation scripts

If all you have done in `gimp` is to add
```
echo "$DOWLOAD_PAGE" > ./verif
```
why don't you add a patch [here](https://github.com/ivan-hc/AM/blob/5f9fce63bfe9f7e4bac9d9af6c54c7a99bebcef9/modules/install.am#L171), in the install.am module?

It is enough to detect lines containing

```
echo "$version" > ./version
```
...and use `sed` to add your command to the script, like this
```
echo "$version" > ./version; echo "$DOWLOAD_PAGE" > ./verif
```
...or something. But I suppose that the more common patch would be
```
echo "$version" > ./version; echo "$version" > ./verif
```
...among all installation script we have and that you would use as an example, you have chosen one of the more "strange" (if compared to the others) :laughing: 

NOTE: maybe it is also not necessary at all. If all you need is a download URL, we already have the "version" file, that in most cases is a URL itself.

Take this piece in consideration.

The approach would be easier for you. We would only need to `wget` the URL inside the "version" file (if it exists) by adding the extension .zsync or .digest/.shasum/.whateveritis

-----------------------------------------

## 2. About the change in database.am

I hope this does not conflicts with the option `lock`

------------------------------------------

## 3. About APP-MANAGER

I still need to analyze your function. We need to know how it acts when there are many apps together.

Please work on point 1 above, so we can have more apps on  which we can work and test.

ivan-hc:
## My analysis on your function

```
_use_verify() {
	_online_check
	_determine_args
	## Check if checksum calc tools exist
	if (command -v shasum &> /dev/null) && (command -v md5sum &> /dev/null); then
```
Sorry for my ignorance, this means that we need two programs to use this? `shasum` and `md5sum`?

```
		ENTRIES="$(echo "$@" | cut -f2- -d ' ' | tr ' ' '\n' | grep -v -- "^-\|^$1$")"
		for arg in $ENTRIES; do
			argpath=$(echo "$ARGPATHS" | grep "/$arg$")
			[ ! -d "$argpath" ] && printf $" %b✖\033[0m Checksum for %b cannot be verified, program not found \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')" && continue
			checksum_sha1=`shasum -a 1 $argpath/$arg | awk '{ print $1 }'`
			checksum_sha256=`shasum -a 256 $argpath/$arg | awk '{ print $1 }'`
			checksum_sha512=`shasum -a 512 $argpath/$arg | awk '{ print $1 }'`
			checksum_md5=`md5sum $argpath/$arg | awk '{ print $1 }'`
```
LGTM
```
			## Search for zsync file
			version=`cat "$argpath/version"`.zsync
			digest=`curl -Ls $version | grep -a "SHA\|MD5"`
			## Search for DIGEST file if no zsync file
			if [ "$digest" = "" ]; then
				version=`cat "$argpath/version"`.DIGEST
				digest=`curl -Ls $version`
			fi
```
You need to verify if the content of "$argpath/version" is a download URL or a URL in general.

On my side I could edit all installation scripts that don't add a URL to the "version" file and made this a standard. I'll do my best.

NOTE, `[ "$digest" = "" ]` should be `[ -z "$digest" ]` if what you are thying to do is to verify if the variable is empty.

```
			## Last resort: Search download page or manual URL if verif file exists, else continue to next app
			if [ ! "`echo $digest | grep -i "not found"`" = "" ]; then
```
I have not understand this command. Are you thying to see if `$digest` is empty? Use `[ -z "$digest" ]`.

If instead you are trying to see if "not found" is not in that output, just do `if ! echo "$digest" | grep -qi "not found"; then`

Another thing I noticed is that you use often the
```
 "`"
```
that is a symbols that executes commands. I think it is safer to use variables instead. But its just my opinion.

```
				[ -f "$argpath/verif" ] && version=`cat "$argpath/verif"` && digest=`curl -Ls $version`
				[ ! -f "$argpath/verif" ] && printf $" %b✖\033[0m Checksum for %b cannot be automatically verified, no hashes found \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')" && continue
			fi
```
You would use an `if/fi` statement with an `else` here, since you are checking if a file is available or not in that path.

```
			## Grep for any matching checksum
			results=`echo $digest | grep -a "$checksum_md5\|$checksum_sha1\|$checksum_sha256\|$checksum_sha512"`
			if [ "$results" = "" ]; then
```
Again, use `[ - z "$results" ]`
```
				printf $" %b✖\033[0m Checksum for %b does not match, please manually verify file integrity or reinstall \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
			else
				printf $" %b✔\033[0m Checksum for %b matches, verification successful! \n" "${Green}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
				printf "MD5: $checksum_md5\nSHA1: $checksum_sha1\nSHA256: $checksum_sha256\nSHA512: $checksum_sha512" > $argpath/AM-VERIFIED
			fi
		done
	else
		echo "Checksum calculation tools do not exist, please install shasum/md5sum"
	fi
	wait
}
```
LGTM

-----------------------------

## Conclusions

I like your approach. Well done.

The only thing you need to do is refactor a bit. Not that the commands are bad, but you might want to simplify your function and make it smaller and less complex using the right commands. "Shellcheck" doesn't give errors after all; it's just a matter of style, but everyone has their own tastes. If that's your thing, that's fine by me.

I recommend working on using only the "version" file for each app. This way, you'd only be able to modify the function in APP-MANAGER.

The function that lists the versions is also in APP-MANAGER, and you probably don't even need to modify database.am if you want to show the check mark next to the version in `am -f`.

To simplify, **you might want to modify only APP-MANAGER in this pull request, given the circumstances. You don't need to modify anything else.**

We're trying to work on this for now. We'll work on the documentation later.

If you have any questions, let me know.

ivan-hc:
@vishnu350 you know? Instead of an option, this would become part of `am -s` (that is also called by `am -u` itself).

ivan-hc:
@vishnu350 to prevent touching database.am, you can add your check between these two lines

https://github.com/ivan-hc/AM/blob/03f7cdd38cf6079a44b66cc03d2650b21584ee7f/APP-MANAGER#L844-L845

```
if [ -f "$argpath"/AM-VERIFIED ]; then
	APPVERSION="$APPVERSION✓"
fi
```

ivan-hc:
## Schema: Here's how it should work
We'll put your function in `-u`. After the update has been performed (with or without new versions), we'll verify the checksum:
- If it exists, we create the "AM-VERIFIED" file in the app folder.
- If it doesn't exist, we delete the "AM-VERIFIED" file, if it exists.

The same thing will be done in `-i`, at the end of the installation.

In APP-MANAGER, once the version is set, we'll add a check mark to the value of $APPVERSION, as I mentioned before. This way, every time we run an option that displays the version number (`-f`, `-l`, and `-a`, or even running `-u`), we'll see the check mark indicating that the app is verified for that version.

--------------------------

## Summary
1. Create the function
2. Ensure that the function is used in `-u` (APP-MANAGER) and `-i` (install.am)

You don't need to change anything else.

ivan-hc:
NOTE: This means you don't need to do this for all apps at once. Your function must be used on a single app.

You don't need this anymore :point_down: 

		ENTRIES="$(echo "$@" | cut -f2- -d ' ' | tr ' ' '\n' | grep -v -- "^-\|^$1$")"
		for arg in $ENTRIES; do

...nor this :point_down: 

	_online_check
	_determine_args

All the above is already included in `-i` and `-u`.

As it stands, your function would change like this
```
_use_verify() {
	## Check if checksum calc tools exist
	if (command -v shasum &> /dev/null) && (command -v md5sum &> /dev/null); then
		[ ! -d "$argpath" ] && printf $" %b✖\033[0m Checksum for %b cannot be verified, program not found \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')" && continue
		checksum_sha1=`shasum -a 1 $argpath/$arg | awk '{ print $1 }'`
		checksum_sha256=`shasum -a 256 $argpath/$arg | awk '{ print $1 }'`
		checksum_sha512=`shasum -a 512 $argpath/$arg | awk '{ print $1 }'`
		checksum_md5=`md5sum $argpath/$arg | awk '{ print $1 }'`
		## Search for zsync file
		version=`cat "$argpath/version"`.zsync
		digest=`curl -Ls $version | grep -a "SHA\|MD5"`
		## Search for DIGEST file if no zsync file
		if [ "$digest" = "" ]; then
			version=`cat "$argpath/version"`.DIGEST
			digest=`curl -Ls $version`
		fi
		## Last resort: Search download page or manual URL if verif file exists, else continue to next app
		if [ ! "`echo $digest | grep -i "not found"`" = "" ]; then
			[ -f "$argpath/verif" ] && version=`cat "$argpath/verif"` && digest=`curl -Ls $version`
			[ ! -f "$argpath/verif" ] && printf $" %b✖\033[0m Checksum for %b cannot be automatically verified, no hashes found \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')" && continue
		fi
		## Grep for any matching checksum
		results=`echo $digest | grep -a "$checksum_md5\|$checksum_sha1\|$checksum_sha256\|$checksum_sha512"`
		if [ "$results" = "" ]; then
			printf $" %b✖\033[0m Checksum for %b does not match, please manually verify file integrity or reinstall \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
		else
			printf $" %b✔\033[0m Checksum for %b matches, verification successful! \n" "${Green}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
			printf "MD5: $checksum_md5\nSHA1: $checksum_sha1\nSHA256: $checksum_sha256\nSHA512: $checksum_sha512" > $argpath/AM-VERIFIED
		fi
	else
		echo "Checksum calculation tools do not exist, please install shasum/md5sum"
	fi
}
```

ivan-hc:
@vishnu350 I have tried to simplify your function ever more, and I added some checks to made it more robust and secure.

This is my two cents

```
_use_verify() {
	# Check if checksum calc tools exist
	if command -v shasum &> /dev/null && command -v md5sum &> /dev/null; then
		if [ -f "$argpath"/"$arg" ]; then
			checksum_sha1=$(shasum -a 1 "$argpath/$arg" | awk '{print $1}')
			checksum_sha256=$(shasum -a 256 "$argpath/$arg" | awk '{print $1}')
			checksum_sha512=$(shasum -a 512 "$argpath/$arg" | awk '{print $1}')
			checksum_md5=$(md5sum "$argpath/$arg" | awk '{print $1}')
		fi
		if [ -f "$argpath"/version ] && grep -q "^http" "$argpath"/version; then
			# Search for zsync file
			digest=$(curl -Ls "$(sort "$argpath/version").zsync" | grep -a "SHA\|MD5")
			# Search for DIGEST file if no zsync file
			if [ -z "$digest" ]; then
				digest=$(curl -Ls "$(sort "$argpath/version").DIGEST")
			fi
		else
			printf $" %b✖\033[0m Checksum for %b cannot be automatically verified, no hashes found \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
			return 1
		fi
		# Grep for any matching checksum
		results=$(echo "$digest" | grep -a "$checksum_md5\|$checksum_sha1\|$checksum_sha256\|$checksum_sha512")
		if [ -z "$results" ]; then
			printf $" %b✖\033[0m Checksum for %b does not match, please manually verify file integrity or reinstall \n" "${RED}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
		else
			printf $" %b✔\033[0m Checksum for %b matches, verification successful! \n" "${Green}" "$(echo "$arg" | tr '[:lower:]' '[:upper:]')"
			printf "MD5: %b\nSHA1: %b\nSHA256: %b\nSHA512: %b" "$checksum_sha1" "$checksum_sha256" "$checksum_sha512" "$checksum_md5" > "$argpath"/AM-VERIFIED
		fi
	else
		[ -z "$checksum_tools_on" ] && echo "Checksum calculation tools do not exist, please install shasum/md5sum"
		checksum_tools_on="1" # Do not repeat the message above
	fi
	unset checksum_sha1 checksum_sha256 checksum_sha512 checksum_md5 digest
}
```

### What is changed
1. since we are not going to use a "verif" file because it is the same as "version", I removed that part because it was duplicated
2. the check for missing dependencies should show the message only one time
3. no "$version" variable, it is used directly in "$digest"
4. `continue` is not suitable in functions, I replaced it with `return 1`
5. we detect if the "version" file exists and if it contains a URL
6. set variables and unset them at the end
7. a general code refactoring


Now all we need to do is to test this.

vishnu350:
Hi @ivan-hc,

> ensure that the function is used in -u (APP-MANAGER) and -i (install.am)

This will increase the curl traffic, +2 per app. Shouldn't we test it a little more first?

> [ "$digest" = "" ] should be [ -z "$digest" ] ...

I will fix this, as well as the other refactoring work you mentioned in your function analysis.

> this means that we need two programs to use this? shasum and md5sum?

Yes, these are usually available by default in most Linux/Unix distros by default. Basically we just compute all four of the most common type of checksums (SHA1, SHA256, SHA512, and MD5). Then we grep the URL sources and look for any match.

> why don't you add a patch [here](https://github.com/ivan-hc/AM/blob/5f9fce63bfe9f7e4bac9d9af6c54c7a99bebcef9/modules/install.am#L171), in the install.am module? It is enough to detect lines containing ...and use sed to add your command to the script, like this ...or something.

I tried many ways and found that it is not possible. Because the custom URL may not be a download page, but can be a custom link. For GIMP, this is https://www.gimp.org/downloads (HTML page). But for Krita it will be $DOWNLOAD_URL/**Krita-Plus-x86_64.appimage.zsync**, while app itself is $DOWNLOAD_URL/krita-5.2.16-prealpha-07989fedfa-x86_64.AppImage, because it is fun to have a different name for your zsync file.

So I thought for us, the install script maintainers have to put that info in the verif file for the use_verify function to pick up.

> I hope this does not conflicts with the option lock

In database.am the AM-LOCK check is like this (line 343):
`[ -f ./"$arg"/AM-LOCK ] && APPVERSION="$APPVERSION🔒"`

So it made sense to me to just add the AM-VERIFIED check in the next line:
`[ -f ./"$arg"/AM-VERIFIED ] && APPVERSION="$APPVERSION✓"`

It just adds it cumulatively, so it looks like this if lock is enabled:
```
 - APPNAME      | VERSION          | TYPE           | SIZE
 ◆ nvim         | 0.11.6🔒✓        | appimage       | 11 MiB
```

ivan-hc:
> Now all we need to do is to test this.

I have changed this function to update a single app (the function that runs the AM-updater)
```
_update_run_updater() {
	if grep -q "api.github.com" "$argpath"/AM-updater; then
		if [ "$GH_API_ALLOWED" -le 5 ]; then
			sed -i "s#https://api.github.com#$ALT_GH#g" "$argpath"/AM-updater
		fi
	fi
	if [ -z "$debug_update" ]; then
		"$argpath"/AM-updater >/dev/null 2>&1
	else
		"$argpath"/AM-updater
	fi
	sed -i "s#$ALT_GH#https://api.github.com#g" "$argpath"/AM-updater 2>/dev/null
	end=$(date +%s)
	timelapsed=$((end - start))
	[ "$timelapsed" = 1 ] && printf $" %b✔\033[0m $APPNAME is updated, $timelapsed second elapsed! \n" "${Green}" || printf $" %b✔\033[0m $APPNAME is updated, $timelapsed seconds elapsed! \n" "${Green}"
	_use_verify
	[ -n "$debug_update" ] && echo "$DIVIDING_LINE"
}
```

...this is the result

https://github.com/user-attachments/assets/12f003fb-4d07-49d8-8bef-60220d4a91f9

This will look great when updating everything!

vishnu350:
> Now all we need to do is to test this.

😆 you've already refactored it. Commit your changes on top of mine so I can help to test? They look good to me.

ivan-hc:
@vishnu350  in install.am,  have changed the header of the post installation function like this
```
_post_installation_processes() {
	LASTDIR=$(ls -td "$APPSPATH"/* | head -1 | sed 's:.*/::')
	LASTDIRPATH="${APPSPATH}/${LASTDIR}"
	argpath="$LASTDIRPATH"
	_use_verify
	echo ""
```
this is the result

https://github.com/user-attachments/assets/7d4e1328-c1c3-409b-a9c5-e22e3c465308

ivan-hc:
The last thing to do is this

https://github.com/ivan-hc/AM/pull/2124#issuecomment-4013692678

ivan-hc:
> > ensure that the function is used in -u (APP-MANAGER) and -i (install.am)
> 
> This will increase the curl traffic, +2 per app. Shouldn't we test it a little more first?

Nope, when we update or install, we are going to touch api.github.com, and this has a limit.

Once that from that site we print the download URL to the "version" file, we can use `curl` on the download URL directly, without touching the APIs. there is no limit on this.

ivan-hc:
@vishnu350 its 4:53 a.m. here in Italy, I'm going to hit the bed.

Please, in the meantime, restore your branch and add the function I have changed where you have put yours.

Only APP-MANAGER.

Thanks to you I have learned something new today.

I will merge your changes to the dev branch to refine it and prepare the new AM version 9.9.5 later.

vishnu350:
Hi @ivan-hc, I have finished merging your fixes.
1. I have undone the changes to the gimp install script and database.am, and have added in all the changes you have described in your posts. Please check.
2. Verification: I have verified and refactored the function as much as I can.
3. Optimization: I have moved the checksum calculation part to after searching for the digest files, since the app may not even have them and will call `return 1`. Checksum calculation may take a while for large binaries on slow computers, and if it fails finding a hash file online we just wasted the user's time. Better to do it after.
4. Optimization: I have added a part to determine if the source download was a zip/gzip/tarball, as these have builtin checksum calculation for unzipping. The unzip process will fail if the checksum is wrong, so the verification is automatic. So, some "dynamic-binary" apps that are distributed as zip files can be considered auto-verified (eg. Blender).
5. I have improved the messages to show as warnings (gold color), so we do not cause panic for the user. Checksum mismatch will still show up as red since it is a legitimate error, success will be green.

The only thing I didnt add was this part, not sure how to do this:
```
_post_installation_processes() {
	LASTDIR=$(ls -td "$APPSPATH"/* | head -1 | sed 's:.*/::')
	LASTDIRPATH="${APPSPATH}/${LASTDIR}"
	argpath="$LASTDIRPATH"
	_use_verify
	echo ""
```

ivan-hc:
I'm going to merge this to the "dev" branch. To use it run
```
am --devmode-enable
am -s
```

PS: I will remove the option
```
	'verify'|'-V')
		_use_verify "$@"
		;;
```
...since `-u` will take care of it instead.

PR #2125, opened 2026-03-07
"AM" 9.9.5 - add checksum verification

ivan-hc:
continuation of https://github.com/ivan-hc/AM/pull/2124

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="9.9.4-20"
+AMVERSION="9.9.5"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"


ivan-hc:
Test update UI

https://github.com/user-attachments/assets/0a086ca1-3353-4018-92af-fc91fcb3b318

ivan-hc:
Test check marks in `-f` and `-l` (version number)

Istantanea_2026-03-08_03-13-35

ivan-hc:
Test installation UI

https://github.com/user-attachments/assets/03f507ad-abd3-49aa-b0eb-aeccf6c98fcd

ivan-hc:
@vishnu350 In your code, I changed this:
- I shortened the messages without naming the apps, which are already named in the options that use the function (see the videos above)
- I extended the integrity check to apps from third-party databases (see my gtk3 apps, in `-f`)
- I extended the check to archives by directly checking the `wget` command that downloads them, if the "version" file doesn't contain a URL, and I used AM-updater in this case.

--------------------------------------

## What I'm not convinced about

As you can see from my tests, three fairly popular and official programs are installed on my system: `appimagetool`, `kdenlive`, and `qdiskinfo`.

There's no .zsync file or other files for integrity checking in their repositories, that's true. But these are still three programs that are more or less popular in their own way. They're not mainstream apps published by a random amateur like me.

I'm speaking as a complete ignoramus in the field of integrity checks of this kind: isn't this damaging to their reputation?

Do we need other ways to verify integrity?

vishnu350:
You integrated it and made it look so seamless. Well done!

> isn't this damaging to their reputation?

It just means we don't have a way to validate it (yet), because they usually do publish the checksums. Double checking the checksum will actually help their reputation in the long run.

Lets say I have an appimage that is 4 bytes long, which is `[0x1, 0x2, 0x3, 0x4]`. A simple checksum is **1+2+3+4=10**.

Then someone downloads it, but for whatever reason in his computer the bytes became `[0x1, 0x2, 0x3, *0x5*]`. A simple checksum is **1+2+3+5=11**. **11** is different for **10**, so it is a checksum mismatch, meaning the binary is corrupted. Corrupt binaries can be due to many reasons, sometimes even a dev publishing mistake or software/hardware/network issues during download/upload.

Unfortunately, **Appimages will most likely still run even though there is a mistake in the binary** (try it for yourself with a hexeditor and poison a single byte). Somewhere along the way it might crash, but the user will never know what went wrong. If its a popular Youtuber or a noisy user, their opinion can negatively affect the project.

> Do we need other ways to verify integrity?

Yes, these apps usually do publish the checksums somewhere, just not as zsync or digest files.

For kdenlive, it is $version.mirrorlist. Maybe we can put a second line in the version file containing the checksum url (like my original verif file), and then we add the curl call in use_verify? We do this only in the install scripts for such apps, the rest can remain a single line.

Alternatively, for github sometimes the SHA256 is provided by default, we need to use github API to query it (not sure how to do this but Im sure it is possible).
Screenshot From 2026-03-08 13-22-18

vishnu350:
For `appimagetool`, doing this gives us a response with a greppable SHA256:
`curl -s -H "Accept: application/vnd.github+json" "https://api.github.com/repos/AppImage/appimagetool/releases" | grep digest`

We can use this as a last resort option since it uses an API call, and only if zsync/digest files are not available. Only works for github repos, and it is not ideal because devs may have made a mistake during upload process (eg. uploaded incorrect binary by accident).

3rd party downloads like kdenlive will still be an issue. We will have no choice but to insert a 2nd link into the version file that is the URL for the checksum. This can be a follow up pull request that I can work on as an improvement after we finish with this commit, let me think on it and propose something.

ivan-hc:
@vishnu350

GitHub API calls are something we should avoid as much as possible.

For this and other applications, this will encourage other developers to bring their AppImages up to modern standards, including using newer tools to remove libfuse2 and creating .zsync files in the same path as their AppImages.

What we've done here is a lot—even too much, I'd say.

Let's get others to follow suit for once.

ivan-hc:
@vishnu350

There's another thing that bothers me. We've considered files published with AppImages.

But what about programs available as dynamic/static binaries and scripts (for example, my `appimagen` and `archimage-cli`)?

Does the same rule apply to them as archives?

ivan-hc:
> @vishnu350
> 
> There's another thing that bothers me. We've considered files published with AppImages.
> 
> But what about programs available as dynamic/static binaries and scripts (for example, my `appimagen` and `archimage-cli`)?
> 
> Does the same rule apply to them as archives?

Same thing, some Appimages are built on-the-fly like AUR does with several programs

https://github.com/ivan-hc/AM/tree/main/appimage-bulder-scripts

vishnu350:
> What we've done here is a lot—even too much, I'd say.

I cant believe how nicely integrated the feature is, its awesome. I didn't expect you to plug it in so quickly, I thought maybe just use it for apps individually first and see how it goes.

But I think your decision is 100% the right direction.

> Let's get others to follow suit for once.

Right. Let your userbase chase the devs to follow the proper zsync standard (like libfuse). Probono specifically recommends it for appimages, its odd that his own tool doesn't publish it.

I honestly think the zsync spec is the future for app update flows, and is a major feature of appimages over snaps/flatpaks/etc. Downloading a 1MB zsync patch to update a 1000MB appimage sounds like magic. We have verification now, so this can be considered in the future.

ivan-hc:
> Probono specifically recommends it for appimages, its odd that his own tool doesn't publish it.

:rofl:

vishnu350:
> But what about programs available as dynamic/static binaries and scripts (for example, my appimagen and archimage-cli)? Does the same rule apply to them as archives?

Unfortunately only dynamic/static binaries that are distributed as archives like gzip/tar/xz/zip are OK. Because they have an extra step in the install process, whereby we unzip it. This unzip process will deal with the checksum, so we that's why we don't need to verify it again.

Raw executable binaries/scripts do not require unzipping, so it is considered not integrity verified by default.

ivan-hc:
> Raw executable binaries/scripts do not require unzipping, so it is considered not integrity verified by default.

This means two things:
1. Encourage users to rely solely on AppImage and load only AppImages in AM.
2. Exempt such programs from this check (and notify them with an appropriate message to normalize this, preferably without causing alarm).

vishnu350:
For you own tools, maybe just add a DIGEST file in the release section? You can add it in your GitHub release action, and then never need to think about it anymore.
`sha256sum $release_file > $release_file.DIGEST`

ivan-hc:
@vishnu350 I could do something else instead: since portable apps (not AppImages) are a minority, I could add a message "Checksum not supported for portable apps," or something similar.

It's enough to check whether the installed program is an AppImage or not.
```
if ! head -c10 "$argpath/$arg" 2>/dev/null | grep -qa '^.ELF....AI$'; then
```

ivan-hc:
> For you own tools, maybe just add a DIGEST file in the release section? You can add it in your GitHub release action, and then never need to think about it anymore. `sha256sum $release_file > $release_file.DIGEST`

I have added this custom action in Thunar
```
sha256sum %n > %n.DIGEST
```
I have tested the resulting files on `appimagen` and `archimage-cli`: the first works, the second don't.

Istantanea_2026-03-08_19-05-06

------------------

EDIT: it seems that only works when created on the installed program, not the one downloaded from releases.

I have compared the two files and are different.

vishnu350:
> I could do something else instead: since portable apps (not AppImages) are a minority, I could add a message "Checksum not supported for portable apps," or something similar.

Yeah, or we could just make the warning less scary like "Checksum verification not yet supported for this app", instead of "no hashes found". Or maybe even remove it altogether, since we can see the verified marks in am -f anyway. We can also wait for user feedback.

ivan-hc:
I think that is ready to be merged. I will do it in the next hours.

ivan-hc:
done

PR #2131, opened 2026-03-09
Use coreutils checksum tools instead of shasum

ivan-hc:
ref https://github.com/ivan-hc/AM/issues/2104#issuecomment-4022954246


vishnu350:
Tested on my side, works for me.

PR #2136, opened 2026-03-11
more 6 emulators and games

Link4Electronics:
Made a small fix to an issue that I forgot to do for enhanced appimages.


ivan-hc:
Why there were so many "APP=" entries wrong in the AM-updater scripts?

ivan-hc:
Anyway, thank you also for putting the entries on top of the lists. I can read better if the number of the scripts and the entries of the list are the same number.

PR #2149, opened 2026-03-14
Improve checksum verification

ivan-hc:
The first commit is "Checksum verification: check for sha256 file", that provides verified flag on some servers that are not github.com. I have tested this with `kdenlive` and works.

@vishnu350 take a look here.

I'm trying to solve the issue also for apps that are hosted on github.com

I won't deny that the new feature has caused quite a bit of alarm, as in the last five days I've been contacted by concerned people who hadn't read the release notes.

Many valid mainstream apps are marked as "unverified".

As I explained in the release notes, FAILURE TO VERIFY THE CHECKSUM DOESN'T MEAN THE APP IS INVALID. However, it's worth improving this feature to normalize the situation for less attentive users.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1429,6 +1429,12 @@ _use_verify() {
 		if [ -n "$download_url" ]; then
 			# Search for zsync file
 			digest=$(curl -Ls "$download_url.zsync" | grep -a "SHA\|MD5")
+			# Search for sha256 file
+			if ! grep -q "https://github.com/" "$argpath"/version; then
+				if [ -z "$digest" ]; then
+					digest=$(curl -Ls "$download_url.sha256")
+				fi
+			fi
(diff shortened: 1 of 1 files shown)


ivan-hc:
This commit has a regression. Many apps are displaying this message:

" ✖ Checksum does not match, please manually verify file integrity."

This situation needs to be studied further. I'm closing this PR.

vishnu350:
Hi @ivan-hc, I agree. The original function was built to be a standalone function, so I added warning/errors prompts. But your integration into the update flow makes it always prompt the warnings, so it will naturally alarm users.

Since we can see the ✓ to know the verification status, should we hide the warnings? Only show errors. Bash tools usually have a verbose switch.

Hang on, let me help you with this commit. I'll make a pull request.

PR #2153, opened 2026-03-15
Source digest, DIGEST, sha512/256/1 and md5 for checksum

ivan-hc:
This is an idea of @fiftydinar 

If `zsync` file is not hosted near the app, check the following file extensions, in this order:

1. digest
2. DIGEST
3. sha512
4. sha256
5. sha1
6. md5

@vishnu350 take a look at this change.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1439,16 +1439,6 @@ _use_verify() {
 				done
 				if echo "$digest" | grep -qi "not found"; then
 					return 1
-					## Disable these warnings?
-					#if head -c10 "$argpath"/"$arg" 2>/dev/null | grep -qa '^.ELF....AI$' || head -c10 "$argpath"/"$pure_arg" 2>/dev/null | grep -qa '^.ELF....AI$'; then
-					#	checksum_msg=$(echo $"Checksum cannot be verified, no hashes found.")
-					#	printf "%b⚠️\033[0m %b \n" "${Gold}" "$checksum_msg" | _fit
-					#	return 1
-					#else
-					#	checksum_msg=$(echo $"Checksum verification not yet supported for this program.")
-					#	printf "%b?\033[0m %b \n" "${LightBlue}" "$checksum_msg" | _fit
-					#	return 1
-					#fi
 				fi
 			fi
 		else


vishnu350:
That is a brilliant refactor. It will work.

But this will result in many curl calls per update, is that OK?

ivan-hc:
> But this will result in many curl calls per update, is that OK?

Again, the issue come out when trying to call api.github.com or other domains with rate limit. And this is what we are NOT doing.

Also
```
for ext in digest DIGEST sha512 sha256 sha1 md5; do
	digest=$(curl -Ls "$download_url.$ext")
	if ! echo "$digest" | grep -qi "not found"; then
		break
	fi
done
```
...means that if an output doesn't contain "not found," all subsequent outputs will be ignored, in the order I indicated in the first comment.

And yes, the number of `curl` requests on certain domains could increase if a program doesn't comply with the standards. So...
```
if echo "$digest" | grep -qi "not found"; then
	return 1
fi
```
It might definitely slow down updates by a few seconds, but not by much. It's a slowdown I've noticed since we added this feature, but it's negligible.

vishnu350:
Yes, I went through the code. It is the best we can do so far, looks very good to me. Maybe you also want to just remove the commented out section as well? Since it is no longer needed.

> Again, the issue come out when trying to call api.github.com or other domains with rate limit.

I see, so then as long as we don't exceed the rate limit we should be OK. If we ever do see issues, we can always add a microsecond sleep command in between to reduce the rate.

ivan-hc:
> > Again, the issue come out when trying to call api.github.com or other domains with rate limit.
> 
> I see, so then as long as we don't exceed the rate limit we should be OK. If we ever do see issues, we can always add a microsecond sleep command in between to reduce the rate.

You don't understand how it works, let me explain:

Have you ever accessed a website to view some videos or photos, only to be greeted by a message saying "Subscribe"? It's the same thing.

Github provides you with a maximum limit per hour. You can contact api.github.com using your IP a maximum of 60 times per hour. Once the counter is 0, you'd have to wait an hour to install or update a single app.

There are three ways to work around this:
1. Create a token on GitHub (you must have an account). In AM, you can use it via the `apikey` option.
2. Use Tor (or a Tor frontend, such as `torsocks`, which we've used in the past for this project), with its speed limits and the need to restart the service (as root) if you reach the API limit again.
3. Use a Proxy.

When AM reaches a minimum of 5 remaining calls, the proxy kicks in.

However, since it is a limited resource, I try not to overload it, trying to conserve the available Github APIs as much as possible, to use them according to the rules.

ivan-hc:
You can set a custom proxy by exporting the variable "$ALT_GH", below you can see what we use:

https://github.com/ivan-hc/AM/blob/53605a45eb093310e1ec7d751e950a46abfb482d/APP-MANAGER#L1534

ivan-hc:
To see API left, you can run
```
curl -Ls "https://api.github.com/rate_limit" | tr '{,' '\n' | grep -i remaining | tail -1 | grep -Eo "[0-9]*"
```

ivan-hc:
Obviously, anyone with more than 50 apps installed from GitHub will never use the APIs via AM, because the difference between the number of apps and the available APIs will be equal to or less than 5, so they will always use the Proxy.

People like me, who have 21 apps, 14 of which are from api.github.com, will drop to 45 with an update. I can update all my apps three times in an hour without touching the Proxy.

I would abolish the Proxy if GitHub were more permissive. But I understand that such a large platform needs to limit access to its APIs to avoid overloading itself.

While waiting for improvements, we use this mechanism as a plan B, but never as the primary method, trying to comply with the rules as much as possible.

vishnu350:
Thanks @ivan-hc, I only knew a little about the API limits, but did not know the exact details and history of AM with proxy/Tor. Your solution is interesting.

Actually I was referring to the curl request returning a "Not Found" response by the server. I think it is likely there is an internal rate limit for that, maybe in the range of X number of requests per second. If we exceed X/s, Github will throttle the response, but I'm not sure if there is any other penalty.

Issue #2141, opened 2026-03-12
Many apps now report a checksum warning while updating


marco-calautti:
# Premise
By opening this issue, I certify:
- that I have updated "AM"/"AppMan" to the latest version
- that I have read the README https://github.com/ivan-hc/AM/blob/main/README.md
- that I have read the documentation https://github.com/ivan-hc/AM/tree/main/docs
- that I have searched for an option or flag in `-h` that covers my use case
- that I have searched for an issue that covers my use case https://github.com/ivan-hc/AM/issues

--------------------------------------------------

## Are you using "AM" or "AppMan"?
AppMan

## What GNU/Linux distribution are you using it on?
Arch Linux

## Please, describe the bug
When updating with `appman -u` for most of my apps, I get checksum warnings or even some errors:

 ✔ VIBER è aggiornato, 0 secondi trascorsi! ⚠ Impossibile verificare il checksum, URL di download mancante.
 ✔ RETROARCH è aggiornato, 1 secondo trascorso! ⚠ Impossibile verificare il checksum, URL di download mancante.
 ✔ PCSX-REDUX è aggiornato, 1 secondo trascorso! ✔ Checksum non è richiesto per i file tarball/zip.
 ✔ ZOOM è aggiornato, 0 secondi trascorsi! ⚠ Impossibile verificare il checksum, nessun hash trovato.
 ✔ MOONLIGHT è aggiornato, 2 secondi trascorsi! ⚠ Impossibile verificare il checksum, nessun hash trovato.
 ✔ JDOWNLOADER2 è aggiornato, 2 secondi trascorsi! ✔ Il checksum corrisponde, verifica riuscita.
 ✔ WINBOAT è aggiornato, 1 secondo trascorso! ⚠ Impossibile verificare il checksum, nessun hash trovato.
 ✔ SUPERSONIC è aggiornato, 2 secondi trascorsi! ⚠ Impossibile verificare il checksum, nessun hash trovato.
 ✔ RUSTDESK è aggiornato, 2 secondi trascorsi! ⚠ Impossibile verificare il checksum, nessun hash trovato.
 ✔ DUCKSTATION è aggiornato, 3 secondi trascorsi! ✔ Il checksum corrisponde, verifica riuscita.
 ✔ TELEGRAM è aggiornato, 3 secondi trascorsi! ✔ Il checksum corrisponde, verifica riuscita.
[...]

PR #2159, opened 2026-03-17
Updated app listing to display checksum verification failures.

vishnu350:
This update will show apps that failed checksum verification, on the rare occasion that it does happen. Users may not have noticed the error during `am -u`, so this will help to remind them to reinstall the app.

```
 - APPNAME            | VERSION         | TYPE           | SIZE
 - -------            | -------         | ----           | ----
 ◆ cromite            | 145.0.7632.120✓ | appimage       | 183 MiB
 ◆ inkscape           | 1.4.3✖          | appimage       | 115 MiB

```

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -849,7 +849,11 @@ _check_version() {
 			fi
 			APPVERSION=$(echo "$APPVERSION" | sed 's/  / /g; s/ $//g')
 			if [ -f "$argpath"/AM-VERIFIED ]; then
-				APPVERSION=$(echo "$APPVERSION✓" | awk '{gsub(/✓/, "\033[32m✓\033[0m"); print}')
+				if grep -q "Checksum failure" "$argpath"/AM-VERIFIED; then
+					APPVERSION=$(echo "$APPVERSION✖" | awk '{gsub(/✖/, "\033[31m✖\033[0m"); print}')
+				else
+					APPVERSION=$(echo "$APPVERSION✓" | awk '{gsub(/✓/, "\033[32m✓\033[0m"); print}')
+				fi
 			fi
 			if echo "$obsolete_apps" | grep -q "^$arg [0-9]"; then
 				obsolete_last_release=$(echo "$obsolete_apps" | grep "^$arg " | awk '{print $2}')
@@ -1467,6 +1471,7 @@ _use_verify() {
 		if [ -z "$results" ]; then
 			checksum_msg=$(echo $"Checksum does not match, please manually verify file integrity.")
 			printf "%b✖\033[0m %b \n" "${RED}" "$checksum_msg" | _fit
+			printf "Checksum failure" > "$argpath"/AM-VERIFIED
 		else
 			checksum_msg=$(echo $"Checksum verified.")
 			printf "%b✔\033[0m %b \n" "${Green}" "$checksum_msg" | _fit


ivan-hc:
I don't think it's necessary. We already have a green flag next to the version, everywhere.

ivan-hc:
Also, an if/else would be enough for this task
```
if [ -f "$argpath"/AM-VERIFIED ]; then
	APPVERSION=$(echo "$APPVERSION✓" | awk '{gsub(/✓/, "\033[32m✓\033[0m"); print}')
else
	APPVERSION=$(echo "$APPVERSION✖" | awk '{gsub(/✖/, "\033[31m✖\033[0m"); print}')
fi
```

Istantanea_2026-03-17_04-10-34

...But I repeat, it's not necessary. And it's ugly.

ivan-hc:
> ...But I repeat, it's not necessary. And it's ugly.

And it's alarming. People are more easily shocked when they see this.

ivan-hc:
As I said in the release notes, just because an app doesn't pass the checksum doesn't mean it's invalid.

We can point to all the signs we want; it wouldn't change the way users interpret things.

It was evident after the release; in five days, I was contacted by three concerned users (two in issues and one on Discord). Many don't even know what a checksum is, nor do they care.

Want to know how many apps fail the checksum?

```
am -f | grep -v ✓
```

Istantanea_2026-03-17_04-21-30

vishnu350:
Oh no, that's not what it will do.

It will only highlight the app with **✖ checksum mismatch** during `am -u`, which is a real error. Like below (inkscape only):
```
 - APPNAME            | VERSION         | TYPE           | SIZE
 - -------            | -------         | ----           | ----
 ◆ krita              | 5.2.16✓         | appimage       | 322 MiB
 ◆ gimp               | 3.2.0           | appimage       | 218 MiB
 ◆ cromite            | 145.0.7632.120✓ | appimage       | 183 MiB
 ◆ inkscape           | 1.4.3✖          | appimage       | 115 MiB
 ◆ sayonara           | 1.11.0          | appimage*      | 48 MiB
 ◆ yt-dlp             | 2026.03.13      | dynamic-binary | 35 MiB
 ◆ nvim               | 0.11.6✓🔒       | appimage       | 11 MiB
 ◆ appimageupdatetool | a76b0ca08✓      | appimage       | 5.9 MiB
 ◆ am                 | 9.9.5-5         | bash-script    | 232 KiB
```

Reinstalling inkscape will fix the problem, it will become green because it has a .zsync file.

Actually, this error happened to me and I accidentally closed my terminal, maybe a wget failure. And I forgot which app failed. When a real checksum error occurs, running am -f will skip the green flag for this app, so it looks like other apps that have no .zsync files.

ivan-hc:
Why inkscape? Is this only a sample or you really have it unverified?

I installed it and it results verified (it is my Archimage).

vishnu350:
It is a random failure, possibly due to poor network conditions, the downloaded binary is bad, etc. Maybe around a 0.01% chance to happen. Like Marco's checksum failure for imhex in #2141.

In this case, the downloaded appimage binary is not good to use, it must be re-downloaded. Removing it and reinstalling it will solve the problem, it will become verified again.

ivan-hc:
Oh, now I understand what do you mean with this https://github.com/ivan-hc/AM/issues/2141#issuecomment-4046797407

ivan-hc:
@vishnu350  From Reddit

> Hi Ivan, I noticed this error/warning:
>
> ✔ PLEXAMP is updated, 4 seconds elapsed! ✖ Checksum does not match, please manually verify file integrity.
>
> But I'm not sure what to do. Uninstalling/reinstalling doesn't fix it.

Istantanea_2026-03-18_20-40-20

This is the fourth one.

Any idea?

vishnu350:
Hi @ivan-hc,

The plex.tv server is not responding "Not Found". Instead, it is giving an "Access Denied" response.
`curl -Ls https://plexamp.plex.tv/plexamp.plex.tv/desktop/Plexamp-4.13.0.AppImage.sha1`<?xml version="1.0" encoding="UTF-8"?>
`<Error><Code>AccessDenied</Code><Message>Access Denied</Message><RequestId>.......</HostId></Error>0`

So we need to modify line that does this check. I'll make a pull request.

vishnu350:
Basically, the use_verify function thinks curl found a digest file. But actually it didnt, it was actually an AccessDenied response. Then it checks the checksum and thinks it is a mismatch, instead of going to `return 1`.

PR #2164, opened 2026-03-19
Fix for some servers not responding "Not Found" (plex)

vishnu350:
Updated curl "Not Found" error checking for digest file curl search loop.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1437,11 +1437,11 @@ _use_verify() {
 				# Search other sources if no zsync file
 				for ext in digest DIGEST sha512 sha256 sha1 md5; do
 					digest=$(curl -Ls "$download_url.$ext")
-					if ! echo "$digest" | grep -qi "not found"; then
+					if ! echo "$digest" | grep -qi "not found\|access denied"; then
 						break
 					fi
 				done
-				if echo "$digest" | grep -qi "not found"; then
+				if echo "$digest" | grep -qi "not found\|access denied"; then
 					return 1
 				fi
 			fi


ivan-hc:
Thanks for this fix

vishnu350:
You are a true trooper @ivan-hc. This is the least I can do to help, do not worry about thanking me.

ivan-hc:
@vishnu350 , @psadi had this issue

Istantanea_2026-03-19_19-06-57

Istantanea_2026-03-19_19-08-02


Screenshot_From_2026-03-19_23-38-06

vishnu350:
Hi @ivan-hc, this is the ~0.01% chance random error I was talking about. Most likely due to network problems during `am -u`, the binary either did not finish downloading or somehow became corrupted.

This issue also exists in previous versions of AM (without _use_verify_), but the users just have no way to know there was an error. The bad Appimage will still run.

Running `am -u` again will not help, the error will still remain as the _use_update_ function will not re-download the Appimage binary. Right now, the only way to solve the issue is to do `am -r app` and then `am -i app`.

Here is an example way to reproduce the problem by manually corrupting the appimage binary using `hexedit`. In this example, I do it to nvim.

https://github.com/user-attachments/assets/332a6e2f-118d-446e-8602-9db1676410a6

ivan-hc:
Thanks for clarifications.

PS: why your `am -f` table have no ◆ at start of each appname? This causes you table to look weird.

vishnu350:
> PS: why your am -f table have no ◆ at start of each appname? This causes you table to look weird.

I didn't want to corrupt my real nvim install, so for testing/coding I always play with AM inside a docker container (debian:latest). AM behaves a little strange in docker containers, not sure why. Will need to debug.

vishnu350:
@ivan-hc, it seems AM needs these for `am -f` to work properly within a Debian docker container:
1) `column` command missing: `apt install bsdextrautils`
2) `file` command missing: `apt install file`
3) The locale needs to be set: `export LC_ALL=en_US.UTF-8`

After that `am -f` works great.

Docker containers have an extremely minimal setup, so this is expected.

ivan-hc:
Oh, I see

1. We are using a function to fallback if `column` does not exists https://github.com/ivan-hc/AM/blob/afa22ad1ad159daa7748048da4d03c3e2a965398/modules/database.am#L165-L187
2. For `file`is a bit more complex, since there are no POSIX shell alternative to it. We have already tried to replace it in the past, but sometime it is necessary.
3. For LC_ALL  I think the issue is the ◆symbol, isn't it? But it appears during the installation.

vishnu350:
> For LC_ALL I think the issue is the ◆symbol, isn't it? But it appears during the installation.

Yes. I'm pretty sure the problem is only because of column during `am -f`, piping ◆ or ✓ into column messes it up if LC_ALL is not set. Everywhere else it seems OK.

If I remove column, the output looks like this:
```
 - APPNAME            | VERSION               | TYPE        | SIZE
 - -------            | -------               | ----        | ----
  openrgb            | 1.0rc2                | appimage*   | 32 MiB
  appimagetool       | 1.9.1                 | appimage    | 15 MiB
  appimageupdatetool | a76b0ca08✓ | appimage    | 5.9 MiB
  am                 | 9.9.5-7               | bash-script | 232 KiB
```

ivan-hc:
> If I remove column, the output looks like this:

Yes, because there's a tab space between a separator and the content.

If the ball appeared at the beginning of the lines, everything would look tidy.

That missing column ruins everything. I don't remove the balls and dashes from the first column for visibility reasons. It's more readable. The apps stand out more.

Anyway...

PR #2169, opened 2026-03-21
Allow am -u to redownload Appimages in the event of checksum failure

vishnu350:
I added a check in _update_run_updater() to see if the checksum failed. If yes, it will delete the version file. This will force the AM-updater scripts to redownload the Appimage binary file.

Can be simulated and tested with the following steps:
1. Manually corrupt an Appimage using hexedit.
2. Run `am -u`, it will show checksum failure ✖. `am -f` will also show ✖.
3. Run `am -u` again, this time it will trigger the re-download and fix the error. `am -f` will show a green ✓.

This will negate the need to do a reinstall via `am -r app` and then `am -i app`

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1551,7 +1551,7 @@ _update_run_updater() {
 		fi
 	fi
 	if grep -sq "Checksum failure" "$argpath"/AM-VERIFIED; then
-		rm "$argpath"/version # Triggers AM-updater to re-download the Appimage binary
+		rm -f "$argpath"/version # Triggers AM-updater to re-download the Appimage binary
 	fi
 	if [ -z "$debug_update" ]; then
 		"$argpath"/AM-updater >/dev/null 2>&1


ivan-hc:
What if the error is due to a false-positive?

vishnu350:
> What if the error is due to a false-positive?

Do you mean like the network problems during `am -u`?

ivan-hc:
Yep

vishnu350:
For us, network problems are considered real errors, not false. The checksum fails during am update because the newly downloaded appimage is bad/corrupt or is just still the old version un-updated. It must be re-downloaded.

I don't think there are any false-positive case for us, so far.

ivan-hc:
@psadi @fiftydinar @Samueru-sama

Samueru-sama:
lgtm

vishnu350:
Hi @Samueru-sama, will removing the version file work for **all** AM-updater scripts? They all seem to have these lines:
```
if [ "$version" != "$version0" ]; then
    ...
    wget "$version" || exit 1
    ...
    echo "$version" > ./version
    ...
fi
```
`$version0` is the URL from the version file (old version), `$version` is the new version.

Sometimes `wget` or `appimageupdatetool` fails to update to the latest appimage correctly, but it still stores the latest URL into the version file. This is what makes AM-updater refuse to update the app anymore, and also causes the checksum failure.

So the simplest solution is to delete the version file to force this AM-updater if condition to pass, but I am not 100% sure if it works for all cases.

ivan-hc:
The "version" file only contains a reference to compare. If `appimageupdatetool` is installed and the corrupted program is not updatable, nothing would happen. `appimageupdatetool` is still involved. The "version" file may or may not be created.

If `appimageupdatetool` is not installed, what you say would certainly work.

Samueru-sama:
> Hi @Samueru-sama, will removing the version file work for **all** AM-updater scripts? They all seem to have these lines:

shouldn't be a problem, I actually often remove the `version` file when I want to force an app to update. 

> Sometimes wget or appimageupdatetool fails to update to the latest appimage correctly, but it still stores the latest URL into the version file

with `appimageupdatetool` it works this way: 

```sh
appimageupdatetool -Or ./"$APP" && chmod a+x ./"$APP" && echo "$version" > ./version && exit 0
```

So I don't see how this can update the app without updating the version file, the only way is if `chmod` fails which pretty much impossible. 

Without appimageupdatetool the manual method has an `|| exit 1` if the download fails, which means the `version` file wont be updated as well.

vishnu350:
Ok, you guys are right. It works great, I tested corrupting multiple different apps and they all get fixed by removing the version files and then doing `am -u`. It works regardless of appimageupdatetool install status.

Slightly off topic: Why does appimageupdatetool refuse to fix the corrupt app binary, even though it has a proper .zsync file? I tried manually running `appimageupdatetool -Or $APP` directly (eg. inkscape), but the resulting checksum is still wrong. Only the subsequent wget flow downloads the correct binary.

Samueru-sama:
> Ok, you guys are right. It works great, I tested corrupting multiple different apps and they all get fixed by removing the version files and then doing `am -u`. It works regardless of appimageupdatetool install status.
> 
> Slightly off topic: Why does appimageupdatetool refuse to fix the corrupt app binary, even though it has a proper .zsync file? I tried manually running `appimageupdatetool -Or $APP` directly (eg. inkscape), but the resulting checksum is still wrong. Only the subsequent wget flow downloads the correct binary.

iirc inkscape is hosted on gitlab, those often configure the zsync updates wrong.

ivan-hc:
> iirc inkscape is hosted on gitlab, those often configure the zsync updates wrong.

@Samueru-sama this is inkscape https://github.com/ivan-hc/Inkscape-appimage I am the owner, in AM

Samueru-sama:
> > iirc inkscape is hosted on gitlab, those often configure the zsync updates wrong.
> 
> @Samueru-sama this is inkscape https://github.com/ivan-hc/Inkscape-appimage I am the owner, in AM

uh, upstream inkscape appimage no longer depends on libfuse2 btw.

ivan-hc:
> uh, upstream inkscape appimage no longer depends on libfuse2 btw.

Good to know then, maybe I can archive my repo (if their host is good enough).

ivan-hc:
> if their host is good enough

Same garbage as before.

ivan-hc:
I have just tested this by installing `lxtask` and then replacing the binary with something else. Updating it shows checksum failure and re-updating it uses `wget` to download it again.

I can merge.

PR #2173, opened 2026-03-22
Option -f, improve 🔒 readability in table

ivan-hc:
@Samueru-sama 

| BEFORE | AFTER |
| - | - |
| before | after |

### DIFF
--- a/modules/database.am
+++ b/modules/database.am
@@ -163,31 +163,34 @@ _about() {
 #				FILES
 ################################################################################
 _column() {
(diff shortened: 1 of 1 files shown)


Samueru-sama:
I wouldn't keep the `column -t` code and just always use the `awk` implementation.

ivan-hc:
> I wouldn't keep the `column -t` code and just always use the `awk` implementation.

This is what I'm trying to do the whole afternoon... and this it all I got

Istantanea_2026-03-22_01-57-16

PR #2177, opened 2026-03-23
fix: correct mismatched hash labels in AM-VERIFIED output

xz-dev:
## Summary

Fixes the printf argument order in `_use_verify()` so that hash labels in the `AM-VERIFIED` file match their actual values.

## Bug

The format string is `"MD5: %b\nSHA1: %b\nSHA256: %b\nSHA512: %b"` but the arguments were passed in the wrong order:

| Label | Was receiving | Should receive |
|-------|--------------|----------------|
| MD5 | `$checksum_sha1` | `$checksum_md5` |
| SHA1 | `$checksum_sha256` | `$checksum_sha1` |
| SHA256 | `$checksum_sha512` | `$checksum_sha256` |
| SHA512 | `$checksum_md5` | `$checksum_sha512` |

## Fix

Reorder the printf arguments to match their corresponding format labels:

```diff
-printf "MD5: %b\nSHA1: %b\nSHA256: %b\nSHA512: %b" "$checksum_sha1" "$checksum_sha256" "$checksum_sha512" "$checksum_md5"
+printf "MD5: %b\nSHA1: %b\nSHA256: %b\nSHA512: %b" "$checksum_md5" "$checksum_sha1" "$checksum_sha256" "$checksum_sha512"
```

This is a cosmetic bug — it does not affect pass/fail verification logic (the comparison at line 1470 checks all four hashes against the digest), but the `AM-VERIFIED` file contents were mislabeled.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1475,7 +1475,7 @@ _use_verify() {
 		else
 			checksum_msg=$(echo $"Checksum verified.")
 			printf "%b✔\033[0m %b \n" "${Green}" "$checksum_msg" | _fit
-			printf "MD5: %b\nSHA1: %b\nSHA256: %b\nSHA512: %b" "$checksum_sha1" "$checksum_sha256" "$checksum_sha512" "$checksum_md5" > "$argpath"/AM-VERIFIED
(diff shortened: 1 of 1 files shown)


ivan-hc:
This order was suggested by @fiftydinar , better if we ask him too.

fiftydinar:
> This order was suggested by @fiftydinar , better if we ask him too.

I suggested this order for downloading the checksum files, not for the AM-VERIFIED file info itself. This fixes a bug of incorrectly showing hashes, so you can merge this.

Issue #2191, opened 2026-03-27
Installing same app in system and then user mode and then update in `am` fails


fiftydinar:
Update doesn't work in this scenario.

I ran:
- `am -i featherpad`
- `am -i --user featherpad` (or `am --user` + `am -i featherpad`, no difference)
- `am update`

This is the output:

```
-----------------------------------------------------------------------------
 >> START OF ALL PROCESSES << 
-----------------------------------------------------------------------------
 "AM" CAN MANAGE UPDATES FOR THE FOLLOWING PROGRAMS:
-----------------------------------------------------------------------------

 \u25c6 freetube 0.23.15
 \u25c6 ruffle 2026\u2713

 All self-updatable programs are excluded
-----------------------------------------------------------------------------
/usr/local/bin/am: line 1589: cd: $'/var/opt/featherpad\n/home/gidro/Applications/featherpad': No such file or directory

Update finished (code 1)
```

So should this be fixed, or simply forbidden to be done at all?

ivan-hc:
I've seen this issue before. I don't think it makes sense to have the same program installed on both levels.

PR #2194, opened 2026-03-28
Updated install check condition to avoid app conflicts

vishnu350:
Based on #2191 by @fiftydinar. Only one line changed in install.am.

Updated `am -i app` check conditions to detect if the app command exists and stop the installation to avoid conflicts between system, AM, and appman.

It doesn't make sense to install an app that already exists in the env from any of these sources.

Original code only checks if the app exists in /usr/local/bin AND if it is in AM mode. This is not good because a newer/different version may already exist in the system/appman and should be uninstalled first.

Rationale: If the user insists on installing it anyway, they should just download the install script and do it manually. By default, this should not be allowed as newer apps may corrupt existing configuration/settings that it uses for itself.

### DIFF
--- a/modules/install.am
+++ b/modules/install.am
@@ -596,7 +596,7 @@ case "$1" in
 			# Various cases that may occur during installation
 			if test -f "$APPSPATH"/"$pure_arg"/remove || test -f "$APPSPATH"/"$pure_arg"/remove.old; then
 				echo $" ◆ \"$pure_arg\" is already installed!" | tr '[:lower:]' '[:upper:]'
-			elif [ -n "$(command -v "$pure_arg" 2>/dev/null)" ] || [ -n "$(PATH=/usr/local/bin command -v "$pure_arg" 2>/dev/null)" ] || [ -n "$(PATH="$BINDIR" command -v "$pure_arg" 2>/dev/null)" ]; then
+			elif [ -n "$(PATH="$PATH":"$BINDIR":/usr/local/bin command -v "$pure_arg" 2>/dev/null)" ] ; then
 				echo $" 💀 ERROR: \"$pure_arg\" command already exists!"
 			elif echo "$arg" | grep -q "/"; then
 				if test -f "$arg" 2> /dev/null; then


ivan-hc:
@Samueru-sama

fiftydinar:
This will also stop the installation if the package is installed through system package manager, which is something that is not desired. For example, what about python which `am` has?

We should target `AM` only in this detection, so removing `$PATH` check and only checking `$BINDIR` and `/usr/local/bin`.

```sh
			elif command -v "/usr/local/bin/$pure_arg" 1>/dev/null || command -v "$BINDIR/$pure_arg" 1>/dev/null; then
```

maybe this check can be also better, to check if it is a symlink of `am` / `appman` path, but it's better than what's proposed at least.

Samueru-sama:
`PATH=/usr/local/bin command -v "$pure_arg"` was added because on FreeBSD system packages get installed by the package manager in `/usr/local/bin`.

So to avoid existing packages from being overwritten the check for `/usr/local/bin` was added. I guess you can also add a check for `$BINDIR` when `appman` is in use.

But in any case only check the locations where `am` is going to install, I often install packages that I already have on my system in `/usr/bin` because the ones from `am`/`appman` are newer versions.

vishnu350:
> This will also stop the installation if the package is installed through system package manager, which is something that is not desired. For example, what about python which am has?

Sometimes older distros need a specific version of python v2, wouldn't overriding it be disastrous? Also, I have mistakenly overridden my neovim install as well, and it ruined my config files.

Maybe we give a warning about this?

fiftydinar:
> Sometimes older distros need a specific version of python v2, wouldn't overriding it be disastrous? Also, I have mistakenly overridden my neovim install as well, and it ruined my config files.
> 
> Maybe we give a warning about this?

Warning about installed package in system PATH outside of `am`/`appman` sounds good to me.

Forbidding installation if package is installed in `am`/`appman` PATH.

ivan-hc:
> Forbidding installation if package is installed in `am`/`appman` PATH.

...and freeBSD, that installs regularly its programs in /usr/local

fiftydinar:
> > Forbidding installation if package is installed in `am`/`appman` PATH.
> 
> ...and freeBSD, that installs regularly its programs in /usr/local

Well, `am` path matches the FreeBSD path `/usr/local/bin/`, so it will also cover that.

vishnu350:
> ...and freeBSD, that installs regularly its programs in /usr/local

This should be included in the system PATH. The current pull request will do this, it will forbid installation if it detects existing executables in **system**, **AM**, or **Appman**.

Personally, I just uninstall the system version first before installing AM's version.

So, forbid for all three cases? Or give a warning for system and forbid only AM/Appman?

PR #2208, opened 2026-03-31
add amarok qmmp

Link4Electronics:
they're audio players

### DIFF
--- a/programs/x86_64-apps
+++ b/programs/x86_64-apps
@@ -1,3 +1,5 @@
+◆ amarok : Unofficial, powerful music player that lets you rediscover your music.
+◆ qmmp : Unofficial, Qt-based multimedia player.
 ◆ 0ad : FOSS historical Real Time Strategy, RTS game of ancient warfare.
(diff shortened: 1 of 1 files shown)


ivan-hc:
> amarok

well done! I gave up time ago https://github.com/ivan-hc/Amarok-appimage

Link4Electronics:
have audacious too, but then you already made for it lol
I'd have to do multiple choice later

ivan-hc:
I would prefer to convert all portable apps to AppImages instead.

To have multiple options is OK, each build would work better than another. The user chooses.

But until then, there are the (for now) 408 portable apps that cannot me isolated or sandboxed in AM.

With @Samueru-sama I started doing something here https://github.com/Portable-Linux-Apps and I have also created a tool lately, for this scope https://github.com/ivan-hc/portable2appimage

This would help even more if you lack of fantasy (see the "[wishing list](https://github.com/ivan-hc/AM/blob/main/programs/x86_64-portable)" lol).

PR #2215, opened 2026-04-03
Added two regression test scripts.

vishnu350:
@ivan-hc, @fiftydinar.

These are two simple tests that can be run independently by testers on their systems to help with regression testing. They will display either **pass** or **fail** to help with debugging. 

The idea is we should add more tests here to catch problems that occur in the future like sandboxing failure etc, and run all of them before triggering a release.

Added two sample regression scripts.
1. am_install_test.sh - Installs two apps with the same name and checks if they are listed correctly.
2. am_checksum_test.sh - Installs an app with known checksum verified status and checks if it was verified correctly.

NOTE: I hardcoded zsync2 in the test script because it is a small download. Usually test devs will randomize this for testing, but what if it randomly picks blender (1GB file).

### DIFF
--- a/regress/am_checksum_test.sh
+++ b/regress/am_checksum_test.sh
@@ -1,0 +1,21 @@
+#!/usr/bin/env bash
+
+# Error checking function
+_check_count () {
+if [ "$(grep -i "$2" "$1" | wc -l)" != "$3" ]; then
+	echo Error: Test failed, $2 should occur $3 times.
+	exit
+fi
+}
+
+# Test install (system)
+am --system
+am -r zsync2
+am -i zsync2 > .results
+
+# Check for errors
+_check_count .results "checksum verified" 1
+
+# Pass the test if all was good
+echo "Test passed!"
+
--- a/regress/am_install_test.sh
+++ b/regress/am_install_test.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+# Error checking function
+_check_count () {
+if [ "$(grep -i $2 $1 | wc -l)" != $3 ]; then
+	echo Error: Test failed, $2 should occur $3 times.
+	exit
+fi
+}
+
+# Test install (system)
+am --system
+am -i zsync2
+
+# Test install (user)
+am --user
+am -i zsync2
+
+# Check listing
+am --system
+
+# Check for errors
+am -f > .results
+_check_count .results zsync2 2
+
+# Test removal
+am --user
+am -r zsync2
+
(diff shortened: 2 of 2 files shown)


ivan-hc:
I noticed that in some steps you use `am --user`, but this wont work if you don't set an AppMan configuration file first, in ~/.config/appman/appman-config

Anyway, this is a good idea. I will keep this opened for new implementations.

fiftydinar:
Good start, but for a reference, `AM-CHECK` is an example of regression tests done right.

They are complex, but they are really good.

The idea would be to run the regression scripts in GitHub action and get a report if there is a regression or not.
It can be set to run on every PR made, where you can see the checkmark if PR makes a regression or not.

ivan-hc:
@vishnu350 I have closed this PR by mistake, since the main branch had to be replaced and removed due to a critical bug.

Feel free to re-open this if you can.

fiftydinar:
> due to a critical bug.

For reference, this PR of mine caused the critical bug:
https://github.com/ivan-hc/AM/pull/2232

This situation above further shows why regression tests are needed.

vishnu350:
> Feel free to re-open this if you can.

Hi @ivan-hc. Of course I think this is needed to prevent issues and eventually bring you to POSIX compliance. But I don't want to impose on your project.

It is important that these tests must be able to be executed locally by devs or testers using different systems (32-bit, 64-bit, BSD, docker, etc). I have already made two basic tests, these can be a template for newer tests. We can modify them or keep adding more over time as users report more bugs, or if you guys have ideas on failure scenarios.

Normally these tests will just be in an isolated folder called _regress_ or _tests_ or _verif_. The error checking functions can be common for all, and since AM is a shell program with output, we can just rely on grep as the checking mechanism.

If you are worried about the tests themselves, they do not need to be "perfect" because tests do not affect AM in any way other than for testing. It is completely fine if a test is "wrong" at first.

> AM-CHECK is an example of regression tests done right.

The github actions yaml running regressions should be the next step. It should eventually point to these tests instead of a single hardcoded test within.

Once AM has enough tests, it will be easier to make changes for POSIX compliance without breaking functionality.

ivan-hc:
@vishnu350 I'm unable to re-open this PR. Can you re-do the pull request? Or want to add more scripts first? You choose.

vishnu350:
> Can you re-do the pull request? Or want to add more scripts first? You choose.

I'll think a little bit more on the template and checkers, maybe tweak it a little bit more and redo the PR.

PR #2247, opened 2026-04-10
Add antigravity

cheack:


programs/x86_64/antigravity
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
+
+# AM INSTALL SCRIPT VERSION 3.5
+set -u
+APP=antigravity
+SITE="BOTOOM/google-antigravity-bin-arch"
```
cheack: Will fix now.

ivan-hc:
Is this an official deb package?

Anyway, consider the use of the option 1 of `-t`, AppImage on-the-fly

I have done a test with portable2appimage and works

https://github.com/user-attachments/assets/e6f20698-07c4-4eff-ad90-99f2fcc8d612

You don't even need to add a launcher and an icon in this case. It do everything for you.

The usage is the same as 0, normal ApImages, with the difference that it asks for a binary name (that in this case is the same).

ivan-hc:
@cheack see here


https://github.com/user-attachments/assets/eac0bfe7-c1f9-4bee-83e0-6a415c4e5d1f

ivan-hc:
If you want I can do it for you

cheack:
@ivan-hc 


> Is this an official deb package?

Yes. The .deb is downloaded directly from Google's official apt repository (us-central1-apt.pkg.dev/projects/antigravity-auto-updater-dev), the same source used in their official installation instructions at antigravity.google. The PKGBUILD is only used to extract the current download URL.

> If you want I can do it for you

Yes, that would be great, thank you!
I initially looked at some existing scripts that use .deb packages and they all just extract the archive directly, so I thought that would be the simpler and faster approach.

ivan-hc:
> > If you want I can do it for you
> 
> Yes, that would be great, thank you! I initially looked at some existing scripts that use .deb packages and they all just extract the archive directly, so I thought that would be the simpler and faster approach.

Yes, I'm trying to port as many portable apps as possible to AppImage format, even if it means building them on the fly.

Including this one, we're at 5.

PR #2248, opened 2026-04-10
Use `yes` command for nolibfuse scripting

aleksa-radojicic:
Using the `yes` command will work when multiple programs are supplied, but `echo y` will work only for a single program.

When reading the docs you don't get the impression multiple programs can be supplied. Also goes when using `am -h`.

### DIFF
--- a/docs/guides-and-tutorials/nolibfuse.md
+++ b/docs/guides-and-tutorials/nolibfuse.md
@@ -35,12 +35,12 @@ The original AppImage will be extracted using the `--appimage-extract` option, a
 #### Updating converted AppImages
 The `nolibfuse` option adds the following lines at the end of the AM-updater script
 ```
-echo y | am nolibfuse $APP
+yes | am nolibfuse $APP
 notify-send "$APP has been converted too! "
 ```
 or
 ```
-echo y | appman nolibfuse $APP
+yes | appman nolibfuse $APP
 notify-send "$APP has been converted too! "
 ```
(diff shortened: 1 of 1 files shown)


ivan-hc:
> When reading the docs you don't get the impression multiple programs can be supplied. Also goes when using `am -h`.

Hi, what do you mean?

ivan-hc:
About this

> Using the `yes` command will work when multiple programs are supplied, but `echo y` will work only for a single program.

it only answers to the output of this prompt

https://github.com/ivan-hc/AM/blob/6a2eb1b028bea426e67b1031d9e7d36d2fbf89df/modules/management.am#L494

and here is where is added in the AM-updater script

https://github.com/ivan-hc/AM/blob/6a2eb1b028bea426e67b1031d9e7d36d2fbf89df/modules/management.am#L488

To allow conversion also when updating the app.

aleksa-radojicic:
When `echo y | am nolibfuse $APP1 $APP2` is executed, the old AppImage for $APP1 will be removed automatically, but the prompt "Do you wish to remove the old libfuse2 AppImage? (Y/n)" will still appear for $APP2. Using `yes | am nolibfuse $APP1 ... $APPn` will remove images for all listed apps without interactive prompts. echo y only works for a single program; yes works for an arbitrary list.

ivan-hc:
> When `echo y | am nolibfuse $APP1 $APP2` is executed, the old AppImage for $APP1 will be removed automatically, but the prompt "Do you wish to remove the old libfuse2 AppImage? (Y/n)" will still appear for $APP2. Using `yes | am nolibfuse $APP1 ... $APPn` will remove images for all listed apps without interactive prompts. echo y only works for a single program; yes works for an arbitrary list.

Yes, this is the purpose. AM-updater is for one app. The command is executed only for that app where the AM-updater belong. `nolibfuse` is executed only once, when updating the app.

Your use-case have sense if you run
```
am nolibfuse app1 app2 app3
```
But it is not this the case.

PR #2250, opened 2026-04-11
"AM" 10.1 - Custom PATH for appimagetool if not present

ivan-hc:
This will be downloaded and saved in "$CACHEDIR"/AMCACHEPATH

Once downloaded, "$CACHEDIR"/AMCACHEPATH will be exported in $PATH, so `appimagetool` will be available in the following cases:
- when an app installed (option `-i`) is an AppImage built on-the-fly
- when an old AppImage must be converted to a new one to remove libfuse2 dependency (option `nolibfuse`)

To remove the content of "$CACHEDIR"/AMCACHEPATH, run the option `-c` or `clean`.

This is the first step for an even bigger implementation of portable dependencies throwaway for AM.

@fiftydinar @Samueru-sama @vishnu350

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -540,15 +540,32 @@ _online_check() {
 	fi
 }
 
+# Check AMCACHEPATH integrity
+if [ -d "$CACHEDIR"/AMCACHEPATH ]; then
+	am_bins_allowed=$(echo "appimagetool busybox cat column chmod chown curl file grep less sed wget" | tr ' ' '\n' | sed "s#^#$CACHEDIR/AMCACHEPATH/#g")
+	am_bins_directories=$(find "$CACHEDIR"/AMCACHEPATH/* -type d)
+	am_bins_files=$(find "$CACHEDIR"/AMCACHEPATH/* -type f)
+	for d in $am_bins_directories; do rmdir "${d}" 2>/dev/null; done
+	for f in $am_bins_files; do if ! echo "$am_bins_allowed" | grep -q "$f" 2>/dev/null; then rm "$f" 2>/dev/null; fi; done
+fi
+
 # Check if appimagetool is installed, and if it is not installed, download and use it where it is needed
 _download_appimagetool() {
+	_determine_args
 	if ! command -v appimagetool >/dev/null 2>&1; then
 		mkdir -p "$CACHEDIR"/AMCACHEPATH
 		if [ ! -f "$CACHEDIR"/AMCACHEPATH/appimagetool ]; then
+			_online_check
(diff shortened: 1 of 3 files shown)


vishnu350:
```
	if ! command -v appimagetool >/dev/null 2>&1; then
		mkdir -p "$CACHEDIR"/AMCACHEPATH
		if [ ! -f "$CACHEDIR"/AMCACHEPATH/appimagetool ]; then
			_online_check
			wget -q https://github.com/AppImage/appimagetool/releases/download/continuous/appimagetool-"$(uname -m)".AppImage -O "$CACHEDIR"/AMCACHEPATH/appimagetool
			chmod a+x "$CACHEDIR"/AMCACHEPATH/appimagetool
		fi
	fi
```

Why not just prompt the user to do am -i appimagetool if the command check fails? This is adds a redundant hardcoded URL dependency on https://github.com/AppImage/appimagetool/releases.

Are you afraid of disturbing the translators?

ivan-hc:
If you check the code removed from management.am, you'll notice that the `nolibfuse` option already downloads if `appimagetool` wasn't installed.

The same applies every time an AppImage is compiled on the fly.

In this case, the dependency is downloaded once and for as many subsequent times as needed. It's also removed if the user decides to install `appimagetool` directly.

This approach opens the door to a new implementation, which in future releases will use `busybox` and other static binaries to cover all necessary dependencies.

I plan to remove the "requirements" part from the README. AM will be much more self-sufficient in the future.

I've already discussed this with @fiftydinar , and he seemed to agree. He uses a Fedora-based distro, and as such, uses `wget2` instead of `wget` by default, and they are not the same program https://bugzilla.redhat.com/show_bug.cgi?id=2358792

 In the future, I plan to implement `wget` version 1.x and `curl` as well.

Furthermore, the new PATH can be cleaned with `am -c`, and dependencies (currently only `appimagetool`) are downloaded only when requested.

This is an approach I successfully tested in [portable2appimage](https://github.com/ivan-hc/portable2appimage) and have decided to extend to AM.

vishnu350:
OK, I see. Just asking, because what if probono accidentally deletes AppImage/appimagetool or changes the name.

ivan-hc:
> OK, I see. Just asking, because what if probono accidentally deletes AppImage/appimagetool or changes the name.

@Samueru-sama is not just one of the maintainer but also owns its fork.

appimagetool is a core utility in the AppImage ecosystem. No way to drop it without forks.

ivan-hc:
I also have my own

https://github.com/ivan-hc/appimagetool

PR #2271, opened 2026-04-13
Update FreeCAD download source in installation script

schnoog:
Unfortunately the repo FreeCAD-Bundle isn't kept up to date. Last listed release is from August 2025. (V1.0.1)

New releases are available in the FreeCAD/FreeCAD repo. I've adapted the repo path and added exclusion grep for weekly builds and release candidates to only capture official releases.


ivan-hc:
Thanks @schnoog and welcome on board

PR #2282, opened 2026-04-19
Regression updates: Improved logging, made all tests work unattended, and added test-am-regress (run all tests)

vishnu350:
Regression testing updates:
- Improved logging mechanism for full regression runs
- Made all tests work unattended (user inputs no longer needed)
- Added test-am-regress (runs all tests and prints a summary file).

Just kick off test-am-regress.sh and take a cup of coffee, and then come back and check the results (pass/fail).

PR #2292, opened 2026-04-22
Added sandbox and nolibfuse test

vishnu350:
Regression updates:
- Added sandbox and nolibfuse test.
- Added new tests to regression list.
- Added symlink cleanup in remove_all_apps functions (needed for tmpfs dirs)
- Updated podman run command in README to support sandboxing.

I am not able to get sas to work in a container, this sandbox test uses aisap for now. The test uses an automated way to check file access to confirm if it worked, so next time if sandboxing is broken we will know.

### DIFF
--- a/regress/README.md
+++ b/regress/README.md
@@ -31,9 +31,9 @@ Alternatively, you may select a different container image such as am-ubuntu:\
 `podman build -t am-ubuntu -f am-ubuntu.dockerfile`
 
 3. Start am-debian container session in default mode (persistent Appimages):\
-`podman run -it --device /dev/fuse --cap-add SYS_ADMIN am-debian:latest`\
+`podman run -it --device /dev/fuse --cap-add SYS_ADMIN --security-opt unmask=ALL am-debian:latest`\
 Alternatively, start am-debian container session with tmpfs for Appimage dirs (stores in RAM to avoid wearing out your SSD):\
-`podman run -it --device /dev/fuse --cap-add SYS_ADMIN --tmpfs /opt --tmpfs /root/.local/share/applications am-debian:latest`\
+`podman run -it --device /dev/fuse --cap-add SYS_ADMIN --security-opt unmask=ALL --tmpfs /opt --tmpfs /root/.local/share/applications am-debian:latest`\
 You will now be within the container.
 
 4. Setup AM as needed (eg. switch to dev branch):\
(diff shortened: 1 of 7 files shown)


fiftydinar:
> I am not able to get sas to work in a container, this sandbox test uses aisap for now.

@Samueru-sama

ivan-hc:
@vishnu350 are you on Ubuntu?

Samueru-sama:
> > I am not able to get sas to work in a container, this sandbox test uses aisap for now.
> 
> @Samueru-sama

https://github.com/ivan-hc/AM/pull/2282#issuecomment-4281607255

vishnu350:
> @vishnu350 are you on Ubuntu?

The problem with sas is only inside the am-debian/am-ubuntu testing container. It works fine outside.

> [#2282 (comment)](https://github.com/ivan-hc/AM/pull/2282#issuecomment-4281607255)

I went through it, I think the issue is this line:
`USER="${LOGNAME:-${USER:-${USERNAME}}}"`

All of those env variables are not set in the container.

This is what /etc/passwd looks like:
`root:x:0:0:root:/root:/bin/bash`

Samueru-sama:
> > @vishnu350 are you on Ubuntu?
> 
> The problem with sas is only inside the am-debian/am-ubuntu testing container. It works fine outside.
> 
> > [#2282 (comment)](https://github.com/ivan-hc/AM/pull/2282#issuecomment-4281607255)
> 
> I went through it, I think the issue is this line: `USER="${LOGNAME:-${USER:-${USERNAME}}}"`
> 
> All of those env variables are not set in the container.
> 
> This is what /etc/passwd looks like: `root:x:0:0:root:/root:/bin/bash`

Well set the var in the container if empty then. 

POSIX mandates `LOGNAME` to be set upon user login.

vishnu350:
Hi @Samueru-sama. I did try that a few days back, but ran into other issues and then gave up.

Anyway, I tried again and got it to work. First step is to set the missing vars:
`export USER=root`
`export USERNAME=root`
`export LOGNAME=root`

The next problem is it needs /run/user dir with ID (0 for root user in container):
`mkfifo: cannot create fifo '/run/user/0/sas-xdg-open-pipe': No such file or directory`

So I just create that dir since /run/user is not shared by the system with the container:
`mkdir -p /run/user/0 && chmod 777 /run/user/0`

Next error is:
`bwrap: Can't create file at /bin/xdg-open: Read-only file system`

So I figured that I have to cave in and just install xdg-utils which is a 373MB install. The whole am-debian container is only 200MB+, so it is quite a big install.
`apt install xdg-utils`

After that sas works in the container. But is there a way to do this without installing xdg-utils?

Samueru-sama:
> But is there a way to do this without installing xdg-utils?

Pass `--no-xdgopen` flag to sas.

PR #2298, opened 2026-04-25
Added new test-am-clone and added into regression list.

vishnu350:
Added a new test to check the am clone feature.

@ivan-hc I think the feature is broken, the test-am-clone fails every time. AM clone fails to pick up hidden apps, and also the generated list has duplicates. Please see test-am-clone.sh to reproduce the issue.


ivan-hc:
> @ivan-hc I think the feature is broken, the test-am-clone fails every time.

Can you provide more details?

> AM clone fails to pick up hidden apps

This is why they are named "hidden".

> and also the generated list has duplicates.

I already know this, the important is that the command do not runs it twice.

vishnu350:
> This is why they are named "hidden".

So this behavior is expected? AM clone will not pick up hidden apps?

ivan-hc:
> > This is why they are named "hidden".
> 
> So this behavior is expected? AM clone will not pick up hidden apps?

yes

ivan-hc:
Same thing for the apps with version locked (option `lock`), since an older version isn't always expected to be available online, you should downgrade immediately after installation, and the older version isn't necessarily available on GitHub. Apps not on GitHub, however, are "condemned" to always provide the latest version.

vishnu350:
OK, ill fix the test in regards to hide/lock.

> Can you provide more details?

If there is already an existing ~/Desktop/am-clone.source file (with different contents), the generated file will be incorrect. AM will not give any error. The test will use random app names every time you run it, but doesn't delete the am-clone.source file, so that is how this issue was found.

BTW, the am-clone.source file goes into ~/Desktop. This dir may not exist in some systems, shouldn't it just go into the dir where the user ran the am clone command?

ivan-hc:
> If there is already an existing ~/Desktop/am-clone.source file (with different contents), the generated file will be incorrect. AM will not give any error. The test will use random app names every time you run it, but doesn't delete the am-clone.source file, so that is how this issue was found.
> 
> BTW, the am-clone.source file goes into ~/Desktop. This dir may not exist in some systems, shouldn't it just go into the dir where the user ran the am clone command?

Please read here

https://github.com/ivan-hc/AM/blob/main/docs/guides-and-tutorials/clone.md#clone-a-configuration-and-export-it-to-other-systems

There is a list of places that AM reads to detect a am-clone.source file.

My ~/Desktop directory is named "Scrivania", and it works for me thanks to XDG specifications.

To not detect the old one, you need to remove it.

PR #2300, opened 2026-04-26
Improve check version

ivan-hc:
the table in the foreground is the new one, you can see some more complete versions

Istantanea_2026-04-26_21-29-19


ivan-hc:
again, the table in the foreground is the new one, you can see some more complete versions

Istantanea_2026-04-27_13-41-15

PR #2302, opened 2026-04-28
Improve versions check

ivan-hc:
Continuation of https://github.com/ivan-hc/AM/pull/2300

The table in the foreground is the new one, you can see some more complete versions

Istantanea_2026-04-28_15-54-23

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-AMVERSION="10.1-1"
+AMVERSION="10.1-2"
 
 # Determine main repository and branch
 AMREPO="https://raw.githubusercontent.com/ivan-hc/AM/main"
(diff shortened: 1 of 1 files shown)


vishnu350:
@ivan-hc zsync2 cannot be rolled back anymore after this update, the ./test-am-rollback caught it. It just says `ERROR: No valid links found, sorry!`

The version it is showing seems to be incorrect:
```
- APPNAME | VERSION   | TYPE        | SIZE
◆ zsync2  | fc62ae7 ✓ | appimage    | 11.92 MiB
```
Before this, it would show v93, previous version was v92.

ivan-hc:
@vishnu350 what are you talking about?

Istantanea_2026-05-01_15-03-13

Sure the version is different in the URL, but the latest commit is a value more reliable than a version.

vishnu350:
In Alpine Linux:
` ◆ zsync2  | fc62ae7 ✓ | appimage    | 11.92 MiB`

In Debian Jessie:
` ◆ zsync2  | 2.0 ✓ | appimage       | 11.92 MiB`

In Debian latest:
` ◆ zsync2  | 93 ✓  | appimage       | 11.92 MiB`

Only Alpine Linux has the rollback problem, other distros are OK.

Maybe ignore this for now, I will try to debug it later.

ivan-hc:
In Debian Testing I have the same result as Alpine

ivan-hc:
This should fix the issue @vishnu350 https://github.com/ivan-hc/AM/commit/baa67d7f6e1e09e4f5b1b39b9ddb8eeb91421ab4

PR #2311, opened 2026-05-01
Fix untranslatable strings

cheack:
While working on the Russian translation, I noticed that some strings never appeared translated. 

In a previous discussion, one of these strings (`No apps to update here!`) was pointed out as already translatable - but it was actually using `$'...'`, so it was never being translated.

This PR replaces `$'...'` with `$"..."` for all strings that are intended to be localized, and adds the corresponding translation entries for all supported languages.

### DIFF
--- a/AM-INSTALLER
+++ b/AM-INSTALLER
@@ -43,7 +43,7 @@ _install_am() {
 	elif command -v doas >/dev/null 2>&1; then
 		SUDOCMD="doas"
 	else
-		echo 'ERROR: No sudo or doas found'
+		echo $"ERROR: No sudo or doas found"
 		exit 1
 	fi
 	$SUDOCMD "$CACHEDIR"/INSTALL-AM.sh && rm -f "$CACHEDIR"/INSTALL-AM.sh
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -331,7 +331,7 @@ _am() {
 	elif command -v doas >/dev/null 2>&1; then
 		export SUDOCMD="doas"
 	else
-		echo $'ERROR: No sudo or doas found'
+		echo $"ERROR: No sudo or doas found"
 		exit 1
 	fi
 	APPSPATH="$(readlink -f /opt)"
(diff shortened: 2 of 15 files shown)


cheack:
I also noticed a multi-line output block in `_install_sandbox_script` ([sandboxes.am](https://github.com/ivan-hc/AM/blob/5b3eebe2cfae408d4a4a1ca737e9ac549024912b/modules/sandboxes.am#L278-L286)) that uses the same pattern, but left it out of this PR - it's split across many small `printf` calls for ANSI color formatting, and translating those tiny fragments in isolation didn't seem right.

ivan-hc:
AM-INSTALLER does not uses translations... but no matters.

ivan-hc:
> I also noticed a multi-line output block in `_install_sandbox_script` ([sandboxes.am](https://github.com/ivan-hc/AM/blob/5b3eebe2cfae408d4a4a1ca737e9ac549024912b/modules/sandboxes.am#L278-L286)) that uses the same pattern, but left it out of this PR - it's split across many small `printf` calls for ANSI color formatting, and translating those tiny fragments in isolation didn't seem right.

Some blocks are a mess.

When translation support was added, $ were added wherever was available a message, or a `echo` or `printf`, without considering the possibility of made those messages even shorter and less complicated to translate.

Actually AM has hundreds of lines, containing ASCII characters and symbols.

I would made them easier to translate... but I don't want to overwrite already existing translations of other users.

ivan-hc:
Man, you did an amazing job. Congratulations!

cheack:
Haha, thanks!

ivan-hc:
@cheack I don't know what's happened but I had to fix some lines containing quotes that were marked as invalid due to mismatches between original and translations

https://github.com/ivan-hc/AM/commit/2844fba4affe7bec0992bbb22e7b8f907c8fb14b

For example, if the original had \", the translation had not " escaped.

Maybe because they were in single quote before, and you fixed the original, and not thee translation.

PR #2318, opened 2026-05-04
"AM" 10.1.1

### LINKED ISSUE #2305: Improve AM install script error handling
Is there a way to add better error detection in _install_arg (install.am)?

For example, if xz is not installed in the system, we will not be able to untar some apps during installation. But AM will still consider the app was installed correctly.

7zip install test case with xz/xz-utils uninstalled: `am -i 7zip`

```
 ◆ "7ZIP": starting installation script

7z2601-linux-x64.tar.xz  100%[=====================================>]
tar (child): xz: Cannot exec: No such file or directory
tar (child): Error is not recoverable: exiting now
tar: Child returned status 2
tar: Error is not recoverable: exiting now
chmod: cannot access './7zz': No such file or directory
chmod: cannot access './7zzs': No such file or directory

 Oops! Something went wrong! Integrity check in progress...

 ✔ Checksum auto-verified.

 "7ZIP" INSTALLED (2 MB OF DISK SPACE)
```

#### Suggested solution
Maybe we modify _install_arg() in install.am in this way:
```
# Install script
$SUDOCMD ./"$arg" 2> "$AMCACHEDIR"/install.log || _install_script_retry
# Check logfile for errors
grep -q "Error \|error \|cannot access\|No such file or directory" "$AMCACHEDIR"/install.log && _installation_cleanup
```

The code greps the log for errors and then calls installation_cleanup instead to fail the install.

But this has issues:
1. Progress bar disappears
2. Alternatively, using `script -c $SUDOCMD ./"$arg"` method causes the output to look weird, and some tests cant pipe inputs into AM correctly (eg. YYNNNNN for `am -ias` test).

ivan-hc:
fix https://github.com/ivan-hc/AM/issues/2305


fiftydinar:
I think it would be good to additionally have `notify-send` call (if available) when missing crucial dependencies, as it's a very important issue and some users might have `am -u` run in the background not seeing the terminal message, so it's better to make them aware of it.

ivan-hc:
> I think it would be good to additionally have `notify-send` call (if available) when missing crucial dependencies, as it's a very important issue and some users might have `am -u` run in the background not seeing the terminal message, so it's better to make them aware of it.

done

fiftydinar:
> > I think it would be good to additionally have `notify-send` call (if available) when missing crucial dependencies, as it's a very important issue and some users might have `am -u` run in the background not seeing the terminal message, so it's better to make them aware of it.
> 
> done

Hmm, maybe I didn't explain good.

I meant to have `notify-send $" 💀 ERROR! MISSING ESSENTIAL COMMANDS\033[0m: %s\n\n Install the above and try again! \n" "$(echo "curl, wget")"

`notify-send "Missing commands will be temporarily downloaded as static binaries:\n\n"`
`notify-send printf " - %b, downloaded from https://github.com/ivan-hc/am-utils\n" "$name"`

and similar

ivan-hc:
> Hmm, maybe I didn't explain good.
> 
> I meant to have `notify-send " 💀 E R R O R ! M I S S I N G E S S E N T I A L C O M M A N D S \0 33 [ 0 m : (echo "curl, wget")"
> 
> `notify-send "Missing commands will be temporarily downloaded as static binaries:\n\n"` `notify-send printf " - %b, downloaded from https://github.com/ivan-hc/am-utils\n" "$name"`
> 
> and similar

It's a vicious circle, then... if a user doesn't have wget or curl, how can they have notify-send? In my opinion, they couldn't even have a Linux distro, let alone AM installed.

ivan-hc:
I mean, installing the same app on two levels is unlikely... but something like this is literally absurd.

fiftydinar:
> I mean, installing the same app on two levels is unlikely... but something like this is literally absurd.

Ok, it can be ignored

ivan-hc:
https://github.com/ivan-hc/AM/pull/2318/commits/81ca9dc06264b8803212fe27f36bce83aff21a36


https://github.com/user-attachments/assets/e77062c9-05ec-4d12-a038-d504f1f362da

ivan-hc:
I have done this change for @Samueru-sama that does not like the idea of a package manager that forces its own binaries.

However, the user could choose if rely on them or not.

In my tests on freeBSD, some of them have helped, in particular `column`, for `am -f`

@fiftydinar @vishnu350 do you agree with this or is it too much invasive?

Could it be done better?

vishnu350:
> I have done this change for @Samueru-sama that does not like the idea of a package manager that forces its own binaries.

@Samueru-sama is absolutely right. Providing the binaries is good, but AM should not force it. It does feel invasive, but still better than the original auto-download version.

ivan-hc:
I have done the message and prompt less invasive https://github.com/ivan-hc/AM/pull/2318/commits/3aa49a44a5a04cd735525b039f60c968c8f0ab29

https://github.com/user-attachments/assets/0cd1c5df-b331-437e-9a70-299273616316

ivan-hc:
[Allow use of temporary GNU binaries on BSD systems](https://github.com/ivan-hc/AM/pull/2318/commits/96bbb79f925064624d084aceac6880f3232e1ea4)[Allow use of temporary GNU binaries on BSD systems](https://github.com/ivan-hc/AM/pull/2318/commits/96bbb79f925064624d084aceac6880f3232e1ea4)


on freeBSD:

| before | after |
| - | - |
| Istantanea_2026-05-08_19-36-25 | Istantanea_2026-05-08_19-35-47 |

ivan-hc:
[Update AM-INSTALLER](https://github.com/ivan-hc/AM/pull/2318/commits/f8cf075aa9ad9f3fab4d8483187bcf627cdfcedc)

No need to have both curl and wget, it is enough to have one to use it.

Also, new UI in table, to made the installation less confusing.

Istantanea_2026-05-09_04-37-22

ivan-hc:
I think it is all ready.

I will merge this on Sunday, if none have to say something against.

vishnu350:
https://github.com/user-attachments/assets/174067ed-daf2-4d55-8986-2c5d73588045

Regression tests are passing.

ivan-hc:
[If binary in custom PATH fail checksum, remove it](https://github.com/ivan-hc/AM/pull/2318/commits/c4264da67219983d620b787a77c21b992d6bf7a6)

This is a last touch. If an attacker replaces the binary in PATH before you enter your password, the suspect binary is removed

https://github.com/user-attachments/assets/35d3a2a0-5515-48b4-bfc7-43f8889ca5de

ivan-hc:
This other one checks for garbage in the custom PATH

https://github.com/ivan-hc/AM/commit/edfb6024207d8995dde3f56d1878d35ee0b6dfc3

ivan-hc:
[If corrupted, remove and re-download the binary before installation](https://github.com/ivan-hc/AM/pull/2318/commits/82043a78cc50218c0785a0fd0a8b68ac71ed55dc)

Istantanea_2026-05-10_02-14-05

ivan-hc:
I have just finished the tour of my virtual machines to fix remaining issues

Istantanea_2026-05-10_03-41-42

Only Windows 10 is left!

lol

- [Allow downloading missing dependencies](https://github.com/ivan-hc/AM/pull/2318/commits/c3b1d18cfdb494383a409c193f5e05ace2bf59c2) so in case it is missing, required and the user have approved previously, it can be downloaded again
- [Handle only standard dependencies in custom PATH](https://github.com/ivan-hc/AM/pull/2318/commits/74065038430b3fc3cb6d17e99274de2e9c0a68d6), so `appimagetool` is keept, in case it is present (being it not a standard dependency)
- [Remove fake dependencies](https://github.com/ivan-hc/AM/pull/2318/commits/fc880a1264450e51d0b1ad72883ac1a9e1a50a1a), this means that, in case that there is a binary that is not a dependency you need, it is removed. I did the test with a script I named `file`, a dependency that exists but is not needed, and it has been removed
- [Check if appimagetool in custom PATH is an AppImage](https://github.com/ivan-hc/AM/pull/2318/commits/391075674e96a89da6c6de73198daba6596b627a), here I did the test with a different file, and `appimagetool` is removed if it is not an AppImage. Well, on this tool we should add another check, since it is the only program that can be used in our custom PATH since [AM 10.1](https://github.com/ivan-hc/AM/releases/tag/10.1) have been released

@vishnu350 @fiftydinar I did so many tests that I think the new release need another day or two to be released

ivan-hc:
[Fix appimagetool URL in freeBSD](https://github.com/ivan-hc/AM/pull/2318/commits/a11c73b105e0f59dd187c9ed07f2ec7900517096)

To made it work on that system is another task -_-"

ivan-hc:
Time to merge.

PR #2323, opened 2026-05-05
README.md: Replace Main Index with generated TOC

andy5995:
A little utility from https://github.com/ekalinin/github-markdown-toc.go

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,46 +39,22 @@ You can use the command `am -a {PROGRAM}` to view the description and get the so
 # Main index
 ------------------------------------------------------------------------
 
-[Installation](#installation)
+* [Installation](#installation)
+  * [Using the "AM\-installer" script to choose between local and system\-wide installation](#using-the-am-installer-script-to-choose-between-local-and-system-wide-installation)
+  * [Using "GIT" (only system\-wide installation)](#using-git-only-system-wide-installation)
+  * [Using a one\-line command (only system\-wide installation)](#using-a-one-line-command-only-system-wide-installation)
+  * [What is AppMan?](#what-is-appman)
+  * [Recommended packages](#recommended-packages)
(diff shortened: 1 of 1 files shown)


ivan-hc:
Hi Andy, the utility have made an incorrect Index:
- USAGE and THIRD_PARTY FLAGS do not redirects to anywhere useful
- The whole part that explains what is AppMan and the difference with AM is disappeared

andy5995:
Yes, I noticed that too... I'm trying to figure out why... I'll check back with you...

andy5995:
Seems like those are fixed now. But there a few links missing from this version that were included in your original version. The level headings in README.md would probably need to be adjusted. Maybe each of the usage options listed would need `####`; then they would be omitted using `--depth 3`. I used `gh-md-toc --depth 2 README.md` so using a depth of 3 would add some of the missing links, but might add a lot more than what you want. I'll leave this until tomorrow. If you wanna keep things as they are, that' s fine, just close this. ;)

ivan-hc:
This tool really doesn't like to explain how apps are installed and what AppMan is.

To paraphrase Agent Smith: "Never send a machine to do a human's job."

ivan-hc:
@andy5995 Don't get me wrong, I'm closing this because, as a change, it's a regression at the moment.

I like the fact that many trailing spaces are identified and removed. This is very useful. But I don't like the fact that the index is stripped of important information. Not to mention that the index is written so "wide" for readability reasons (I'm shortsighted; when the text is so tightly packed, I find it difficult to focus on the essential lines, and based on this, I write guides, messages, and texts by dividing everything into paragraphs).

Maybe another time. You can continue practicing on your fork, though. I'd be happy to accept a pull request.

andy5995:
Good call. :)

PR #2327, opened 2026-05-06
Added _set_access_privilege() to securely lock AMCACHEPATH in AM mode.

vishnu350:
The current implementation of _deps_dl in the dev branch is still vulnerable to a [timing attack](https://en.wikipedia.org/wiki/Timing_attack), even though the checksum feature is well done. Please, go through my proposed solution carefully as this is a security issue.

I have added a _set_access_privilege function to securely lock the files/folder (AMCACHEPATH) if in AM mode, which I have tried to explain a few times in #2305. This will securely prevent tampering and effectively block off timing attacks.

Still issues to resolve:
1. Switching from AM to APPMAN mode will not fully work (yet).
2. Keying in the password is annoying.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -511,13 +511,25 @@ _online_check() {
 	fi
 }
 
+_set_access_privilege() {
+       if [ "$AMCLI" = am ]; then
+               # Lock ownership to prevent tampering (AM)
+               "$SUDOCMD" chmod a+x "$1"
+               "$SUDOCMD" chown "$("$SUDOCMD" id -u):$("$SUDOCMD" id -g)" "$1"
+       else
+               # Else the user is the owner (APPMAN)
+               chmod a+x "$1"
+               chown "$(id -u):$(id -g)" "$1"
+       fi
+}
+
 _checksum_on_standard_dependencies() {
 	if [ -f "$CACHEDIR"/AMCACHEPATH/"$name" ]; then
 		dep_sum256_local=$(sha256sum "$CACHEDIR"/AMCACHEPATH/"$name" 2>/dev/null | awk '{print $1}')
 		dep_sum256_host=$(sort "$AMCACHEDIR/$ARCH.sha256sum.txt" | grep "$name-$ARCH-static$" 2>/dev/null | awk '{print $1}')
 		if [ "$dep_sum256_local" != "$dep_sum256_host" ]; then
-			rm -Rf "$CACHEDIR"/AMCACHEPATH/"$name"
-			_deps_dl && chmod a+x "$CACHEDIR"/AMCACHEPATH/"$name" && printf $" - %b, downloaded from https://github.com/ivan-hc/am-utils\n" "$name"
+			"$SUDOCMD" rm -Rf "$CACHEDIR"/AMCACHEPATH/"$name"
+			_deps_dl && _set_access_privilege "$CACHEDIR"/AMCACHEPATH/"$name" && printf $" - %b, downloaded from https://github.com/ivan-hc/am-utils\n" "$name"
 		fi
 	fi
 }
@@ -536,13 +548,14 @@ _deps_dl() {
 
 if [ "$ARCH" = aarch64 ] || [ "$ARCH" = x86_64 ]; then
 	mkdir -p "$CACHEDIR"/AMCACHEPATH
+	_set_access_privilege "$CACHEDIR"/AMCACHEPATH
 	# Download static binaries if commands are not available
 	amutils="$AM_standard_dependencies"
 	AM_UTILS_SOURCE="https://github.com/ivan-hc/am-utils/releases/download/continuous-$ARCH"
 	for name in $amutils; do
 		if ! command -v "$name" >/dev/null 2>&1; then
 			if [ -d "$CACHEDIR"/AMCACHEPATH/"$name" ] || [ -L "$CACHEDIR"/AMCACHEPATH/"$name" ] || [ ! -f "$CACHEDIR"/AMCACHEPATH/"$name" ]; then
-				rm -Rf "$CACHEDIR"/AMCACHEPATH/"$name"
+				"$SUDOCMD" rm -Rf "$CACHEDIR"/AMCACHEPATH/"$name"
 				if [ -z "$fallback_binaries_needed" ]; then
 					fallback_binaries_needed="$name"
 				else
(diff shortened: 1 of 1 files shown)


ivan-hc:
I use SUDOCMD "unquoted" because in appman SUDOCMD is empty, and the result would be `"" command`

ivan-hc:
Are you sure that there is no alternative to this?

vishnu350:
I doubt there is a non-complicated way to lock file/folder write access without `sudo chown`. Since AM is a root application, our options are limited.

ivan-hc:
Can you explain step-by-step how an attacker would hit a binary our workflow even if we do checksum verification?

I'm a total ignorant... but I'd like to analyze alternative solutions to this.

vishnu350:
1. Wait for the user to execute `am -i some_app`.
2. Wait for n seconds for checksum verification to finish.
3. Overwrite the binary file with a fake one (eg. tar).
4. When some_app.tar.gz download completes, it will execute this fake tar as sudo.
5. Now the fake tar binary has root access.

ivan-hc:
> 1. Wait for the user to execute `am -i some_app`.
> 
>  2. Wait for n seconds for checksum verification to finish.
> 
>  3. Overwrite the binary file with a fake one (eg. tar).
> 
>  4. When some_app.tar.gz download completes, it will execute this fake tar as sudo.
> 
>  5. Now the fake tar binary has root access.

Uhm... this means that the local binaries should be removed/re-downloaded between an installation and another?

ivan-hc:
@vishnu350 I think that if this issue is only for apps installed with root permissions, trough AM, the change you say have to be done in install.am, not in APP-MANAGER.

Do you know why?

Because if the script have to be executed as root, while password is inserted, permissions on the custom PATH have to be changed temporarily, and restored when the installation process is finished/interrupted.

vishnu350:
Downloading/re-downloading is the same as the current checksum solution. A determined attacker will just time the waiting period and then overwrite.

We cannot mitigate attacks 100%. We can only reduce the attack surface. That's why I said install only once, and then lock it.

The attacker only has one chance if you lock the dir. If you rely on the current implementation, the attacker has unlimited chances.

ivan-hc:
@vishnu350 let focus on this function
```
_install_exec_installation_script() {
	if [ "$(ls -A "$CACHEDIR"/AMCACHEPATH 2>/dev/null)" ] && [ -n "$SUDOCMD" ]; then
		$SUDOCMD PATH="$PATH" ./"$arg"
	else
		$SUDOCMD ./"$arg"
	fi
}
```
what we can do is to
1. run _checksum_on_standard_dependencies as root
2. run the installation script as root
3. restore privileges on the custom PATH

All this need to be TRAP, so in case the process is interrupted, privileges are restored also if step 1 and 2 are incomplete.

ivan-hc:
> @vishnu350 let focus on this function
> 
> ```
> _install_exec_installation_script() {
> 	if [ "$(ls -A "$CACHEDIR"/AMCACHEPATH 2>/dev/null)" ] && [ -n "$SUDOCMD" ]; then
> 		$SUDOCMD PATH="$PATH" ./"$arg"
> 	else
> 		$SUDOCMD ./"$arg"
> 	fi
> }
> ```
> 
> what we can do is to
> 
>     1. run _checksum_on_standard_dependencies as root
> 
>     2. run the installation script as root
> 
>     3. restore privileges on the custom PATH
> 
> 
> All this need to be TRAP, so in case the process is interrupted, privileges are restored also if step 1 and 2 are incomplete.

Something like this
```
_install_exec_installation_script() {
	if [ "$(ls -A "$CACHEDIR"/AMCACHEPATH 2>/dev/null)" ] && [ -n "$SUDOCMD" ]; then
		$SUDOCMD _checksum_on_standard_dependencies && $SUDOCMD PATH="$PATH" ./"$arg" && $SUDOCMD chown -R "$USER" "$CACHEDIR"/AMCACHEPATH \
		|| $SUDOCMD chown -R "$USER" "$CACHEDIR"/AMCACHEPATH
	else
		$SUDOCMD ./"$arg"
	fi
}
```

Its just an example, a draft... I need to perfect it.

vishnu350:
OK, its a start. But before that, you also need to consider another issue. 

I don't want to alarm you. But you need to know that /opt/am is also unlocked, currently. It is equally vulnerable.

ivan-hc:
> OK, its a start. But before that, you also need to consider another issue.
> 
> I don't want to alarm you. But you need to know that /opt/am is also unlocked, currently. It is equally vulnerable.

This is a known issue. At least from version 6.6.1. I don't plan to fix it (yet).

Keep focus on this issue or we will not go nowhere.

vishnu350:
Because the problem (and solution) will ultimately be the same for both, that's why I am bringing this up.

> what we can do is to
> 
>     1. run _checksum_on_standard_dependencies as root
> 
>     2. run the installation script as root
> 
>     3. restore privileges on the custom PATH

Consider:
3. restore user privileges upon switching to APPMAN (_appman() restores AMCACHEPATH and /opt/am).
4. enforce root privilege upon switching to AM (_am() locks AMCACHEPATH and /opt/am)

Most users will not switch back and forth, so it is really OK to password prompt annoy the small group of users that do.

ivan-hc:
@fiftydinar can you help?

ivan-hc:
Folks this seems to work well

```
_install_exec_installation_script() {
	if [ "$(ls -A "$CACHEDIR"/AMCACHEPATH 2>/dev/null)" ] && [ -n "$SUDOCMD" ]; then
		$SUDOCMD chmod a-w "$CACHEDIR"/AMCACHEPATH
		trap '$SUDOCMD chmod u+w "$CACHEDIR"/AMCACHEPATH' EXIT INT TERM
		$SUDOCMD PATH="$PATH" ./"$arg"
	else
		$SUDOCMD ./"$arg"
	fi
}
```

The directory become read-only while installing the app.

ivan-hc:
I have closed this, but we can keep talking here about this, or at https://github.com/ivan-hc/AM/pull/2318 directly.

vishnu350:
The primary question: **Is AM a root application (/root/.cache) or is it a user application (/home/user/.cache)**? It is incredibly difficult to be both, no matter how hard you try.

There will be plenty of holes just waiting to be found if a confused root app tries to keep its settings/binaries in the user's homedir. You will be playing [whack-a-mole](https://www.youtube.com/watch?v=TZdPNt1_nbE) closing the holes. That is why Debian's apt only works as root by default.

ivan-hc:
When I created it, my goal was to have applications installed at the system level, but updateable without root privileges.

In this sense, AM belongs to a single user. Other users or system administrators have no access to the app directories if they're installed by another administrator. Specifically, /opt/am have one owner. The one that have installed it.

AM is for an administrator (one); AppMan is for all non-privileged users.

All I can do is provide the choice between the two.

PR #2364, opened 2026-05-16
Added new test-am-utils.sh and other regression testing updates

vishnu350:
Updates:
- Added new test-am-utils.sh, a dependency test to rely on am-utils as fallback.
- Resolved all dependencies for Arch and Debian Jessie containers.
- Improved keypress printf readability + cleanups.
- Added four new test related functions: 
  - Function to hide/restore libfuse.so.2: _hide_libfuse2()/_restore_libfuse2()
  - Function to hide/restore a list binaries: _hide_binaries()/_restore_binaries()
  - Function to test if an app was properly installed: _test_apps()
- Used test_apps() in all tests where appropriate.
- Improved nolibfuse test to hide lifuse.so.2 if it exists.

@ivan-hc Please take a look at test-am-utils.sh, it tests am-utils by hiding these binaries: **7z ar md5sum sha1sum sha256sum sha512sum tar unzip wget xz xzcat**. I couldn't add everything because stuff like **mv** just cannot be removed without breaking the system.

Regression is passing in all containers. AM utils binaries work even in alpine(musl) and Debian Jessie.

### DIFF
--- a/regress/am-alpine.dockerfile
+++ b/regress/am-alpine.dockerfile
@@ -17,7 +17,7 @@ ENV LANG en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 
 # Setup AM with safe defaults
-RUN printf "y\n\n" | am --user && am --system && am --disable-notifications
+RUN printf "Y\n\n" | am --user && am --system && am --disable-notifications
 
 # Setup env
 RUN echo "export PATH=$PATH:/root/.local/bin" >> ~/.bashrc
--- a/regress/am-arch.dockerfile
+++ b/regress/am-arch.dockerfile
@@ -6,7 +6,7 @@
 FROM archlinux:latest
 
 # Install dependencies and AM
-RUN pacman-key --init && pacman -Sy && pacman -Su --noconfirm sudo wget curl less git glibc fuse3 file unzip xz 7zip libnotify
+RUN pacman-key --init && pacman -Sy && pacman -Su --noconfirm sudo wget curl less git glibc fuse3 file unzip xz 7zip libnotify which
 RUN cd && wget https://raw.githubusercontent.com/ivan-hc/AM/main/INSTALL && chmod a+x ./INSTALL && sudo ./INSTALL && rm ./INSTALL
 
 # Copy regression folder
--- a/regress/am-arch.dockerfile
+++ b/regress/am-arch.dockerfile
@@ -18,7 +18,7 @@ RUN locale-gen && echo "LANG=en_US.UTF-8" > /etc/locale.conf
 ENV LC_ALL en_US.UTF-8
 
 # Setup AM with safe defaults
-RUN printf "y\n\n" | am --user && am --system && am --disable-notifications
+RUN printf "Y\n\n" | am --user && am --system && am --disable-notifications
 
 # Setup env
 RUN echo "export PATH=$PATH:/root/.local/bin" >> ~/.bashrc
--- a/regress/am-debian-old.dockerfile
+++ b/regress/am-debian-old.dockerfile
@@ -21,7 +21,7 @@ RUN locale-gen && update-locale LANG=en_US.UTF-8
 ENV LC_ALL en_US.UTF-8
 
 # Setup AM with safe defaults
-RUN printf "y\n\n" | am --user && am --system && am --disable-notifications
+RUN printf "Y\n\n" | am --user && am --system && am --disable-notifications
 
 # Setup env
 RUN echo "export PATH=$PATH:/root/.local/bin" >> ~/.bashrc
--- a/regress/am-debian-old.dockerfile
+++ b/regress/am-debian-old.dockerfile
@@ -5,12 +5,11 @@
 # Use the official Debian Jessie image as a parent image
 FROM debian:jessie
 
-# Install dependencies and AM
-# Notes: 7zip not available in Jessie, and GPG warnings is normal, repos are archieved
+# Install dependencies and AM (GPG warnings are normal as repos are archieved)
 RUN echo 'deb http://archive.debian.org/debian jessie main contrib non-free' > /etc/apt/sources.list
 RUN echo 'deb http://archive.debian.org/debian-security jessie/updates main contrib non-free' >> /etc/apt/sources.list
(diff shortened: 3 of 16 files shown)


vishnu350:
@ivan-hc Even though the test is passing, AM seems to be re-downloading am-utils binaries after the first app install (due to checksum failure). The newly downloaded static binaries work fine afterwards, confirmed via the app install scripts. This issue is seen in all containers.

https://github.com/user-attachments/assets/b1addab7-f41e-47f3-aa63-b55cb3f74216

This issue does not occur in --user mode (APPMAN), which is the second half of the test. Please check test-am-utils.sh for more details.

ivan-hc:
Yes, I'm able to reproduce the issue.

I added a this check to collect sha256sum old value before installation starts

https://github.com/ivan-hc/AM/blob/67eca84c64400291cf1c77232c50ba1855f13221/modules/install.am#L277-L288

Even if I have `sha256sum` installed with `coreutils`, this check seems to fail

https://github.com/ivan-hc/AM/blob/67eca84c64400291cf1c77232c50ba1855f13221/modules/install.am#L288-L338

ivan-hc:
I have found the issue

ivan-hc:
Try now https://github.com/ivan-hc/AM/commit/8e1a223ee3b94be941a87c5eb877ddb1c75ee412

ivan-hc:
@vishnu350 I see "auto-verified", this means that you are testing installation scripts for apps in archive. What about AppImages?

ivan-hc:
Another thing, even if `wget` is downloaded, I see the `curl` progress bar. Strange.

vishnu350:
> Try now 8e1a223

OK, looks good now.

> Another thing, even if wget is downloaded, I see the curl progress bar. Strange.

But I'm sure the other am-utils static binaries work correctly, eg. **tar** properly extracts the apps even though it is removed/hidden.

> @vishnu350 I see "auto-verified", this means that you are testing installation scripts for apps in archive. What about AppImages?

It does install one Appimage: **$app_name4**.

The goal for this test is to make sure am-utils binaries are working correctly, so it mostly installs archives because they need tar, ar, xz. The test then checks for **checksum.\*verified** msg 4 times (for all 4 apps), because it means the checksum binaries worked.

The apps will be randomly selected from these lists in test-common.sh:

```
# List of test apps with zsync files
TEST_APP_LIST_ZSYNC="zsync2 xeyes rofi sas aisap clagrange"

# List of test apps with only digest files
TEST_APP_LIST_DIG="conky" #Too big: joplin dinox

# List of test apps that are zipped
TEST_APP_LIST_ZIP="clifm gotimer dra nyan navi lsd hyperfine fcp"

# List of test apps that cant be verified
TEST_APP_LIST_NOCHK="bench-cli helio appimagen crabfetch colorstatic-bash"

# List of test apps that are outdated
TEST_APP_LIST_OLD="aisap bench-cli colorstatic-bash"

# List of test apps that can be downgraded
TEST_APP_LIST_DOWN="clifm aisap fcp zsync2"

# List of test apps that use tar.xz
TEST_APP_LIST_XZ="boxxy onefetch spotifetch lovesay"
```

vishnu350:
@ivan-hc, the issue seems resolved for debian/ubuntu/alpine. But it still occurs in arch:latest.
```
============================================================================

                  START OF ALL INSTALLATION PROCESSES

============================================================================

 ◆ "7ZIP": starting installation script

 - 7z, downloaded from https://github.com/ivan-hc/am-utils

 - ar, downloaded from https://github.com/ivan-hc/am-utils

 - md5sum, downloaded from https://github.com/ivan-hc/am-utils

 - sha1sum, downloaded from https://github.com/ivan-hc/am-utils

 - sha256sum, downloaded from https://github.com/ivan-hc/am-utils

 - sha512sum, downloaded from https://github.com/ivan-hc/am-utils

 - tar, downloaded from https://github.com/ivan-hc/am-utils

 - unxz, downloaded from https://github.com/ivan-hc/am-utils

 - unzip, downloaded from https://github.com/ivan-hc/am-utils

 - wget, downloaded from https://github.com/ivan-hc/am-utils

 - xz, downloaded from https://github.com/ivan-hc/am-utils

 - xzcat, downloaded from https://github.com/ivan-hc/am-utils

##################################################################### 100.0%
Downloaded 1572208 bytes in 2.348788 seconds

 ✔ Checksum auto-verified.

 "7ZIP" INSTALLED (7 MB OF DISK SPACE)
____________________________________________________________________________
============================================================================

                  END OF ALL INSTALLATION PROCESSES

             The following new programs have been installed:

 ◆ 7zip 26.01✓

============================================================================
```

And yeah, it still seems to be using curl for some reason.

ivan-hc:
Well, `curl` may be in use because there is no `wget` or `wget` is a symlink to `wget2`

Are you using the latest release of AM?

vishnu350:
> Are you using the latest release of AM?

Always :)

This issue is only for Arch. Its a little strange, as it doesn't say "checksum does not match, re-downloading...".

ivan-hc:
> > Are you using the latest release of AM?
> 
> Always :)
> 
> This issue is only for Arch. Its a little strange, as it doesn't say "checksum does not match, re-downloading...".

This means that it cannot export PATH

https://github.com/ivan-hc/AM/blob/84e55d896a5b7df5ea58fd67c01ef721eaca9e7f/modules/install.am#L323-L332

The above is the second reference of _use_online_downloader_as_root, used if command in PATH is missing.

The first is few lines above

https://github.com/ivan-hc/AM/blob/84e55d896a5b7df5ea58fd67c01ef721eaca9e7f/modules/install.am#L312-L320

Maybe that
```
elif [ -z "$test_dep_sum256_old" ]; then
```
...has to do with this.

ivan-hc:
> Maybe that
> 
> ```
> elif [ -z "$test_dep_sum256_old" ]; then
> ```
> 
> ...has to do with this.

The above is used to verify if files in PATH were already detected and checked before. If not, they are not allowed.

vishnu350:
The arch re-download issue was resolved by just using a fresh AM-Arch container for the am-utils test.

I must have damaged my Arch container while preparing the am-utils test, because of hiding core binaries too many times (mv/wget/curl/etc), and then sometimes forget to restore them.

Apologies for the confusion.

ivan-hc:
np

PR #2365, opened 2026-05-17
remove palemoon

Samueru-sama:
@ivan-hc The palemoon repo has been deleted following a request from upstream as they pointed out the palemoon license explicitly does not allow appimage. 😖


Samueru-sama:
cc @davidhedlund Your feedback is greatly appreciated here. 

We were contacted by palemoon as their terms of service state [that AppImage redistribution is explicitly not allowed](https://www.palemoon.org/redist.shtml)

Following the message, we immediately removed the palemoon appimage repostiory (they actually wanted us to just take down the released binaries but anyways, my bad I did not read the original issue correctly).

---

What's odd is that the reasoning they have this restriction in their terms of service is because according to them it is not possible to mix GPL and MPL software.

This is the first time I heard such statement, and even more crazy considering that firefox officially ships a snap for ubuntu that bundles GPL software like ffmpeg. Note it doesn't link statically ffmpeg just in case.

We also never shipped additional libraries in the palemoon appimage. Just what they already shipped in their released binaries.

davidhedlund:
> cc @davidhedlund Your feedback is greatly appreciated here.

Thank you for your inquiry. I emailed you how to get in touch with the Free Software Foundation to get professional assistance.

ivan-hc:
> We also never shipped additional libraries in the palemoon appimage. Just what they already shipped in their released binaries.

:point_up:  **This** :point_up: 

If we can't redistribute their package as an AppImage, then we can't redistribute their official package either.

We haven't added anything except the launcher, the icon, and the AppRun, which are the three basic elements of an AppImage.

In essence, we were redistributing their portable package, in a different archive. Nothing more.

But that's okay, I hope they'll understand. Until then, no Palemoon in this database.

ivan-hc:
According with:

> 3. The binaries and/or archives are completely UNALTERED. This includes addition or removal of any component of the browser or redistributing modified parts of the browser as-built (including, without limitation, predefined profile components, altered jar or omni files, or modified .exe or dll components). Redistribution is therefore not permitted if, without limitation: 
>
> [...]
>
> b. The binaries are packaged in "bundle" archive formats like AppImage, flatpak, SNAP or similar formats that include any other libraries or files not in the original package.*

That:

> **that include any other libraries or files not in the original package.**

We were right. No third-library was added, then no license conflicts or anything.

The no way the AppImage was against their rules.

PR #2366, opened 2026-05-17
Add checksum change detection to update summary table

cheack:
The update summary included apps whose version didn't actually change. This happened because the raw file comparison included ANSI color codes and checksum markers (✓/✖) - if only the checksum status changed after update, the app still appeared as "updated" with no explanation.

image

Extended the `_update_print_updated_apps_table` function to detect when an app's checksum marker changed but the version stayed the same. Such apps now appear in the table with a `(checksum changed)` note, making it clear the version didn't actually change:

```
 The following apps have been updated:

     App        Previous  Current

  1. anydesk    8.0.2     8.0.2     (checksum changed)
  2. blender    5.1.0     5.1.1
  3. ocenaudio  3.18.2-1  3.18.2-1  (checksum changed)
```


ivan-hc:
I was looking for something like this, thank you

PR #2373, opened 2026-05-20
Resolved incorrect invocation of curl when am-utils wget is available

vishnu350:
Fix for curl issue shown in #2364.

The problem is because wget was only checked in the beginning, before am-utils $PATH has been added.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -114,24 +114,24 @@ MODULES_SOURCE="$AMREPO/modules"
 CURLCMD="curl"
 WGETCMD="wget"
 
+# Check for essential commands
+if ! command -v wget >/dev/null 2>&1 && ! command -v curl >/dev/null 2>&1; then
+	printf $" ${RED}💀 ERROR! MISSING ESSENTIAL COMMANDS\033[0m: %s\n\n Install the above and try again! \n" "$CURLCMD, $WGETCMD"
+	exit 0
+fi
+
 # Export variables for such commands to avoid confusion
 export CURLLS="$CURLCMD -Ls" # This will be added in variables
 export CURLPROGRES="$CURLCMD -LJO --progress-bar --write-out 'Downloaded %{size_download} bytes in %{time_total} seconds'"
 export WGETPROGRESS="$WGETCMD -q --no-verbose --show-progress --progress=bar"
 
-# Check if essential commands are available
-_check_essential_commands() {
-	if ! command -v wget >/dev/null 2>&1 && ! command -v curl >/dev/null 2>&1; then
-		printf $" ${RED}💀 ERROR! MISSING ESSENTIAL COMMANDS\033[0m: %s\n\n Install the above and try again! \n" "$CURLCMD, $WGETCMD"
-		exit 0
-	fi
+# Check wget version
+_check_wget_version() {
 	if command -v wget >/dev/null 2>&1 && wget --version 2>/dev/null | head -1 | grep -qi "$WGETCMD.* 1." ; then
 		WGETV="1"
 	fi
 }
-
-# Check for essential commands first before running anything else
-_check_essential_commands
+_check_wget_version
 
 # Identify objects online
 _use_online_check_tool() {
@@ -684,8 +684,8 @@ if [ "$ARCH" = aarch64 ] || [ "$ARCH" = x86_64 ]; then
 			done
 			wait
 			export PATH="$CACHEDIR"/AMCACHEPATH/:"${PATH}"
-			# Check for essential commands (again but with new $PATH)
-			_check_essential_commands
+			# Check wget version again but with new $PATH
+			_check_wget_version
 		fi
 	fi
 fi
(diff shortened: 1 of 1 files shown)


ivan-hc:
No, wait! You added the check after the functions needed to check online connection, which is made by alternating `curl` and `wget.`

ivan-hc:
The check is also done first also to download locale files.

vishnu350:
Should I move it up? The code was a little messy on the top, so I moved it down.

ivan-hc:
It is up to give the guarantee that one between `curl` and `wget` will be used for other core functions (for connection and downloads).

ivan-hc:
Such commands are two different things
```
if ! command -v wget >/dev/null 2>&1 && ! command -v curl >/dev/null 2>&1; then
	printf $" ${RED}💀 ERROR! MISSING ESSENTIAL COMMANDS\033[0m: %s\n\n Install the above and try again! \n" "$CURLCMD, $WGETCMD"
	exit 0
fi
```
it is what I said above, while
```
if command -v wget >/dev/null 2>&1 && wget --version 2>/dev/null | head -1 | grep -qi "$WGETCMD.* 1." ; then
	WGETV="1"
fi
```
is needed for other things, like replacing `wget` in the installation scripts if AM is running on Fedora or other systems that use a fake `wget`

vishnu350:
You are right. Let me fix it.

ivan-hc:
NOTE that they must be out of the function to be used in AM globally, not in few moments.

ivan-hc:
> NOTE that they must be out of the function to be used in AM globally, not in few moments.

Do not use the function. Just do it directly.

APP-MANAGER
```diff
@@ -124,9 +125,13 @@ export CURLLS="$CURLCMD -Ls" # This will be added in variables
 export CURLPROGRES="$CURLCMD -LJO --progress-bar --write-out 'Downloaded %{size_download} bytes in %{time_total} seconds'"
 export WGETPROGRESS="$WGETCMD -q --no-verbose --show-progress --progress=bar"
 
-if command -v wget >/dev/null 2>&1 && wget --version 2>/dev/null | head -1 | grep -qi "$WGETCMD.* 1." ; then
-	WGETV="1"
-fi
+# Check wget version
```
ivan-hc, line 128: I have said that this must be global, not in function

vishnu350: It runs globally here (line 134): _check_wget_version

I use a function because it needs to be called again in line 688. Just in case, so if we change this wget check again in the future, the bug will not reappear (since it is a difficult to catch bug).

ivan-hc:
@vishnu350 I have executed AM installations on a VM of Arch Linux that don't have `7z` and `unzip`, and repeated the test without `wget`. Installed apps were
```
zsync2 xeyes rofi sas aisap clagrange
```
Like the ones in your container.

I'm not able to reproduce the issue. Maybe the issue is the container itself.

Anyway, in the test without `wget` and the latter downloaded as a static binary, AM uses `curl` in the installation scripts, not the static `wget`.

vishnu350:
> Anyway, in the test without wget and the latter downloaded as a static binary, AM uses curl in the installation scripts, not the static wget.

Yes, this is what I am fixing with this PR. It will now use **am-utils wget** correctly instead of curl:
```
 The following commands are missing and can be temporarily downloaded from https://github.com/ivan-hc/am-utils as static binaries:

 7z ar md5sum sha1sum sha256sum sha512sum tar unxz unzip wget xz xzcat

 Press "Y" to store them in cache until you clear it (with "-c" option), or press "N" to ignore this message and install them manually (default).

 - unzip, downloaded from https://github.com/ivan-hc/am-utils
 - unxz, downloaded from https://github.com/ivan-hc/am-utils
 - xz, downloaded from https://github.com/ivan-hc/am-utils
 - xzcat, downloaded from https://github.com/ivan-hc/am-utils
 - ar, downloaded from https://github.com/ivan-hc/am-utils
 - sha256sum, downloaded from https://github.com/ivan-hc/am-utils
 - md5sum, downloaded from https://github.com/ivan-hc/am-utils
 - 7z, downloaded from https://github.com/ivan-hc/am-utils
 - wget, downloaded from https://github.com/ivan-hc/am-utils
 - tar, downloaded from https://github.com/ivan-hc/am-utils
 - sha1sum, downloaded from https://github.com/ivan-hc/am-utils
 - sha512sum, downloaded from https://github.com/ivan-hc/am-utils

----------------------------------------------------------------------

 YOU HAVE INSTALLED 1 PROGRAM MANAGED BY "AM"

 - APPNAME | VERSION | TYPE        | SIZE
 - ------- | ------- | ----        | ----
 ◆ am      | 10.2-2  | bash-script | 234.96 KiB

 TOTAL SIZE: 234.96 KiB of disk space in use

----------------------------------------------------------------------
"AM" 10.2-2: DEVELOPER MODE
----------------------------------------------------------------------
Installing all apps in system mode...
spotifetch-x86_64-unknown-linux-gnu.tar.xz          100%[========================>]   1.66M  --.-KB/s    in 0.03s

```


> I'm not able to reproduce the issue. Maybe the issue is the container itself.

I don't have that issue anymore as well. I just reset the arch container and the issue went away. Maybe too much testing hiding core binaries like mv/wget/curl broke it.

ivan-hc:
Yes, tested, now it works

PR #2378, opened 2026-05-21
Extend fuzzy suggest to "-ia" and "-ias"

ivan-hc:
@cheack this could be done better


ivan-hc:
For some strange reason, with `-ia` and `-ias` static binaries in custom PATH cannot be removed.

PR #2382, opened 2026-05-22
Add --restore-desktop option to reset a modified .desktop file

### LINKED ISSUE #2253: Make it possible to restore .desktop files
# Premise
By opening this issue, I certify:
- that I have updated "AM"/"AppMan" to the latest version
- that I have read the README https://github.com/ivan-hc/AM/blob/main/README.md
- that I have read the documentation https://github.com/ivan-hc/AM/tree/main/docs
- that I have searched for an option or flag in `-h` that covers my use case
- that I have searched for an issue that covers my use case https://github.com/ivan-hc/AM/issues

--------------------------------------------------

# Is your feature request related to a problem? Please describe.
.desktop files are sometimes modified—for instance, when a user wants to add extra text to the Name= field to distinguish one application from others that use the same name. For example `Name=Emacs (appman)`.

# Describe the solution you'd like
It would be useful to have an option to reset the .desktop files without needing to reinstall the application. Currently, no copies of the .desktop files are stored when they’re generated in ~/.local/share/applications/. As an alternative, the original files could be archived when copied there, enabling manual restoration instead of relying on reinstallation—or a dedicated repair feature could be added for this purpose.

# Describe alternatives you've considered
Write here

andy5995:
Closes: #2253

When a user (or the app itself) modifies an installed app's .desktop launcher — for example by appending a custom Name= to distinguish two parallel installs — there was no way to roll the change back short of reinstalling. Add a snapshot-and-restore flow:

* modules/install.am: in _post_installation_processes, after the local Exec= patching block, snapshot the live .desktop to $APPDIR/.am-installer/$arg-AM.desktop.orig. Capturing it AFTER the install script's post-processing (e.g. google-chrome's StartupWMClass fix) means restore brings back what AM intended, not the raw upstream .desktop from inside the AppImage.

* APP-MANAGER: route `restore-desktop` and `--restore-desktop` to management.am; add both forms to the completion list.

* modules/management.am: add _restore_desktop() and a matching case branch. The live .desktop path is read from the app's `remove` script (it always contains an absolute path matching "*AM.desktop"), matching the technique sketched in the issue thread. Attempts an unprivileged copy first, falls back to $SUDOCMD for system-wide installs, then refreshes the desktop database if available. Clear errors for: not a managed install, no snapshot (points the user at -i to reinstall), or unknown .desktop path.

* docs/guides-and-tutorials/launcher.md: new section documenting the option, including the post-install-patch caveat and the migration note for apps installed before this feature existed.

Tested in a clean debian:stable-slim container: real-flow restore via `am --restore-desktop test-app` rolls back user edits to the snapshot; the snapshot-missing error path fires correctly through the dispatcher for fixtures simulating pre-feature installs.

### DIFF
--- a/modules/management.am
+++ b/modules/management.am
@@ -389,6 +389,46 @@ _launcher(){
 }
 
 ################################################################################################################################################################
+#				RESTORE-DESKTOP
+################################################################################################################################################################
+
+# Restore an app's .desktop file from the snapshot saved at install time.
+# The snapshot is created by _post_installation_processes in install.am and
+# lives at $argpath/.am-installer/$arg-AM.desktop.orig. The live .desktop
+# path is read from the app's `remove` script (it always contains an
+# absolute path matching "*AM.desktop").
+_restore_desktop() {
+	backup="$argpath/.am-installer/$arg-AM.desktop.orig"
+	if [ ! -f "$argpath"/remove ]; then
+		printf $" 💀 ERROR: \"%s\" is not a managed install (no remove script).\n" "$arg"
+		return 1
+	fi
+	if [ ! -f "$backup" ]; then
+		printf $" 💀 ERROR: no .desktop snapshot for \"%s\".\n" "$arg"
+		printf $"    Snapshots are taken at install time; reinstall with \"%s -i %s\" to create one.\n" "$AMCLI" "$arg"
+		return 1
+	fi
+	desktop_path=$(grep -Eo "/[^[:space:]\"']*AM\.desktop" "$argpath"/remove 2>/dev/null | head -1)
+	if [ -z "$desktop_path" ]; then
+		printf $" 💀 ERROR: could not determine the .desktop path for \"%s\".\n" "$arg"
+		return 1
+	fi
+	# System paths (/usr/local/share/applications) need root; user paths
+	# (~/.local/share/applications) do not. Try unprivileged first.
+	if cp "$backup" "$desktop_path" 2>/dev/null \
+		|| $SUDOCMD cp "$backup" "$desktop_path" 2>/dev/null; then
+		printf $" ✔ Restored \"%s\" -> %s\n" "$arg" "$desktop_path"
+		if command -v update-desktop-database >/dev/null 2>&1; then
+			update-desktop-database "$(dirname "$desktop_path")" >/dev/null 2>&1 \
+				|| $SUDOCMD update-desktop-database "$(dirname "$desktop_path")" >/dev/null 2>&1
+		fi
+	else
+		printf $" 💀 ERROR: failed to write %s\n" "$desktop_path"
+		return 1
+	fi
+}
+
+################################################################################################################################################################
 #				LOCK/UNLOCK
 ################################################################################################################################################################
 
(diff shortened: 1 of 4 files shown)


andy5995:
@ivan-hc I haven't manually reviewed all the code yet. As noted in the comments above, I did some testing. Do you want to try this yourself, as you are more familiar with how AM works or would you like me to manually review the code more and get back to you?

modules/management.am
```diff
@@ -388,6 +388,46 @@ _launcher(){
 	fi
 }
 
+################################################################################################################################################################
+#				RESTORE-DESKTOP
+################################################################################################################################################################
+
+# Restore an app's .desktop file from the snapshot saved at install time.
+# The snapshot is created by _post_installation_processes in install.am and
+# lives at $argpath/.am-installer/$arg-AM.desktop.orig. The live .desktop
```
ivan-hc, line 397: The .am-installer directory is only for installation scripts.

I don't know if adding other files would affect the goal.

For example, in `-a` it is used to detect the real name of an app, for example, if it come from a third-party source or is not in the AM database.

ivan-hc:
I would not add a new option, but I would like to improve an existing one by adding "flags", so we don't need to add too much code in many modules and we don't need to rewrite lines that may be already used in other options or creating backup files anywhere.

For all this, the solution would be the existing option `reinstall`, but with a flag `--launcher` (for now used as a standalone option).

The way .desktop files are created may vary, but we can get the more common ones from the installation script directly.

In the end, all we need to do will be to add this usage in the documentation.

ivan-hc:
For example

		'--launcher')
			echo "$DIVIDING_LINE"
			_determine_args
			ARGS=$(echo "$ARGS" | tr ' ' '\n' | sort -u | xargs)
			for arg in $ARGS; do
				argpath_items=$(echo "$ARGPATHS" | grep "/$arg$" | xargs)
				for argpath in $argpath_items; do
					if [ -f "$argpath"/"$arg" ] && grep -Eaoq -m 1 "appimage-extract" "$argpath"/"$arg" || file "$argpath"/"$arg" | grep -q 'static'; then
 						mkdir -p "$SCRIPTDIR"/AMLAUNCHER_DIR
						exec 3< <("$argpath"/"$arg" --appimage-mount)
						read -r MOUNT_POINT <&3
						APP_PID=$!
						export MOUNT_POINT
						cp "$MOUNT_POINT"/*.desktop "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop
						COUNT=0
						while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
							if [ -L "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop ]; then
								LINKPATH="$(readlink "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop | sed 's|^\./||' 2>/dev/null)"
								cp "$MOUNT_POINT"/"$LINKPATH" "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop
							fi
							[ ! -L "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop ] && break
							COUNT=$((COUNT + 1))
						done
						[ -f "$SCRIPTDIR"/AMLAUNCHER_DIR/"$arg"-AM.desktop ] && echo " - $SCRIPTDIR/AMLAUNCHER_DIR/$arg-AM.desktop"
						kill "$APP_PID" 2>/dev/null
					fi
				done
			done
			echo "$DIVIDING_LINE"
			;;
			
The above extracts and saves all .desktop files from all AppImages installed

https://github.com/user-attachments/assets/c0589dfd-be27-41e8-a17f-7425dfb11e28

This action is a prof of concept. We here check if the file is an AppImage, and then uses `--appimage-mount` (not `--appimage-extract`) to search and save the .desktop file.

The above also works with sharun-based AppImages on ZFS file systems.

ivan-hc:
done here https://github.com/ivan-hc/AM/commit/dde9475ecefe33dbb6385734c7555156da1be5a8

ivan-hc:
Thank you @andy5995

PR #2384, opened 2026-05-23
am2pla-site: emit categories list as a CSS grid of pill-links

andy5995:
_categories_buttons() previously emitted a single line of bold-italic markdown links separated by " - ". On a wide page that's a long wrapping run; on a narrow viewport it's worse. Switch to raw HTML:

  <div class="cat-grid">
    <a class="cat-pill cat-pill--all" href="appimages.html">AppImages</a>
    <a class="cat-pill" href="am-utils.html">am-utils</a>
    ...
  </div>

CSS lives in Portable-Linux-Apps.github.io's assets/css/style.scss (.cat-grid uses grid-template-columns: repeat(auto-fit, ...) for responsive reflow; .cat-pill--all gives AppImages an accent background and bold weight so it stands out). Light/dark variants mirror the existing theme switch in that file.

Order is alphabetical (AppImages first as the meta-category, then the CATEGORIES list which is already alphabetical). Applies in every place that calls _categories_buttons: index.md, apps.md, appimages.md, and each category page.

## Desktop

image

## Mobile

image

### DIFF
--- a/tools/am2pla-site
+++ b/tools/am2pla-site
@@ -84,7 +84,19 @@ function _table_head() {
 }
 
 function _categories_buttons() {
-	printf "\n#### *Categories:* ***[AppImages](appimages.md)***%b\n\n" "$(for c in $CATEGORIES; do printf " - ***[$c]($c.md)***"; done)"
+	# Raw HTML grid; styles live in assets/css/style.scss on the website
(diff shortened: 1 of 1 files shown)


andy5995:
Depends on https://github.com/Portable-Linux-Apps/Portable-Linux-Apps.github.io/pull/149

PR #2385, opened 2026-05-23
Fixed _check_version_filters() in APP-MANAGER to be more accurate & added new test-am-versions.sh to validate it

vishnu350:
Regression testing updates and _check_version_filters() fixes :
- Added new test-am-versions.sh to check known app download strings against awk/mawk/gawk/goawk/original-awk to see if the expected version number is correct. This will prevent regressions in AM _check_version_filters().
- Added an extra check for test-am-utils.sh to catch the redownload bug.
- Fixed _check_version_filters() so that test-am-versions will pass with awk/mawk/gawk/goawk/original-awk:
  - Eg version "2603a" doesn't contain a dot, so pattern 3 won't match. Added a dedicated pattern for alphanumeric version identifiers like "2603a" that don't contain dots.
  - Allow any number of additional dot-separated components with digits and letters for handling "1.0rc2" and "0.15.2.4".
  - If the version string on the filename is longer than the one in the URL, it will select that instead.

List of apps with wrong versions (checked in test-am-versions):
- **nvim**:
  - String: https://github.com/neovim/neovim/releases/download/v0.11.7/nvim-linux-x86_64.appimage
  - Expected Version: "**0.11.7**" but instead gets "**e**" (with mawk, gawk works correctly)
- **dolphin-emu**
  - String: "https://github.com/pkgforge-dev/Dolphin-emu-AppImage/releases/download/2603a%402026-05-01_1777638978/Dolphin-2603a-anylinux-x86_64.AppImage"
  - Expected Version: "**2603a**" but instead gets "**1777638978**"
- **openrgb**:
  - String: "https://codeberg.org/OpenRGB/OpenRGB/releases/download/release_candidate_1.0rc2/OpenRGB_1.0rc2_x86_64_0fca93e.AppImage"
  - Expected Version: "**1.0rc2**" but instead gets "**1.0**"

@ivan-hc This new test is safe to run on the host system, it is just a string checker script.


ivan-hc:
LGTM

Please change
```
AMVERSION="10.2-3"
```
to
```
AMVERSION="10.2-4"
```

PR #2386, opened 2026-05-23
Antigravity: fix desktop files installation

cheack:
The url-handler .desktop file was never installed, which meant that after OAuth login, the browser couldn't redirect back to the app.

Both .desktop files are now downloaded directly from AUR and installed after the AppImage build step.

Simplified the install script by removing the `--appimage-extract` loop that resolved symlinks - it's no longer needed since desktop files and icon are fetched directly.

### DIFF
--- a/programs/x86_64/antigravity
+++ b/programs/x86_64/antigravity
@@ -16,8 +16,6 @@ version=$(curl -Ls https://raw.githubusercontent.com/BOTOOM/google-antigravity-b
 download="$version"
 wget "$download" || exit 1
 # Add custom AppRun here
-curl -s -Lo ../"$APP".desktop "https://raw.githubusercontent.com/archlinux/aur/refs/heads/antigravity-ide/antigravity-ide.desktop" || exit 1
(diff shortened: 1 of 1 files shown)


ivan-hc:
This made the AppImage different from the others.

For debugging purposes, it is better to leave them as much equal as possible.

ivan-hc:
I'll merge this anyway, don't worry.

But a prebuilt AppImage of this would be better. The issue is that we need a reliable source from where we can download the upstream package. AUR isn't reliable enough.

ivan-hc:
@cheack this is exactly what I mean

https://github.com/ivan-hc/AM/commit/cc86048351471e62b0feaca605968bdf361343d2

I have done a video that shows you what happens if you run
```
am reinstall --launcher
```
The only AppImage that complain about .desktop file update is `antigravity`

https://github.com/user-attachments/assets/f46bf0b3-93b2-4252-a4ca-a5b9ff6b198c

ivan-hc:
@cheack I'm working on a way to assure that the created AppImage is ready to run --appimage-extract

I will apply this to all similar scripts.

You will see in the next pull request. I will tag you there.

PR #2402, opened 2026-05-28
Added new AM Fedora container dockerfile and fine tuned _check_version_filters() awk code

vishnu350:
1. Added new AM Fedora container (am-fedora.dockerfile). Regression is fully passing with no issues.
2. Resolve rare cases of incorrect version numbers from _check_version_filters() awk code:
- Fix: `sed` should be case insensitive when replacing architecture numbers (eg. match both AMD64 and amd64). The example below will wrongly gives version **64** instead of **-** :
"https://github.com/dev90/qarma1212/releases/download/no-version/qarma-anylinux-AMD64.AppImage"
- Fix: Type-3 can sometimes wrongly be matched from a long hash number, eg. string below will wrongly give alphanumeric **33d39** instead of hash **aed33d39**:
"https://github.com/dev90/qarma1212/releases/download/version-aed33d39/qarma-anylinux-x86_64.AppImage"
- Fix: Allow for simple Type-3 alphanumeric build numbers, eg. **2b** (minimum 2 chars), which is at the lowest priority. Example below:
"https://github.com/dev90/qarma1212/releases/download/ver2b/qarma-anylinux-x86_64.AppImage"
- Added final if-else to at least choose the better type, instead of just picking version[0] which may be the worst Type-5 (no match).
  ```
  else if (type[0] < type[1]) print version[0]
  else print version[1]
  ```
- Added more failing corner cases (including all of the above) into test-am-version.sh, passing for awk/mawk/goawk/nawk/original-awk in all containers including Debian Jessie.
- Improved inline comments.


vishnu350:
Btw @ivan-hc, the Fedora container seems to be using curl instead of wget. I assume this is because of wget2. Is this expected?

ivan-hc:
> Btw @ivan-hc, the Fedora container seems to be using curl instead of wget. I assume this is because of wget2. Is this expected?

yes, correct

PR #2405, opened 2026-05-30
Update AppImage on-the-fly scripts

ivan-hc:


ivan-hc:
@cheack please see if `antigravity` works as it should now

I also edited the script in order to create the AppRun using `printf`. I have done a test with the previous one and didn't started.

cheack:
@ivan-hc 

The `antigravity-ide-url-handler.desktop` file isn't being created (so you can't redirect back to the app after logging in through the browser), though it looks like they fixed the login check (you can just click Next and it works now, whereas before it would get stuck forever on "Logging in..."):
image

 So, it's probably not an issue.

ivan-hc:
@cheack try now

cheack:
@ivan-hc Okay, it seems to work now, thanks.

PR #2413, opened 2026-06-07
Removed  st+ install script sed on desktop entry.

vishnu350:
Removed sed command on st+.desktop file and removed symlink search loop.

### DIFF
--- a/programs/x86_64/st+
+++ b/programs/x86_64/st+
@@ -63,19 +63,5 @@ fi
 # LAUNCHER & ICON
 ./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
 ./"$APP" --appimage-extract .DirIcon 1>/dev/null && mv ./squashfs-root/.DirIcon ./DirIcon
-COUNT=0
-while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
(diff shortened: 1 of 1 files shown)


ivan-hc:
> Removed sed command on st+.desktop file and removed symlink search loop.

I do not agree. Scripts must be almost equal, for debugging purposes.

For AppImages in particular, the only thing you can do is to remove the whole part of launchers and icons until the end, if you want to made it a CLI app. Completely, not in part. Your app has no issues to justify that the script must be different.

EDIT: also, `sed` patches the .desktop file for AM by default, but such references are changed on the fly for AppMan users.

vishnu350:
OK, but can I at least remove the sed line? It is changing the original Exec line and breaking it.

ivan-hc:
> OK, but can I at least remove the sed line?

I said

> EDIT: also, `sed` patches the .desktop file for AM by default, but such references are changed on the fly for AppMan users.

Aniway...

> It is changing the original Exec line and breaking it.

What do you mean? What is broken?

vishnu350:
The Exec command looks like this:
```Exec=sh -c "command -v tmux >/dev/null 2>&1 && st+ -t 'st+tmux' -e tmux new -A -s st_session || st+"```

The sed command replaces sh with st+.

vishnu350:
I think I have a better solution, I will just fix the .desktop file so that this can be kept simple. I will close this PR.

PR #2414, opened 2026-06-08
KiCad source switch

schnoog:
Switching source from sourceforce to github where most recent appimages are published (packed as tar).

Unfortunately the sourceforge repo wasn't updated for some time  (V9.0.3 which is the most recent one available was released in July 2025, 9.0.4 released a month afterwards fixed some critical bugs).

The new kicad repo is managed by kidac self, so I hope there's some sustainability. 

remark:
I had trouble with the "original" un-tar version, so I added a piped tar x

### DIFF
--- a/programs/x86_64/kicad
+++ b/programs/x86_64/kicad
@@ -3,7 +3,7 @@
 # AM INSTALL SCRIPT VERSION 3.5
 set -u
 APP=kicad
-SITE="https://sourceforge.net/p/kicad-appimage"
+SITE="KiCad/kicad-source-mirror"
 
 # CREATE DIRECTORIES AND ADD REMOVER
 [ -n "$APP" ] && mkdir -p "/opt/$APP/tmp" "/opt/$APP/icons" && cd "/opt/$APP/tmp" || exit 1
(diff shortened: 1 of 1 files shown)


ivan-hc:
> remark: I had trouble with the "original" un-tar version, so I added a piped tar x

What trouble?

Anyway, the original `wget` is done like this to be replaced with `curl` on system without `wget` or where `wget` is a fake one (see Fedora).

ivan-hc:
I can reproduce the issue: we use `[ -e ./*tar.* ] && tar fx ./*tar.* && rm -f ./*tar.*` but there is no dot.

Changing it to `[ -e ./*tar* ] && tar fx ./*tar* && rm -f ./*tar*` solves the issue.

schnoog:
Yes, that was the issue. So I did what I do for 20 years and piped it into tar. I'm a lazy old git ;)

Thanks for merging the PR.

vishnu350:
Hi @schnoog, sorry but do you mind sharing how does the appimage version of KiCAD hold up to daily use? It has many sub-utilities and is very hard to properly package as an appimage.

schnoog:
Good question. I'm in the process of switching from EasyEDA to KiCad and dont have any meaningfull experience yet.
But I prefer to start with the most recent (stable) release.

Since the repo contains the official appimage released by the kicad team (same md5) I trust that they take care of it. 

Sorry to be not more helpful.

Cheers

PR #2426, opened 2026-06-14
Add bookdb (x86_64)

cadwal:
Adds **BookDB**, a desktop book-catalog application, for `x86_64`.

- Upstream: https://github.com/cadwal/BookDB (I'm the project author)
- Install script: `programs/x86_64/bookdb` — Option Zero (AppImage from GitHub releases), generated from `AM-SAMPLE-AppImage` (`AM INSTALL SCRIPT VERSION 3.5`), unmodified apart from `APP`/`SITE` and the version-detection line.
- Catalog entry added to `programs/x86_64-apps` (left `x86_64-appimages` untouched, per CONTRIBUTING.md).

Tested locally with `am -i ./bookdb` under WSL: installs, the `bookdb` command launches, icon/.desktop are extracted, and `am -R bookdb` removes cleanly.

An `aarch64` script exists too; I'll submit it separately once I've tested it on real arm64 hardware.

PR #2430, opened 2026-06-15
Add bookdb (aarch64)

cadwal:
Companion to #2426 (x86_64). Tested on a Raspberry Pi 500.


cadwal:
A new micro-HDMI to HDMI converter arrived today, I don't quite know what I did with the previous one...

PR #2435, opened 2026-06-17
megaglest: pin x86_64 to stable release, remove aarch64

andy5995:


ivan-hc:
Hi, when you remove an app it is enough to remove the script, the action will remove it from lists too.

PR #2443, opened 2026-06-26
Check AppImage while install and alternate --appimage-extract with 7z

### LINKED ISSUE #2442: "--appimage-extract" step on some install scripts && 'nolibfuse' option fail on void linux (x86_64-musl)
first and foremost i apologize in advance if this is a redundant and obvious fix, (or if musl is out of scope to begin with, i read some closed issues and concluded it is not, but i could be wrong), this is like my 4th "serious" github issue ever so please bear with me, AND sorry if this is a big read, i tried to cut as much irrelevant details as i considered, i respect what you guys do very much and i'd hate to be a nuisance

so my thing is that: when installing some appimage apps through ```am``` (like ```localsend``` and ```osu``` for example), their ```'am install X'``` script fails at the extraction step (for their .DirIcon and .desktop files accordingly), and not only that but the ```am nolibfuse``` option aborts itself as well for some reason

these 2 apps specifically have pretty much the exact same install script sequence (i checked with am -d), both of them fail at line 60, which is the ```--appimage-extract && mv``` step for both of them
ss:

Image

 (excuse the weird warped text)

---
this on itself is not that big of a deal for me because i can just modify and run the scripts to make it so they extract with ```7z x "$APP" -o"./squashfs-root"``` rather than ```--appimage-extract```, (because i hadn't mentioned this yet but: according to ```squashfs-tools``` extraction seems to fail becuase "they are not valid squashfs appimages" apparently)

````
sanker@SankerLv62 ~/osu $ doas cp /opt/osu/osu ./ && unsquashfs -d ./squashfs-root ./osu
FATAL ERROR: Can't find a valid SQUASHFS superblock on ./osu
````
same goes for localsend

this of course is weird to me because these failed extraction issues never occurred in something like artix linux for example (with these 2 very apps as well), so i concluded that either i'm missing essential utilities which i am not aware of, or this is some glibc boogery (i of course suppose it's not the latter since these are all bash/shell scripts and not actual compiled binaries, but still)

----
also worth noting:
**with gcompat installed**
````
sanker@SankerLv62 ~/osu $ ./osu --appimage-extract
This doesn't look like a squashfs image.
Failed to open squashfs image
````
**without**
````
sanker@SankerLv62 ~/osu $ ./osu --appimage-extract
bash: ./osu: cannot execute: required file not found
````

----

i know i can just create a little glibc chroot environment and run these conflicting apps from there (matter of fact i did do that), but i wanted to open an issue about this regardless mostly becuase i found it weird that SOME apps did work and some did not, steam from ```am``` for example works like a charm no issues outside the chroot (im actually very impressed about it), so it was definitely odd to me

so yeah
apologies again, and thank you for reading

### PR DESCRIPTION
**Pull request by ivan-hc**
fix https://github.com/ivan-hc/AM/issues/2442

## To do
- [x] add change to option `nolibfuse` (done in https://github.com/ivan-hc/AM/commit/15ce87c89fffcd48da593e220c1cc662cbafe9ab)
- [x] patch installation scripts during the installation

### CODE REVIEW

**Comment by ivan-hc** [comment]
https://github.com/ivan-hc/AM/pull/2443/commits/f9fe456ccc66f4c900a6155e2c1f931788fef5eb patches the script on the fly by adding a function to the script

However, this become a default patch. What I would like to do is to apply it only on Musl systems, since old AppImages depending on libfuse2 also depend on GLIBC.

On the other hand, this may be a shield against **fake AppImages** (i.e. executable files replaced upstream and shipped as AppImages). It would be enough to extend the patch to check if the AppImage is an AppImage.

For example:
```
_appimage_extract() { if ! ./"$APP" --appimage-extract "$@"; then command -v 7z >/dev/null 2>&1 && 7z x ./"$APP" -o./squashfs-root "$@"; fi }
```
...would include a
```
head -c10 ./"$APP" 2>/dev/null | grep -qa '^.ELF....AI$'
```
or a
```
head -c1000000 ./"$APP" 2>/dev/null | grep -qa 'AppImages require FUSE to run'
```
or other commands in POSIX shell as alternative.

The above are what we already use to determine if a file is an AppImage or an old AppImage, respectively, and without `file` dependency.

## Why don't adding this directly in the scripts?

Because this change is dynamic, thye day @probonopd and CO. will decide to release Type3, Type4 or TypeN runtime for AppImages, the way to detect if they are AppImages may vary.

I plan to made this project as much time-proof as possible, in view of my possible departure.

I already have not much time lately, so I hope you could help me test and improve this feature as much as possible.

Thanks in advance.

**Comment by ivan-hc** [comment]
The last commit improves the new function added dynamically like this
```
_appimage_extract() {
	if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then
		if ! ./"$APP" --appimage-extract "$@"; then
			command -v 7z >/dev/null 2>&1 && 7z x ./"$APP" -o./squashfs-root "$@"
		fi
	fi
}
```
For reference, this common scheme of desktop AppImages is patched and changes from this
```
# LAUNCHER & ICON
./"$APP" --appimage-extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
./"$APP" --appimage-extract .DirIcon 1>/dev/null && mv ./squashfs-root/.DirIcon ./DirIcon
COUNT=0
while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
	if [ -L ./"$APP".desktop ]; then
		LINKPATH="$(readlink ./"$APP".desktop | sed 's|^\./||' 2>/dev/null)"
		./"$APP" --appimage-extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./"$APP".desktop
	fi
	if [ -L ./DirIcon ]; then
		LINKPATH="$(readlink ./DirIcon | sed 's|^\./||' 2>/dev/null)"
		./"$APP" --appimage-extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./DirIcon
	fi
	[ ! -L ./"$APP".desktop ] && [ ! -L ./DirIcon ] && break
	COUNT=$((COUNT + 1))
done
```
to this
```
# LAUNCHER & ICON
_appimage_extract() { if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then if ! ./"$APP" --appimage-extract "$@"; then command -v 7z >/dev/null 2>&1 && 7z x ./"$APP" -o./squashfs-root "$@"; fi; fi }
_appimage_extract *.desktop 1>/dev/null && mv ./squashfs-root/*.desktop ./"$APP".desktop
_appimage_extract .DirIcon 1>/dev/null && mv ./squashfs-root/.DirIcon ./DirIcon  2>/dev/null
COUNT=0
while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
	if [ -L ./"$APP".desktop ]; then
		LINKPATH="$(readlink ./"$APP".desktop | sed 's|^\./||' 2>/dev/null)"
		_appimage_extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./"$APP".desktop
	fi
	if [ -L ./DirIcon ]; then
		LINKPATH="$(readlink ./DirIcon | sed 's|^\./||' 2>/dev/null)"
		_appimage_extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./DirIcon  2>/dev/null
	fi
	[ ! -L ./"$APP".desktop ] && [ ! -L ./DirIcon ] && break
	COUNT=$((COUNT + 1))
done
```

Having keywords `appimage-extract` and `# LAUNCHER & ICON` in the script is core, or the patch will not be added.

I have done the test with my Firefox AppImage (from where I'm writing) and `osu` that is one of the libfuse2 apps of the [issue](https://github.com/ivan-hc/AM/issues/2442) from where this change is started.

I did the test on Debian, now I'm going to do the test on freeBSD too.

**Comment by ivan-hc** [comment]
> I did the test on Debian, now I'm going to do the test on freeBSD too.

Done, if on freeBSD you install `gsed` (GNU `sed`, BSD ports, `sudo pkg install gsed`) instead of using BSD `sed`, it works as expected.

**Comment by ivan-hc** [comment]
> Having keywords `appimage-extract` and `# LAUNCHER & ICON` in the script is core, or the patch will not be added.

According to this, I have standardized the following apps:
- calligra
- kdegames
- kdeutils
- libreoffice
- wps-office

All the AppImages with multiple launchers and icons.

**Comment by ivan-hc** [comment]
@Samueru-sama @fiftydinar @vishnu350 @cheack 

I'm looking for review (and suggestions).

File: modules/install.am
Comment (vishnu350) [inline modules/install.am:1]: AppImage extraction can sometimes fail with the built-in --appimage-extract flag, by adding a 7z fallback, the script makes the extraction process more reliable, rather than stopping completely. I think this is an excellent change.

Comment (vishnu350) [reply modules/install.am:1]: @ivan-hc I think the . in ./ should be escaped in the regex:
```sed -i -- 's#\./"$APP" --appimage-extract #_appimage_extract #g' ./"$arg"```

Comment (ivan-hc) [reply modules/install.am:1]: > @ivan-hc I think the . in ./ should be escaped in the regex: `sed -i -- 's#\./"$APP" --appimage-extract #_appimage_extract #g' ./"$arg"`

what happens if I leave it as is?

**Comment by ivan-hc** [comment]
If there were a way to made
```
_appimage_extract() {
	if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then
		if ! ./"$APP" --appimage-extract "$@" 2>/dev/null; then
			command -v 7z >/dev/null 2>&1 && 7z x ./"$APP" -o./squashfs-root "$@"
		fi
	fi
}
```
exportable from install.am directly trough the various shells (I thing `export -f _appimage_extract` only would work in BASH), we had more ways to improve it in the module, by adding more checks instead of pushing it directly in the script.

Better if i do more tests.

File: modules/install.am
Comment (vishnu350) [reply modules/install.am:1]: It will still work. In regex `.` means any character, like *

`\.` is the real `.`

Comment (ivan-hc) [reply modules/install.am:1]: Done https://github.com/ivan-hc/AM/pull/2443/commits/233e93719427f0876d526175845a44d4f404747f

**Comment by ivan-hc** [comment]
https://github.com/ivan-hc/AM/commit/a173773c18d6f8de67e00ebc581b6a4cc24c7683

With this change we can read the function even better and add features, like in this case, with 7zz support.

The script is patched via `awk` and replaced in place.

	# Check safely if executable file is an AppImage and then alternate --appimage-extract with 7z if the first fails
	if grep -q "appimage-extract" ./"$arg" && grep -q "# LAUNCHER & ICON" ./"$arg"; then
		sed -i -- 's#\./"$APP" --appimage-extract #_appimage_extract #g' ./"$arg"
		# Add function to extract AppImages
		block=$(
			cat <<-'HEREDOC'
			_appimage_extract() {
			   if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then
			      if ! ./"$APP" --appimage-extract "$@" 2>/dev/null; then
			         if command -v 7z >/dev/null 2>&1; then
			            7z x ./"$APP" -o./squashfs-root "$@"
			         elif command -v 7zz >/dev/null 2>&1; then
			            7zz x ./"$APP" -o./squashfs-root "$@"
			         fi
			      fi
			   fi
			}
			HEREDOC
		)
		awk -v block="$block" '
			{ print }
			/# LAUNCHER & ICON/ { print block }
		' "$arg" > ./"$arg.tmp" && mv ./"$arg.tmp" ./"$arg" && chmod a+x ./"$arg" || return 1
	fi

This is the function that will be added in the script

		_appimage_extract() {
		   if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then
		      if ! ./"$APP" --appimage-extract "$@" 2>/dev/null; then
		         if command -v 7z >/dev/null 2>&1; then
		            7z x ./"$APP" -o./squashfs-root "$@"
		         elif command -v 7zz >/dev/null 2>&1; then
		            7zz x ./"$APP" -o./squashfs-root "$@"
		         fi
		      fi
		   fi
		}

If you want to add support for `unsquashfs` or something else let me know... or PR me.

**Comment by ivan-hc** [comment]
```
_patch_installation_script_with_function_to_extract_appimages() {
	sed -i '/# LAUNCHER & ICON/a\
	\
	_appimage_extract() {\
		if head -c10 ./"$APP" 2>/dev/null | grep -qa "^.ELF....AI$"; then\
			if grep -Eaoq -m 1 "appimage-extract" ./"$APP"; then\
				if ! ./"$APP" --appimage-extract "$@" 2>/dev/null; then\
					if command -v 7z >/dev/null 2>&1; then\
						7z x ./"$APP" -o./squashfs-root "$@"\
					elif command -v 7zz >/dev/null 2>&1; then\
						7zz x ./"$APP" -o./squashfs-root "$@"\
					else\
						exit 0\
					fi\
				fi\
			fi\
		else\
			exit 0\
		fi\
	}\
	' ./"$arg"
}
```
this will add the function `_appimage_extract` to the script with various checks:

1. check if the file is an appimage
2. check if it supports --appimage-extract
3. use 7z/7zz if --appimage-extract fails

`exit 0` if one of the above fails, this way no other operation is done in the script (prevents to `sed`/`mv` files that does not exist)

PR #2448, opened 2026-06-28
Remove HarbourMasters Games due to Incompatibilities

### LINKED ISSUE #2427: Shipwright, 2ship, spaghettikart Pollute Home and Interfere With Each Other
These three Harbour Masters releases all have the same issue, they work directly in the path they're run from. As such, when running them as installed, they'll dump all this into the user's home folder. For SpaghettiKart specifically, that's this file tree:

```
~
├── controllerPak_header.sav
├── default.sav
├── imgui.ini
├── logs
│   └── Spaghetti Kart.log
├── mk64.o2r
├── mods
├── spaghettify.cfg.json
└── torch.hash.yml
```

For these apps specifically, this can be worked around by changing the working directory of the app. In this case, I added `Path=/home/user/.config/spaghettikart` to the desktop file to force it to work from a more sensible path as a test to ensure this is in-fact how it works and confirmed it. 

Alternatively, `shipwright` has `soh-ehanced` and `spaghettikart` has `spaghettikart-enhanced` as alternative options that utilize better paths by default, but they require external intervention to download and utilize the official appimage to generate game files on ever major version change, can fall behind on updates occasionally, and there's no such alternative for `2ship`.

It's also worth noting that `starship-port` has this issue as well, but additionally has extra necessary assets in its root directory and no enhanced option.

SkyNinja:
Removes the following items from the repo:

- `2ship`
- `shipwright`
- `spaghettikart`
- `starship-port`

Closes #2427


ivan-hc:
Welcome on board

PR #2455, opened 2026-07-01
Add more info about using `am translate auto`

fiftydinar:
Currently, the option was not exposed to the user. Now we expose it through `am translate` and in `am help`.


ivan-hc:
@fiftydinar just checked with Poedit, and all (except ja.po and zh_CN.po) **need to be verified**.

Something went wrong when adding translations. Please fix it.

fiftydinar:
idk, what happened, I'll have to re-do this

PR #2457, opened 2026-07-01
fix: expand env vars in install path

QaidVoid:
Expands `~` and env vars (`$VAR`/`${VAR}` )in the AppMan install-path prompt.

## Before

image

## After

image


ivan-hc:
@QaidVoid thanks for this your first contribution! Welcome on board!

PR #2459, opened 2026-07-02
Allow AppImage templates to handle multiple launchers

ivan-hc:
This change will update templates to v3.6

Some AppImages have multiple launchers in the root of the AppDir or squashfs-root directory

This change will mark launcher containing the `NoDisplay=true` entry as "helper", while the regular one will be renamed as "$APP"-AM.desktop normally

```
# LAUNCHER & ICON
./"$APP" --appimage-extract *.desktop 1>/dev/null
./"$APP" --appimage-extract .DirIcon 1>/dev/null && mv ./squashfs-root/.DirIcon ./DirIcon
COUNT=0
while [ "$COUNT" -lt 10 ]; do # Tries to get the actual icon/desktop if it is a symlink to another symlink
	for f in squashfs-root/*.desktop; do
		if [ -L "$f" ]; then
			LINKPATH="$(readlink "$f" | sed 's|^\./||' 2>/dev/null)"
			./"$APP" --appimage-extract "$LINKPATH" 1>/dev/null
			if grep -qi '^NoDisplay=true$' ./squashfs-root/"$LINKPATH" 2>/dev/null; then
				mv ./squashfs-root/"$LINKPATH" ./"$APP"-handler-AM.desktop && printf '\n%s' "rm -f /usr/local/share/applications/$APP-handler-AM.desktop" >> ./remove
			else
				mv ./squashfs-root/"$LINKPATH" ./"$APP"-AM.desktop
			fi
		else
			if grep -qi '^NoDisplay=true$' "$f"; then
				mv "$f" ./"$APP"-handler-AM.desktop && printf '\n%s' "rm -f /usr/local/share/applications/$APP-handler-AM.desktop" >> ./remove
			else
				mv "$f" ./"$APP"-AM.desktop
			fi
		fi
	done
	if [ -L ./DirIcon ]; then
		LINKPATH="$(readlink ./DirIcon | sed 's|^\./||' 2>/dev/null)"
		./"$APP" --appimage-extract "$LINKPATH" 1>/dev/null && mv ./squashfs-root/"$LINKPATH" ./DirIcon
	fi
	[ ! -L ./"$APP".desktop ] && [ ! -L ./DirIcon ] && break
	COUNT=$((COUNT + 1))
done
sed -i "s#Exec=[^ ]*#Exec=$APP#g; s#Icon=.*#Icon=/opt/$APP/icons/$APP#g; s# --no-sandbox##g" ./*-AM.desktop
[...]

PR #2468, opened 2026-07-08
Add opencode-desktop

VictorVow:
## Summary
- Add `opencode-desktop`: install script for the opencode desktop app (Electron-based AppImage from anomalyco/opencode releases), listed alongside the existing `opencode` and `opencode-enhanced` packages.
- Fix `opencode`: upstream anomalyco/opencode no longer ships a plain AppImage, so the package's generic `mage$` filter was silently matching the new `opencode-desktop-*.AppImage` asset instead of the CLI. Switched it to fetch the `linux-x64` CLI tarball (`the-way`-style tar.gz install pattern), and removed it from the appimages/stats lists since it no longer creates a desktop launcher.

## Test plan
- [x] Verified the `opencode-desktop` grep pattern resolves to `opencode-desktop-linux-x86_64.AppImage`
- [x] Verified the fixed `opencode` grep pattern resolves to `opencode-linux-x64.tar.gz`
- [x] Downloaded and extracted the `opencode` tarball locally and confirmed `opencode --version` runs correctly

### DIFF
--- a/programs/stats-appimages
+++ b/programs/stats-appimages
@@ -1593,7 +1593,6 @@
 ◆ openbor : Ultimate 2D side scrolling engine for beat em ups, shooters... #itsdesktopapp
 ◆ openclaw : Unofficial, reimplementation of Captain Claw (1997) platformer. #itsdesktopapp
 ◆ opencloud : Desktop application to synchronize files from OpenCloud with your computer. #itsdesktopapp
-◆ opencode : The open source coding agent. #itsdesktopapp
 ◆ opencode-desktop : Desktop application for opencode, the open source coding agent. #itsdesktopapp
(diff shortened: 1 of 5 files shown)


ivan-hc:
Welcome on board

About lists, it is enough to edit the x86_64-apps file, the workflow will do the rest

ivan-hc:
You have not tested the opencode-desktop app, you didn't noticed that the .desktop file was not created

https://github.com/ivan-hc/amcheck/actions/runs/28937551716/job/85851648302

This is happened because you have not created the script using `am -t`, that now have the fix for all AppImages with multiple .desktop file in the root directory https://github.com/ivan-hc/AM/pull/2459

VictorVow:
Ah sorry was creating the script by using a recent PR as a template, I'll use `am -t` in the future. Thanks!

Issue #2514, opened 2026-07-24
Vocalinux now has an appimage available. Please consider adding


aaylnx:
1. vocalinux
2. https://github.com/jatinkrmalik/vocalinux

Thank you

ivan-hc:
Thanks @vishnu350 to care about this in my absence

aaylnx:
Nicely done. Thank you!

PR #2515, opened 2026-07-24
Added vocalinux and regress fixes.

vishnu350:
Two updates:
- Added vocalinux #2514.
- Fixed cp -i alias causing issues in am-fedora during regress testing.


ivan-hc:
Thanks, I took the liberty of modifying the file so as not to explicitly use the "latest" tag, which may not include an AppImage in the future.

At the same time, I added keywords to avoid so you can use the stable version without tag restrictions.

PR #2516, opened 2026-07-25
Add official WeChat AppImage and change some prompt in template.am

DtDEncryption:
Both "wewechat" and "electron-wechat" are outdated and Tecent has been publishing native WeChat AppImage for years. So I propose to add the official WeChat AppImage.

---

Besides, I was kind of confused about the prompts on creating the script:
```
 USE A ONE-LINE COMMAND TO CHECK THE URL TO THE PROGRAM OR THE VERSION

 if the URL is fixed, simply add the "echo" command at the beginning
```
I was wondering if it means "add the 'echo' command at the beginning of the URL" or "only an 'echo' command needs to be added at the beginning of what user should type and that's all". It was only after I checked the script that I understood what should I type.

My change to the prompt here is merely served as an illustration.

### DIFF
--- a/modules/template.am
+++ b/modules/template.am
@@ -108,7 +108,7 @@ _edit_script_ending() {
 
 _template_custom_download_url_as_version_variable() {
 	# IF YOU CAN, USE A ONE-LINE COMMAND TO DOWNLOAD THE PROGRAM
-	read -r -ep "$(printf $" USE A ONE-LINE COMMAND TO CHECK THE URL TO THE PROGRAM OR THE VERSION\n\n if the URL is fixed, simply add the \"echo\" command at the beginning\n\n :") " DOWNLOADURL
+	read -r -ep "$(printf $" USE A ONE-LINE COMMAND TO CHECK THE URL TO THE PROGRAM OR THE VERSION\n\n if the URL is fixed, simply input the \"echo\" command and the static URL\n\n for example: echo https://www.EXAPLE.com\n\n :") " DOWNLOADURL
 	case "$DOWNLOADURL" in
 	*)
 		_edit_script_head
--- a/programs/x86_64/wechat
+++ b/programs/x86_64/wechat
@@ -1,0 +1,85 @@
+#!/bin/sh
+
+# AM INSTALL SCRIPT VERSION 3.6
+set -u
+APP=wechat
+SITE="https://dldir1v6.qq.com/weixin/Universal/Linux/WeChatLinux_x86_64.AppImage"
+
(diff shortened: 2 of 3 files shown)


modules/template.am
```diff
@@ -108,7 +108,7 @@ _edit_script_ending() {
 
 _template_custom_download_url_as_version_variable() {
 	# IF YOU CAN, USE A ONE-LINE COMMAND TO DOWNLOAD THE PROGRAM
-	read -r -ep "$(printf $" USE A ONE-LINE COMMAND TO CHECK THE URL TO THE PROGRAM OR THE VERSION\n\n if the URL is fixed, simply add the \"echo\" command at the beginning\n\n :") " DOWNLOADURL
+	read -r -ep "$(printf $" USE A ONE-LINE COMMAND TO CHECK THE URL TO THE PROGRAM OR THE VERSION\n\n if the URL is fixed, simply input the \"echo\" command and the static URL\n\n for example: echo https://www.EXAPLE.com\n\n :") " DOWNLOADURL
```
ivan-hc, line 111: Are you willing to translate this line into all other supported languages?

https://github.com/ivan-hc/AM/tree/main/translations

programs/x86_64/wechat
```diff
@@ -0,0 +1,85 @@
...
+ln -s "/opt/$APP/$APP" "/usr/local/bin/$APP"
+
+# SCRIPT TO UPDATE THE PROGRAM
+cat >> ./AM-updater << 'EOF'
+#!/bin/sh
+set -u
+APP=wechat
+SITE="https://dldir1v6.qq.com/weixin/Universal/Linux/WeChatLinux_x86_64.AppImage"
+version0=$(cat "/opt/$APP/version")
+version=$(echo https://dldir1v6.qq.com/weixin/Universal/Linux/WeChatLinux_x86_64.AppImage)
```
ivan-hc, line 36: This means that wechat have no updates? Do they rely on only one release without a version?

DtDEncryption:
@ivan-hc 
I'm afraid I cannot take care of "all other supported languages". But if it is en or zh_Hans, I would be glad to help.

As for the version variable, sorry I am not good with code and could only figure out if some input are illegal or not.  AFAIK, wechat needs update, but doesn't come with a repo or an archive. Normally it is downloaded from https://linux.weixin.qq.com/en, which can be accessed from https://www.wechat.com. (Suprisingly, Linux option only appears when choosing simplified Chinese.) I found the download URL seemed static, so I pasted it in the process of template creation.

PR #2518, opened 2026-07-25
Add official WeChat AppImage

DtDEncryption:
Both "wewechat" and "electron-wechat" are outdated and Tecent has been publishing native WeChat AppImage for years. So I propose to add the official WeChat AppImage.

---

PS: After some research, I found the "version" variable is used extensively both for version checking and for package download. But in the case of static download URL without version annotation, such a design ruins version checking, signature verification, etc. Therefore I hope a more viable design would replace the current one.


ivan-hc:
You added the x86_64 directory and the x86_64-apps list to the root of the repository.

They must be in /programs.

PS: the version is 4.1.1, while the download URL is the one you used for $version. This script require manual intervention.

Allow me to help you adding this app.

ivan-hc:
> Both "wewechat" and "electron-wechat" are outdated and Tecent has been publishing native WeChat AppImage for years. 

If `wewechat` and `electron-wechat` are archived, obsolete and outdated, @DtDEncryption would you like to remove them from AM?

It is enough that you remove the two scripts, github actions will update the lists for you.

Thanks.

DtDEncryption:
@ivan-hc  No problem.

PR #2555, opened 2026-08-06
Added Citron Nightly 

NeoPlayer13:
Switch Emulator


ivan-hc:
In the five years that AM has existed, I've added and removed several applications. One of the most frequently added and removed is this one, from other authors (or perhaps the same authors, with a different repository or account), as the repository was repeatedly shut down. Citron is a name that hasn't found peace in AM.

I'm adding it, but I don't expect to see it for much longer, to be honest.

If I were the author, I'd completely change the domain where the artifacts are hosted.

PR #2569, opened 2026-08-08
Adding app - Ente Ensu Stripped

melvinquick:
Adding app - Ente Ensu Stripped. It's my own project that pulls GPU/Display libraries out of the AppImage Ente offers for Ensu that cause it to not run on Wayland unless specifying the host system libraries to run at runtime.


ivan-hc:
Welcome on board!

PR #2584, opened 2026-08-11
Option "downgrade", ask if you want to "include" snapshots

ivan-hc:
...not "exclude", to made the behavior consistent with all "confirmation" prompts


vishnu350:
It was always a confusing prompt, because you have to enter "No" to get more items in the list.

Now "Yes" will give more. Much better.

PR #2614, opened 2026-08-17
Update the version variable for fdroidcl

fazlerabbi37:
With the previous string manipulation it was getting the release URL and
then downloading the JSON instad of the asset. Now it looks for the
`browser_download_url` of `linux_amd64` and downloads that.

### DIFF
--- a/programs/x86_64/fdroidcl
+++ b/programs/x86_64/fdroidcl
@@ -12,7 +12,7 @@ printf "#!/bin/sh\nset -e\nrm -f /usr/local/bin/$APP\nrm -R -f /opt/$APP" > ../r
 chmod a+x ../remove || exit 1
 
 # DOWNLOAD AND PREPARE THE APP, $version is also used for updates
-version=$(curl -Ls https://api.github.com/repos/Hoverth/fdroidcl/releases | sed 's/[()",{} ]/\n/g' | grep -oi "https.*" | grep -vi "i386\|i686\|aarch64\|arm64\|armv7l" | head -1)
+version=$(curl -Ls https://api.github.com/repos/Hoverth/fdroidcl/releases/latest | grep -o 'browser_download_url": "[^"]*linux_amd64[^"]*"' | grep -v '\.gz' | sed 's/.*"\(https[^"]*\)".*/\1/' | head -1)
 wget "$version" || exit 1
 [ -e ./*7z ] && 7z x ./*7z && rm -f ./*7z
 [ -e ./*tar.* ] && tar fx ./*tar.* && rm -f ./*tar.*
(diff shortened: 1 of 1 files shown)


ivan-hc:
Hi, I just checked and you are right!

This program was added here https://github.com/ivan-hc/AM/pull/2121 by @shuvashish76 and I was unaware of this error, since amcheck always installs something from there (below the last test for `fdroidcl`)

https://github.com/ivan-hc/amcheck/actions/runs/32035782019/job/95405857573

However, I would not use the `browser_download_url` approach at all, since Json format of api.github.com is not formatted the same way across the clients (in this case `curl`, when `wget2` would behave differently).

We use `sed 's/[()",{} ]/\n/g'` to bypass the needing of dependencies like `jq` in AM. all scripts have this.

Also, according with https://github.com/ivan-hc/AM/pull/2121/changes, `fdroidcl` have also an aarch64 script, not just the x86_64 one.

If you want to fix it, you should use the option `-t`, select the template 2 and follow the instructions. When it asks to add (1) or remove (2) a keyword via `grep`, select 1 (keyword to include) and then write `linux.*64$` (exactly this as I wrote it), since both the linux standalone binaries end with "64" (amd64, arm64... same if they were called x86_64 or aarch64 respectively), and both contain the word "linux". New templates already contain the keyword "download" that is standard for the download URLs on github... but you may do the same like this `download.*linux.*64$` or if you want to manually add the reference, you can do
```
grep -Eoi "http.*download.*linux.*64$"
```
But leave ` | grep -vi "i386\|i686\|aarch64\|arm64\|armv7l"` for x86_64 to avoid different releases, or alternatively use `grep -i "amd64\|x86_64\|x64\|x86-64"` to detect the right release for x86_64 (the name may vary... but I use `grep -vi` because sometime the x86_64 release is the default one and some developers may not specify it in the file name).

For aarch64 leave `grep -i "aarch64\|arm64"`.

This is the approach:

https://github.com/user-attachments/assets/ed6336ed-3639-4716-aefc-890e791561a8

As you can see, the option `-t` creates for you both the scripts, for both the architectures.

Since this is your first contribution to this project, its good to know how the downloads approach works here, in all scripts for apps hosted on github.

Please update both the scripts using the right approach. Thanks in advance.

ivan-hc:
I will merge this in the dev branch to not let you loose the commit, then I will fix this there, just to fix the issue.

@fazlerabbi37 welcome on board btw.

fazlerabbi37:
@ivan-hc sorry for the late reply and thank you merging the MR! Applied the fix in local, tested and then created the MR so it was working ok for me and totally forgot about this MR! 🙏

(came to the repo to check on another app.)

Let me know if you want me to put the fixes above in a separate PR.

ivan-hc:
> Let me know if you want me to put the fixes above in a separate PR.

You are welcome :)

PR #2578, opened 2026-08-10
AM 10.4, add option -y to quickly confirm all y/N prompts

ivan-hc:
Continuation of https://github.com/ivan-hc/AM/pull/2553 by @vishnu350 

Anyone that want to test this, can run
```
am --devmode-enable
am -s
```

## SYNTAX:
```
am -y {OPTION}
am -y {OPTION} {ARGUMENT}
```

PR #2620, opened 2026-08-19
Added support for aborting AM commands that are incompatible with -y

vishnu350:
Three read operations remain from making AM a fully non-interactive user experience with -y in install.am:
1. read -r -p $" Install \"$DID_YOU_MEAN\" with --$DID_YOU_MEAN_FLAG flag? (Y/n) " yn
2. read -r -p $" Install \"$DID_YOU_MEAN\" instead? (Y/n) " yn
3. read -r -p $" Type a number and press ENTER: " choice

Methodology:
1. If the read variable is is named yn_abort or choice_abort, the _read wrapper function will auto select "N" or "0" instead. This will force AM to abort the command.
2. Simple modifications to _read:
```
....
# If variable name is 'yn_abort', default to 'n' to abort the AM command
elif [[ "$var_name" == "yn_abort" ]]; then
	results="n"
# If variable name is 'choice_abort', default to '0' to abort the AM command
elif [[ "$var_name" == "choice_abort" ]]; then
	results="0"
....
```

To test, try to install an app that doesnt exist to trigger fuzzy matching: `am -y -i fccp`
```
 💀 ERROR: "fccp" does NOT exist in the "AM" database, please check
 the list, run the "am -l" command.

 Did you mean "fcp"?
```

AM will not stop and prompt, instead it will just abort and return the user to the CLI.

This behavior matches with how `apt-get -y` works when commands should not receieve a "Y".

All regression tests are fully passing.

### DIFF
--- a/APP-MANAGER
+++ b/APP-MANAGER
@@ -28,6 +28,12 @@ _read() {
 		# If variable name is 'choice', default to '1'
 		elif [[ "$var_name" == "choice" ]]; then
 			results="1"
+		# If variable name is 'yn_abort', default to 'n' to abort the AM command
+		elif [[ "$var_name" == "yn_abort" ]]; then
+			results="n"
+		# If variable name is 'choice_abort', default to '0' to abort the AM command
+		elif [[ "$var_name" == "choice_abort" ]]; then
+			results="0"
 		fi
 	else
 		read -r -ep "$prompt" results
--- a/modules/install.am
+++ b/modules/install.am
@@ -540,9 +540,9 @@ _install_3rd_party_app() {
 			fi
 		elif [ -n "$DID_YOU_MEAN" ]; then
 			if [ -n "$DID_YOU_MEAN_FLAG" ]; then
-				read -r -p $" Install \"$DID_YOU_MEAN\" with --$DID_YOU_MEAN_FLAG flag? (Y/n) " yn
+				_read $" Install \"$DID_YOU_MEAN\" with --$DID_YOU_MEAN_FLAG flag? (Y/n) " yn_abort
 			else
-				read -r -p $" Install \"$DID_YOU_MEAN\" instead? (Y/n) " yn
+				_read $" Install \"$DID_YOU_MEAN\" instead? (Y/n) " yn_abort
 			fi
 			echo ""
 			if ! echo "$yn" | grep -i '^n' >/dev/null 2>&1; then
@@ -686,7 +686,7 @@ _did_you_mean() {
 
 _select_did_you_mean_candidate() {
 	local choice
-	read -r -p $" Type a number and press ENTER: " choice
+	_read $" Type a number and press ENTER: " choice_abort
 	echo ""
 	case "$choice" in
 		''|*[!0-9]*) return 1 ;;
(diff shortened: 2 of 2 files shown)


ivan-hc:
OK @vishnu350 

Sorry... why?

ivan-hc:
> Sorry... why?

Errors are done to tell people "Hey, something is gone wrong. Please fix it first"

If I had `-n` or a like in amcheck, I would simply ignore all failed jobs, and nobody (except people in bug reports) would fix/remove the broken scripts.

ivan-hc:
@cheack what are your thoughts?

cheack:
+1 to merge.

This aligns with the concern I raised in [#2578](https://github.com/ivan-hc/AM/pull/2578#discussion_r3757401853): auto-installing a similarly-named app under -y is worse than just stopping. The "does NOT exist" error already shows up before any guessing happens - this PR doesn't change that. It only changes what happens next: instead of auto-installing a guess or hanging on a prompt (both wrong for -y), it just stops. That's not hiding an error, it's avoiding a bad guess - different from AMCHECK's failed jobs, which are real bugs to fix.

One separate thing worth checking: the install loop currently always exit 0 at the end, regardless of whether individual apps failed inside it. So `am -y -i fccp` returns exit code 0 even though nothing got installed. If a script or CI pipeline is checking `$?` to decide if the install worked, it can't tell success from failure right now - might be worth tracking failures and returning a non-zero exit code at the end.

ivan-hc:
> One separate thing worth checking: the install loop currently always exit 0 at the end, regardless of whether individual apps failed inside it. So `am -y -i fccp` returns exit code 0 even though nothing got installed. If a script or CI pipeline is checking `$?` to decide if the install worked, it can't tell success from failure right now - might be worth tracking failures and returning a non-zero exit code at the end.

what do you mean with this?

vishnu350:
> Sorry... why?

Current behavior for `-y` violates the "non-interactive" premise. By hanging on a prompt, it is "interactive". As @cheack mentioned, auto-selecting "Y" or hanging on the prompt is worse than just aborting.

As reference, here is the manpage for `apt-get -y`. It specifically mentions that it will **abort** in an undesirable situation, instead of waiting for user input.

_-y, --yes, --assume-yes
Automatic yes to prompts; assume "yes" as answer to all prompts and run non-interactively. If an undesirable situation, such as changing a held package, trying to install an unauthenticated package or removing an essential package occurs **then apt-get will abort**._


> Errors are done to tell people "Hey, something is gone wrong. Please fix it first"

The error will still appear. It just wont hang on a prompt for Y/N, and will default to N for these specific cases to force an abort. Give it a try.

vishnu350:
Lastly, the most important reason: The `-y` feature is actually meant for **integration into scripts/tools like AM-GUI** or maybe into something like gnome-software in the future.

These GUIs will have trouble dealing with a hanging prompt, it may hang/crash if it faces a prompt. So we must guarantee -y is **always promptless**, ie. non-interactive.

Hypothetical example: If a user tries to force the GUI to install "fccp", the resulting command will be `am -y -i fcc`, which will force a prompt and break the GUI.

cheack:
@ivan-hc 

> what do you mean with this?

So the point is: exit code is the only signal a script/CI/GUI backend has - it doesn't read terminal text. If `am -y -i somepackage` fails to find/install the app but still returns 0, an automated caller has no way to detect that and will proceed as if it succeeded - e.g. a CI step that installs deps and continues, or a GUI showing "success" when nothing happened. Returning 1 on failure lets the caller actually react (retry, alert, stop the pipeline) instead of silently continuing on a false success.

vishnu350:
> If am -y -i somepackage fails to find/install the app but still returns 0

I checked, it does returns 0 instead of a non-zero number. Seems like a bug.

> what do you mean with this?

I think @cheack is trying to suggest that amcheck (or any script calling AM) should check the return code to determine if the command was successful. Should be bash var $?.

PR #2659, opened 2026-08-29
update firedragon

prodigalshadow:
firedragon's latest release is v13 but the old link still points to v12


ivan-hc:
Hi, please disable workflows in your fork, to not put unneeded commits like the ones for .po files

PR #2660, opened 2026-08-29
Add comic strip browser

prodigalshadow:
A standalone PyQt6 application for browsing a selection of comic strips from GoComics and Comics Kingdom.


ivan-hc:
I just fixed this pull request:
- moved the script to programs/x86_64
- added the line of description to the programs/x86_64-apps file

Previously you added the above two files in the root of the fork.

PR #2667, opened 2026-09-01
Add kate

kazam0180:


ivan-hc:
@kazam0180 same discourse of `knights`: if you have a dedicated appimage (expecially from pkgforge that do a great work on this), just replace the related metapackage.

Metapackage and dedicated appimage can work together for kde apps, since they have their own launcher... but the appimage can be used from terminal, with its own command.

If you can, replace `kate` with your's.

kazam0180:
> If you can, replace `kate` with your's.

that requires removing kate from kdeutils, no?
also, it contains kwrite too.

ivan-hc:
> > If you can, replace `kate` with your's.
> 
> that requires removing kate from kdeutils, no?

Nope, you don't need to touch kdeutils, at all.

Add the app as `kate`, regularly. The workflow will take care of the rest. Trust me.

PR #2677, opened 2026-09-02
Removed android-messages-desktop from obsolete-apps list

melvinquick:
This can be removed since my last [pull request](https://github.com/ivan-hc/AM/pull/2672) transitioned the install script to the newer fork by OrangeDrangon.

### DIFF
--- a/programs/obsolete-apps
+++ b/programs/obsolete-apps
@@ -17,7 +17,6 @@ altitude 2024
 alvr 2022
 amusiz 2021
 anavis 2022
-android-messages-desktop 2019
 animos 2023
 aniship 2023
 ant-downloader 2019


ivan-hc:
Thanks @melvinquick , that file is generated from a workflow that runs every 15 days, so it is outdated and not much reliable sometime.

I still need a solution to made this check in real time, without having issues with API rest on github repositories, hat are very limited... and for a database of so many apps is a big trouble.

melvinquick:
Just a thought @ivan-hc, but after looking at the flow, could you use last push activity instead of release activity (would still show if a repo is alive or not, even if a new release hasn't come out) and just pull the repo data for each repo one time instead of making two individual calls for each one (to /repos/$user_repo instead of making an extra call to /releases also)? You'd cut your API calls in half.

ivan-hc:
This is what I wanted to do from the beginning, but my collaborators preferred that I monitor the latest releases.

To be honest, I'm very torn about this. If there were a better solution, I'd take it immediately.

PR #2700, opened 2026-09-09
Fix v2rayN

CelicaS:


ivan-hc:
I didn't noticed. Why uppercased?

CelicaS:
The project files in uppercase.

ivan-hc:
> The project files in uppercase.

Are there other programs in this database that have uppercase letters in the name?

programs/x86_64/v2rayN
ivan-hc, line 48: If you mean the binary name, replace
```
chmod a+x ./"$APP"
```
with
```
chmod a+x ./v2rayN
```

programs/aarch64/v2rayN
ivan-hc: Also, replace
```
ln -s "/opt/$APP/$APP" "/usr/local/bin/$APP"
```
with
```
ln -s "/opt/$APP/v2rayN" "/usr/local/bin/$APP"
```

ivan-hc:
App names in AM are always lowercase. Not all options are case sensitive.

programs/x86_64/v2rayn
```diff
@@ -54,12 +54,11 @@ fi
 EOF
 chmod a+x ./AM-updater || exit 1
 
-
 # LAUNCHER
 echo "[Desktop Entry]
 Name=v2rayN
 Exec=$APP
-Icon=/opt/$APP/$APP.png
+Icon=/opt/$APP/v2rayN.png
```
ivan-hc: You don't need to do the same for the Icon entry of the .desktop file... and the extension is not needed

CelicaS: This is to show the icon.

ivan-hc: > This is to show the icon.

I see

Istantanea_2026-09-09_04-55-23

The issue here is that the path would not be compatible with the option `--icons` that allows you to use icon themes.

The correct path should be /usr/$APP/icons, so the.png file should be moved there... or uploaded to the catalogue, so it will be downloaded from there, if the script don't provides a way.

CelicaS:
@ivan-hc I referred to the AB Download Manager script. I also use this software and found that it does not update automatically. Is this related?
edit:AB Download Manager icon is missing
image

ivan-hc:
> @ivan-hc I referred to the AB Download Manager script. I also use this software and found that it does not update automatically. Is this related? edit:AB Download Manager icon is missing image

Thanks for reporting this. I just fixed both, and discovered a bug related to the new catalog layout.

Issue #2050, opened 2026-01-27
Add Calibre official and beta/preview


shuvashish76:
To help us make adding applications quicker and easier, please follow the syntax below (**NOTE that if point 2 is a github repository, you can remove points 3 and 4**, and **if the program is a CLI you can remove point 5 and 6**). Also consider learning how to create, test and finally add scripts yourself. Follow the guide https://github.com/ivan-hc/AM/blob/main/docs/guides-and-tutorials/template.md

1. calibre
2. https://calibre-ebook.com/
3. https://download.calibre-ebook.com/
4. The one stop solution to all your e-book needs

---

I know currently there is [unofficial appimage](https://portable-linux-apps.github.io/apps/calibre.html) maintained by you.

Is it possible to add official binary and [beta release](https://download.calibre-ebook.com/betas/). [Calibre v9 beta testing](https://www.mobileread.com/forums/showthread.php?t=371417) is going on, so new betas are released very frequently. It's bit annoying to manually going to that link and download it. 

So we'll have three packages at our list
- `calibre`  - from  https://download.calibre-ebook.com/
- `calibre-appimage` - from https://github.com/ivan-hc/Calibre-appimage
- `calibre-beta` - from https://download.calibre-ebook.com/betas/

Each version is located at different directories and Calibre is an complex software so the script might break at future release. If there is any general suggestion _(like naming convention)_ then let me know I'll inform them. ThankYou.

---

Betas are for major release there is also `preview` - https://download.calibre-ebook.com/preview/ see [this thread](https://www.mobileread.com/forums/showthread.php?t=359648) i.e. _minor changes mostly published on fridays_ for coming stable release. Do you think we need that as well? Calibre is one of my favorite software so I'm bit biased for testing.🫣

Maybe instead of `calibre-preview` and `calibre-beta` we combine them in one name as `calibre-nightly` as they're trageted for testers. Let the user pick option 1(preview) or 2(beta) when they use "nightly" pacakge?

ivan-hc:
My AppImage of Calbre is based on the official binary.

Sadly, the developer has disabled the issues tab in its repository... so I was forced to maintain the AppImage by myself.

The advantages of being AppImage are many, including isolation of dotfiles and bubblewrap sandboxing (via sas/aisap).

So are you asking me to revert this?

ivan-hc:
With this I'll also provide a beta

[...]

PR #2702, opened 2026-09-11
Add description to appimages & remove unofficial where it is not needed

ArqamQazi:
Adds description to many of the anylinux appimages and remove unofficial from app description where it is not needed.

### DIFF
--- a/programs/x86_64-apps
+++ b/programs/x86_64-apps
@@ -1,6 +1,6 @@
 ◆ 0ad : FOSS historical Real Time Strategy, RTS game of ancient warfare.
 ◆ 0ad-prerelease : FOSS historical Real Time Strategy, RTS game of ancient warfare (Pre-release).
-◆ 12to11 : Unofficial, tool for running Wayland applications on an X server.
+◆ 12to11 : Tool for running Wayland applications on an X server.
 ◆ 3d-puzzles : 3D-Puzzles are Rubik, Dogic and Dogic12.
 ◆ 3nity-media : Cross-platform media player built with Lazarus/Free Pascal and libmpv. Features playlist management, internet radio, equalizer, visualizations, stream recording, and 100 language translations.
 ◆ 432hz-player : Because most music is recorded in 440hz, Audio Player.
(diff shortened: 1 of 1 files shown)


ivan-hc:
Hi @ArqamQazi and thanks for this pull request. I'm glad to have someone who cared about descriptions of the apps.

About the "Unofficial" word, we added it as an attempt to report the unofficial AppImages we have in this database... but in these years, many have been submitted and the situation got out of hand.

The other architectures will likely require an update as well, but since x86_64 is the primary architecture, updating the others won't be a problem.

Thanks again.

ivan-hc:
Welcome on board!

shuvashish76:
IMO unofficial is the correct term. Really don't like the changes because when downloading a package I want to investigate just from the description that if the package is actually provided by Upstream dev or not. If not provided then okay I'll take mental note of it - ask upstream for AppImage support or don't want to install Unofficial maintained by someone else that I should trust or not etc and make an informed decision based on it.

With current description it's all gone I've to dig in if the repo is from upstream or not etc. There is a reason why all the AppImages maintained @ivan-hc and pkgforge are always called **Unofficial**.

Request to revert the changes.

---

It's not like grabbing `.exe` file from wherever you find on the internet. Linux in general based upon trust and users should be informed about the Unofficial nature of AppImages, though choice is up-to users.

ArqamQazi:
I have added “unofficial” to several apps where we might have trademark issues.

I have only modified apps from pkgforge-Dev, which are all unofficial and will remain unofficial. Maybe we could mark all pkgforge apps as unofficial.

programs/x86_64-apps
```diff
@@ -392,54 +392,54 @@
 ◆ cadmus : Pulse Audio real-time noise suppression plugin.
 ◆ caesium : Image compression software that helps you store, send and share digital pictures, supporting JPG, PNG, WebP and TIFF formats.
 ◆ cake-wallet : Easily and safely store, send, receive, and exchange your cryptocurrency.
-◆ calibre : Unofficial. The one stop solution to all your e-book needs.
-◆ calibre-beta : Unofficial. The one stop solution to all your e-book needs. Beta version.
-◆ calibre-preview : Unofficial. The one stop solution to all your e-book needs. Preview version.
+◆ calibre : The one stop solution to all your e-book needs.
+◆ calibre-beta : The one stop solution to all your e-book needs. Beta version.
+◆ calibre-preview : The one stop solution to all your e-book needs. Preview version.
```
shuvashish76: Calibre dev has no problem considering bug-reports from @ivan-hc 's AppImages https://github.com/ivan-hc/AM/issues/2050#issuecomment-3808672955

but I think still they should be called Unofficial. Again nothing modified but to trust ivan-hc or not that should be left to users. I/We can't make such decision for users.

ivan-hc:
> With current description it's all gone I've to dig in if the repo is from upstream or not etc. There is a reason why all the AppImages maintained @ivan-hc and pkgforge are always called **Unofficial**.

I will try to re-add the Unofficial word to descriptions for my AppImages and pkgforge's

shuvashish76:
ThankYou, there are also Unofficial AppImages packages maintained by others. I hope they're not changed in this PR, too many changes don't want to check them all. Easy to add Unofficial label when adding NEW packages but difficult to track them down later.

_(If later something changed i.e. if official repo added AppImage support, unofficial discontinued then AM users should report them - that's not an issue.)_ Better if supported by Official repos, less burden on  maintainers :)

PR #2713, opened 2026-09-15
Add librelane, magic

MrAbhi19:


ivan-hc:
Hi @MrAbhi19 , you forgot to add the app to the lists.

See the "list" file created near the two directories, there is a line to add in the two *-apps files near it.

Without that line, the app will not appear in AM, nor in the [catalog](https://portable-linux-apps.github.io/).

ivan-hc:
Thanks for the fix.

I think you should rename this PR since you added also another app.

If you want to add other ones, get the time you need... then let me know when you are ready, and I will merge.

MrAbhi19:
Hi @ivan-hc,

Thanks for catching that — you’re right, I missed adding the app to the lists.
I’ve now added the required line in both *-apps files (as indicated in the list file).
I’ve also added another one: `magic`.

MrAbhi19:
> Thanks for the fix.
> 
> I think you should rename this PR since you added also another app.
> 
> If you want to add other ones, get the time you need... then let me know when you are ready, and I will merge.

I guess that's it for now. you can merge. Thanks : )

ivan-hc:
Welcome on board @MrAbhi19

MrAbhi19:
Hi @ivan-hc, just a doubt 

https://github.com/RTimothyEdwards/magic/releases/tag/8.3.683
here we have 4 AppImages (for x86_64), so what AppImage does the AM ships? and is there any way to select them while installing?

ivan-hc:
> Hi @ivan-hc, just a doubt
> 
> https://github.com/RTimothyEdwards/magic/releases/tag/8.3.683 here we have 4 AppImages (for x86_64)

They should change method to bundle AppImages. They build too many. Made them aware about Anylinux:

https://github.com/pkgforge-dev/Anylinux-AppImages

it will build one that would work for all GNU/Linux distributions.

Open an issue in their repository asking them to fix their build method.

> so what AppImage does the AM ships?

The first in this list:
```
curl -Ls https://api.github.com/repos/RTimothyEdwards/magic/releases | sed 's/[()",{} ]/\n/g' | grep -oi "https.*download.*mage$" | grep -vi "i386\|i686\|armhf\|aarch64\|arm64\|armv7l"
``` 

> and is there any way to select them while installing?

No, you can "downgrade" it btw:
```
am downgrade magic
```
Then select the one you want.

---------------------------------

Again, ask them to fix their build method. They use too many resources and do too much job for a thing that would be simpler.

PR #2719, opened 2026-09-16
Improve WeChat version checking.

DtDEncryption:
Found the official version endpoint for WeChat: https://linux.weixin.qq.com/api/version

 Here is the patch for it.


CelicaS:
Your script fails to install. Right now, I suggest using a repository like the one below. It would be even better if @ivan-hc is interested in forking and maintaining it (that would be ideal), as this would also allow downgrading and adding ARM packages.
https://github.com/trouter-ai/wechat-linux-versions
https://github.com/Rodert/wechat-linux-versions

ivan-hc:
> Found the official version endpoint for WeChat: https://linux.weixin.qq.com/api/version
> 
> Here is the patch for it.

The current script already gets version and download URL. Yours don't just check the number version, but also break the download.

In fact... :point_down: 

> Your script fails to install.

@DtDEncryption according to your code https://github.com/DtDEncryption/AM/blob/9aee37423e95cbfc0a7a1beaaadb9956a616cc61/programs/x86_64/wechat#L14-L18

```
# DOWNLOAD AND PREPARE THE APP, $version is also used for updates
version=$(curl -Ls https://linux.weixin.qq.com/api/version)
sitecontent=$(curl -Ls https://linux.weixin.qq.com/en)
download=$(echo "$sitecontent" | tr '">< ' '\n' | grep -i "http.*appimage$" | grep -vi "arm\|aarch" | head -1)
wget "$version" || exit 1
```
1. $version is a number
2. $sitecontent is called to read the site page
3. $download is called to get the download URL
4. `wget` should download $download, not $version
```
ivan@debian:~$ version=$(curl -Ls https://linux.weixin.qq.com/api/version)
ivan@debian:~$ echo $version
4.1.13
ivan@debian:~$ sitecontent=$(curl -Ls https://linux.weixin.qq.com/en)
ivan@debian:~$ download=$(echo "$sitecontent" | tr '">< ' '\n' | grep -i "http.*appimage$" | grep -vi "arm\|aarch" | head -1)
ivan@debian:~$ echo $download
https://dldir1v6.qq.com/weixin/Universal/Linux/WeChatLinux_x86_64.AppImage
```
This is why `wget "$version"` fails.

> Right now, I suggest using a repository like the one below. [...] as this would also allow downgrading and adding ARM packages. https://github.com/trouter-ai/wechat-linux-versions https://github.com/Rodert/wechat-linux-versions

@CelicaS I prefer to point to official sources, these seems to be random repositories opened in few months and that could be stopped soon or later, without guarantee of continuity.

> It would be even better if @ivan-hc is interested in forking and maintaining it (that would be ideal)

I stopped to build ApImages of programs I don't use/know. And I doubt I will use WeChat. I'm Italian, and I don't know anyone that uses it.

DtDEncryption:
@ivan-hc 
Sorry for not checking the patch in advance... and really thank you for your patient explanation!

ivan-hc:
No problem

PR #2722, opened 2026-09-17
WeChat version checking improvement

DtDEncryption:
Sorry for the lack of verification before. This new patch should be clear of similar problems to the previous patch.

One day I might try to figure out if there is a way to make checksum verification work normally on this app.

---

This is a patch for WeChat app to have better version checking by using the official version endpoint.

It's done by replacing 
```version=$(curl -s https://linux.weixin.qq.com/ | grep -oP "https://[\S]+x86_64.AppImage")```
with
```version=$(curl -Ls https://linux.weixin.qq.com/api/version)```
in ```programs/x86_64/wechat```

### CODE REVIEW

File: programs/x86_64/wechat [hunk 1]
```diff
@@ -31,15 +33,15 @@ cat >> ./AM-updater << 'EOF'
...
 version0=$(cat "/opt/$APP/version")
+version=$(curl -Ls https://linux.weixin.qq.com/api/version)
 sitecontent=$(curl -Ls https://linux.weixin.qq.com/en)
-version=$(echo "$sitecontent" | sed 's/div/\n/g' | grep -i "bd-version" | tr '">< ' '\n' | grep -i "^[0-9].*[0-9]$" | tail -1)
 download=$(echo "$sitecontent" | tr '">< ' '\n' | grep -i "http.*appimage$" | grep -vi "arm\|aarch" | head -1)
 [ -n "$version" ] || { echo "Error getting link"; exit 1; }
 if command -v appimageupdatetool >/dev/null 2>&1; then
 	cd "/opt/$APP" || exit 1
-	appimageupdatetool -Or ./"$APP" && chmod a+x ./"$APP" && echo "$version" > ./version && exit 0
+	appimageupdatetool -Or ./"$APP" && chmod a+x ./"$APP" && echo "$version" > ./version && echo "$download" >> ./version && exit 0
```
ivan-hc, line 44: It is not so that works the checksum verification here.

By default, the version file should contain an URL (standalone). It would be easier if their artifact would contain a version number.

DtDEncryption:
Holy... What a fool I am... Really frustrating that I would keep making such mistakes. An LLM agent might do much better than me...

